PR #7, opened 2019-05-13
Feature/channellist pagination

jaapbakker88:
# Submit a pull request

## CLA

**Added** 
- LoadMorePaginator
- InfiniteScrollPaginator
- Moved channel query logic and state to Chat.js so App.js looks something like this now.
- New message moves channel to top, also when you're not watching the channel. It moves the channel when: you post, on `message.new` and on `notification.message_new` and also on `notification.added_to_channel` and `notification.removed_from_channel`
- Added props `onAddedToChannel` and `onRemovedFromChannel` to override functionality

because the querying now happens in Chat the filters and sort options need to be passed down to Chat.js. This changes the setup of App.js.
```
render() {
    const filters = { type: 'messaging', example: 1 };
    const sort = {
      last_message_at: -1,
      cid: 1,
    };
    const options = {
      member: true,
      watch: true,
      limit: 30,
    };
    return (
      <Chat
        client={this.chatClient}
        filters={filters}
        sort={sort}
        options={options}
        theme={`team ${theme}`}
      >
        <ChannelList
          List={ChannelListTeam}
          Paginator={(props) => (
            <InfiniteScrollPaginator threshold={300} {...props} />
          )}
        />
      </Chat>
    );
  }
}
```


src/components/ChannelList.js
```diff
@@ -14,74 +19,224 @@ import { isPromise } from '../utils';
...
+      console.warn(e);
+      this.setState({ error: true, refreshing: false });
     }
-  }
+  };
 
-  static getDerivedStateFromError() {
-    return { error: true };
+  listenToChanges() {
+    this.props.client.on(this.handleEvent);
```
vishalnarkhede: Better to switch it off in `componentWillUnmount`. It will cause memory leak if customer decides to keep `ChannelList` and `Channel/MessageList` component under separate `Chat` components and it involves lot of mounting and unmounting while switching between channellist and messagelist components, which is the case fpr react native.

PR #34, opened 2019-07-24
Fixing component docs (in progress)

vishalnarkhede:
# Submit a pull request

## CLA


src/components/Channel.js
```diff
@@ -31,14 +31,46 @@ class Channel extends PureComponent {
...
+     *
+     * @param {Event} event DOM Click event
+     * @param {User} user   Target [user object](https://getstream.io/chat/docs/#chat-doc-set-user) which is clicked
+     */
     onMentionsClick: PropTypes.func,
-    /** Function to be called when hovering over a @mention. Function has access to the DOM event and the target user object */
+    /**
+     * Function to be called when a @mention in message is hovered.
+     *
+     * @param {Event} event DOM Click event
```
jaapbakker88: hover event?

src/components/MessageList.js
```diff
@@ -34,19 +36,17 @@ class MessageList extends PureComponent {
...
-    messages: PropTypes.array.isRequired,
-    /** Via Context: The channel object */
-    channel: PropTypes.object.isRequired,
-    /** Via Context: The function to update a message, handled by the Channel component */
-    updateMessage: PropTypes.func.isRequired,
-    /** Via Context: The function is called when the list scrolls */
-    listenToScroll: PropTypes.func,
-    /** Typing indicator component to render  */
+    /**
+     * Typing indicator UI component to render
```
jaapbakker88: Can we do this for every UI component. Had someone ask about the Message prop on MessageList and it was missing from the docs. Saw you added that! Would be awesome if we can have: "xxx UI component to render" for all these props.

src/components/MessageList.js
```diff
@@ -34,19 +36,17 @@ class MessageList extends PureComponent {
...
+     * Defaults to and accepts same props as: [TypingIndicator](https://github.com/GetStream/stream-chat-react/blob/master/src/components/TypingIndicator.js)
+     * */
     TypingIndicator: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
-    /** Date separator component to render  */
+    /**
+     * Date separator UI component to render
+     *
+     * Defaults to and accepts same props as: [DateSeparator](https://github.com/GetStream/stream-chat-react/blob/master/src/components/DateSeparator.js)
+     * */
     dateSeparator: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
jaapbakker88: This should probably be DateSeparator as it's a UI component. But that might just be a breaking change. :/

jaapbakker88: maybe removing the this.props.Message.someMethod could be part of that as well.

PR #81, opened 2019-11-04
set customActiveChannel

jaapbakker88:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/ChannelList.js
+++ b/src/components/ChannelList.js
@@ -125,6 +125,10 @@ class ChannelList extends PureComponent {
      * @see See [Pagination documentation](https://getstream.io/chat/docs/#channel_pagination) for a list of available fields for sort.
      * */
     watchers: PropTypes.object,
+    /**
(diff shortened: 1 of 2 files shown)


src/components/ChannelList.js
```diff
@@ -213,6 +218,17 @@ class ChannelList extends PureComponent {
           refreshing: false,
         };
       });
+
+      // Set a channel as active and move it to the top of the list.
+      if (this.props.customActiveChannel) {
```
vishalnarkhede: Can we move this logic to line 202? That way we will only make call to `setActiveChannel` once

jaapbakker88: Yeah, the problem is we don't have the channels in state yet over there.

vishalnarkhede: ah ok!!

ludovicthomas:
@vishalnarkhede it seems that if the `customActiveChannel` contains a `channelId` that is not loaded from the channels response, it will not load it. Is it a known issue? Thanks for your feedback.

Issue #101, opened 2019-12-23
messageActions/MessageInput parameters support for Thread component


volodymyr-manko:
Currently, **Thread** component has no parameters like **messageActions** or **MessageInput**.
So, regardless of commands/permissions configuration on Dashboard, MessageList component (Thread -> MessageList), gets **messageActions** property as undefined, and in result the default list of actions (["edit", "delete", "mute", "flag"]) is used, which cannot be overridden for Thread.

As well there is no ability to customize MessageInput component for Thread (MessageInputSmall is used as Input prop of MessageInput instead): so the custom MessageInput in chat can differ from the one in Thread in that chat.

Could you please support these input parameters for Thread, like **messageActions**/**MessageInput**, which will be passed to MessageList and MessageInput components respectively?

```
messageActions: PropTypes.oneOfType([PropTypes.bool, PropTypes.array]),
MessageInput: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
...

<MessageList
  ...
  messageActions={this.props.messageActions}
 />
 <MessageInput
   Input={this.props.MessageInput}
   ...
 />
```

vishalnarkhede:
Thanks @volodymyr-manko for bringing this up. We are looking into it and get back to you soon with resolution :)

volodymyr-manko:
Thanks :)

dmcguire81:
@tbarbugli @vishalnarkhede this could also support our use case of specializing thread-handling.

jaapbakker88:
@volodymyr-manko the thread component now supports the following props in version `0.7.15`. This allows you to update the underlying MessageList, MessageInput and Message components, through props.
- `additionalParentMessageProps`
- `additionalMessageListProps`
- `additionalMessageInputProps`

volodymyr-manko:
@jaapbakker88 
Cool, thanks! :)

PR #102, opened 2019-12-24
override sendMessage request in Channel props

mahboubii:
This prop allow users to pass a custom function to catch network request for sending a new message. This can be used to reroute the new messages to custom API and send the message after some validation.

### DIFF
--- a/src/components/Channel.js
+++ b/src/components/Channel.js
@@ -88,6 +88,12 @@ class Channel extends PureComponent {
     acceptedFiles: PropTypes.array,
     /** Maximum number of attachments allowed per message */
     maxNumberOfFiles: PropTypes.number,
+    /** Override send message request (Advanced usage only)
+     *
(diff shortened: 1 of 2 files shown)


tschellenbach:
3rd customer so far that wants to run some checks before sending the message. We should have a pre-send message webhook in the future. @tbarbugli

dmcguire81:
@tbarbugli @tschellenbach we have a use case that could be supported equally well by solutions discussed here or by [adding parameters to the `Thread` implementation](https://github.com/GetStream/stream-chat-react/issues/101). Any thoughts on which is preferred on your end, and/or which is more likely to be scheduled in the near-term?

mahboubii:
> 3rd customer so far that wants to run some checks before sending the message. We should have a pre-send message webhook in the future. @tbarbugli

also webhook for pre-update(edit)

Issue #97, opened 2019-12-06
SendButton missing in types definiton


damianprzygodzki:
I can't add sendbutton to MessageInput despite it is taken as prop in component.

jaapbakker88:
@damianprzygodzki this should be fixed in version `0.7.15`, can you confirm this works for you?

damianprzygodzki:
Yes, thank you!

PR #109, opened 2019-12-30
add sendbutton to types.d.ts

jaapbakker88:
# Submit a pull request

Regarding #97 

## CLA

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -242,6 +242,11 @@ export interface EmptyStateIndicatorProps {
   listType: string;
 }
 
+export interface SendButtonProps {
+  /** Function that gets triggered on click */
+  sendMessage?(message: Client.Message): void;
+}
+
 export interface MessageListProps extends ChannelContextValue {
   /** Typing indicator component to render  */
   TypingIndicator?: React.ElementType<TypingIndicatorProps>;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -286,6 +286,9 @@ export interface MessageInputProps {
   /** The component handling how the input is rendered */
   Input?: React.ElementType<MessageInputUIComponentProps>;
 
+  /** Change the SendButton component */
+  SendButton?: React.ElementType<any>;
```
vishalnarkhede: Instead of `any`, we should create an interface called `SendButtonProps` ... and assign it here!!

damianprzygodzki:
But for god sake! How i should obtain message for send button? What's the idea of requiring message in SendButton?

tbarbugli:
@jaapbakker88 the message is passed to the button component's `sendMessage` function right? perhaps we should include an example of a custom button in our docs

@damianprzygodzki what's your use-case for replacing the SendButton component?

damianprzygodzki:
I just want to place another component in that place to MessageInput as prop (as documented).

MessageInput already have that message, so i suppose SendButton is only responsible for 'clicking'.

jaapbakker88:
@tbarbugli I'll add an example of how to use this in our docs. @damianprzygodzki the sendMessage prop just accepts a function that sends the message. Could have been more conveniently  named I think.

PR #125, opened 2020-01-08
ChannelPreview latestMessage length customizable

mahboubii:


mahboubii:
> @mahboubii Just make sure `latestMessage` is only used in `ChannelPreviewLastMessage.js` !!

the only other place is `ChannelPreviewMessenger` which has it's own truncate function applied to the message

src/components/ChannelPreviewLastMessage.js
```diff
@@ -50,7 +56,10 @@ export class ChannelPreviewLastMessage extends PureComponent {
             <span className="str-chat__channel-preview-last-message">
               {!this.props.channel.state.messages[0]
                 ? 'Nothing yet...'
-                : this.props.latestMessage}
+                : truncate(
```
vishalnarkhede: @mahboubii Does this work? Because as per sytax, 2nd parameter is an object for `truncate` function - https://lodash.com/docs/4.17.15#truncate

mahboubii: yea last time i manually tested it

PR #127, opened 2020-01-08
fix lodash truncate options

mahboubii:


vishalnarkhede:
@mahboubii closing this PR. There is a conflict so I have pushed these changes separately !!

PR #130, opened 2020-01-14
allow changing filters to update the channellist

jaapbakker88:
# Submit a pull request
> I am pretty sure this is also happening on the React SDK (in that case I suggest to solve it there first and then copy/paste the same solution to RN to save dev time)

https://github.com/GetStream/stream-chat-react-native/issues/113

## CLA

### DIFF
--- a/src/components/ChannelList.js
+++ b/src/components/ChannelList.js
@@ -12,6 +12,7 @@ import { withChatContext } from '../context';
 import { ChannelListTeam } from './ChannelListTeam';
 import { smartRender } from '../utils';
 import uniqBy from 'lodash.uniqby';
+import deepequal from 'deep-equal';
 
 /**
  * ChannelList - A preview list of channels, allowing you to select the channel you want to open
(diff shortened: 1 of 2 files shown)


src/components/ChannelList.js
```diff
@@ -193,6 +194,18 @@ class ChannelList extends PureComponent {
     this.listenToChanges();
   }
 
+  async componentDidUpdate(prevProps) {
+    if (!deepequal(prevProps.filters, this.props.filters)) {
```
ferhatelmas: why do we need deep equal ? Isn't shallow equal enough?

src/components/ChannelList.js
```diff
@@ -12,6 +12,7 @@ import { withChatContext } from '../context';
 import { ChannelListTeam } from './ChannelListTeam';
 import { smartRender } from '../utils';
 import uniqBy from 'lodash.uniqby';
+import deepequal from 'deep-equal';
```
ferhatelmas: naming `deepEqual` might be better similar to `uniqBy`

PR #142, opened 2020-02-03
CRS-14: UI and integration Tests and hooks

vishalnarkhede:
# Submit a pull request

## CLA

I am using [testing-library](https://testing-library.com/docs/intro) for testing component. We can also add snapshot tests, but I find it pretty useless for our purpose since our components have too many moving things inside. So we should limit our tests to only user interactions and what user actually sees on screen.

Also moving implementations of components to hooks.

---- Update ----

I will separate out hooks implementation to separate PR actually!!


__tests__/ChannelPreviewMessenger-test.js
```diff
@@ -0,0 +1,103 @@
...
+    const { getByText } = render(
+      <ChannelPreviewMessenger
+        latestMessage=""
+        channel={channel}
+        activeChannel={channel}
+        setActiveChannel={() => {}}
+      />,
+    );
+
+    const emptyMessageText = 'Nothing yet...';
```
tbarbugli: bit of a fragile test, you change the empty message and you break a test :/ (btw probably not a bad idea to allow users to overwrite the empty preview via prop)

__tests__/ChannelPreviewMessenger-test.js
```diff
@@ -0,0 +1,103 @@
...
+    expect(getByText(emptyMessageText)).toBeTruthy();
+    expect(getByText(channelName)).toBeTruthy();
+  });
+
+  it('Latest message should be present in cacatenated state', () => {
```
tbarbugli: typo aside, you should make test names more explicit; this test does 2 things but the description is not very clear about it

src/components/ChannelPreview.js
```diff
@@ -1,116 +1,110 @@
-import React, { PureComponent } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 
-import { ChannelPreviewCountOnly } from './ChannelPreviewCountOnly';
+export const ChannelPreview = ({
```
mahboubii: you can use `useMemo` to achieve the same performance as the `pureComponent`

src/components/ChannelPreview.js
```diff
@@ -1,116 +1,110 @@
...
-      const isActive = this.props.activeChannel.cid === this.props.channel.cid;
-      if (isActive) {
-        this.setState({ unread: 0, lastRead: new Date() });
-      }
-    }
-  }
+    const unread = channel.countUnread(lastRead);
+    setLatestMessage(getPreviewLatestMessage(channel));
+    setUnread(unread);
+  }, [channel, activeChannel]);
```
mahboubii:
```suggestion
  }, [channel, activeChannel, isActive]);
```

src/components/ChannelPreview.js
```diff
@@ -1,116 +1,110 @@
...
+
+  const isActive = useCallback(() => activeChannel.cid === channel.cid, [
+    activeChannel,
+    channel,
+  ]);
```
mahboubii: not that it's a big deal but you can also leverage `useMemo` for something like this too which let you skip one function call
`const isActive = useMemo(() => activeChannel.cid === channel.cid, [activeChannel, channel]);`

src/components/ChannelPreviewCompact.js
```diff
@@ -8,53 +8,61 @@ import PropTypes from 'prop-types';
...
+  channel,
+  unread,
+  setActiveChannel = () => {},
+  watchers,
+  active,
+}) => {
+  const [unreadClass, setUnreadClass] = useState('');
+  const [activeClass, setActiveClass] = useState('');
+  const [name, setName] = useState('');
+  const channelPreviewButton = React.createRef();
```
mahboubii:
```suggestion
  const channelPreviewButton = useRef();
```

src/components/ChannelPreviewCompact.js
```diff
@@ -8,53 +8,61 @@ import PropTypes from 'prop-types';
...
-    unread: PropTypes.number,
-    /** If channel of component is active (selected) channel */
-    active: PropTypes.bool,
-    /** Latest message's text. */
-    latestMessage: PropTypes.string,
-  };
+export const ChannelPreviewCompact = ({
+  channel,
+  unread,
+  setActiveChannel = () => {},
```
mahboubii: nothing wrong with this, but it's best practice to use defaultProps

src/components/ChannelPreviewCompact.js
```diff
@@ -8,53 +8,61 @@ import PropTypes from 'prop-types';
...
+  const channelPreviewButton = React.createRef();
 
-  channelPreviewButton = React.createRef();
+  useEffect(() => {
+    setUnreadClass(
+      unread >= 1 ? 'str-chat__channel-preview-compact--unread' : '',
+    );
+    setActiveClass(active ? 'str-chat__channel-preview-compact--active' : '');
+    setName(channel.data.name || channel.cid);
+  });
```
mahboubii: probably missing dependency array

src/components/ChannelPreviewCompact.js
```diff
@@ -8,53 +8,61 @@ import PropTypes from 'prop-types';
...
 
-  onSelectChannel = () => {
-    this.props.setActiveChannel(this.props.channel, this.props.watchers);
-    this.channelPreviewButton.current.blur();
+  const onSelectChannel = () => {
+    setActiveChannel(channel, watchers);
+    channelPreviewButton.current.blur();
   };
```
mahboubii: could be wrapped in `useCallback`

src/components/ChannelPreviewLastMessage.js
```diff
@@ -10,73 +10,77 @@ import truncate from 'lodash/truncate';
...
+  channel,
+  emptyMessageText = 'Nothing yet...',
+  latestMessageLength = 14,
+  unread,
+  latestMessage,
+  setActiveChannel = () => {},
+  watchers,
+  active,
+}) => {
+  const channelPreviewButton = React.createRef();
```
mahboubii: useRef()

src/components/ChannelPreview.js
mahboubii: also since this function is pretty simple evaluation you can ditch both useCallback and useMemo and just evaluate it on every render, simple as: `const isActive = activeChannel.cid === channel.cid`

__tests__/ChannelPreviewCompact-test.js
```diff
@@ -0,0 +1,116 @@
...
+
+    const containerBtn = container.querySelector(
+      '.str-chat__channel-preview-compact',
+    );
+    fireEvent.click(containerBtn);
+    expect(setActiveChannel).toHaveBeenCalledTimes(1);
+    expect(setActiveChannel).toHaveBeenLastCalledWith(channel, watchers);
+  });
+
+  it('should have str-chat__channel-preview-compact--active class only when channel is active', () => {
```
vishalnarkhede: tests based on css class is not exactly idea, but here thats what end user will see (different styles based on props), so making an exception here.

PR #218, opened 2020-04-22
Eslint config

mahboubii:
# Submit a pull request

## CLA


vishalnarkhede:
@jaapbakker88 @mahboubii should we also ignore example directory at all from eslint (just need extra line in `.eslintignore`)? I hate it when I can't commit while making changes in example app!!

mahboubii:
> @jaapbakker88 @mahboubii should we also ignore example directory at all from eslint (just need extra line in `.eslintignore`)? I hate it when I can't commit while making changes in example app!!

we could, or we can come up with a more relaxed config only for example apps. but in general, it makes sense to apply the same standard to those codes, especially they are more user-facing.

jaapbakker88:
@mahboubii @vishalnarkhede 

> > @jaapbakker88 @mahboubii should we also ignore example directory at all from eslint (just need extra line in `.eslintignore`)? I hate it when I can't commit while making changes in example app!!
> 
> we could, or we can come up with a more relaxed config only for example apps. but in general, it makes sense to apply the same standard to those codes, especially they are more user-facing.

@vishalnarkhede @mahboubii I think it makes sense to check it, but I think the same goes for this one. We're going to have to check and update them anyway. So let's do those when we're editing them. Otherwise this PR become way too big.

PR #224, opened 2020-04-30
API mocks for UI tests - proposal

vishalnarkhede:
# Submit a pull request

## CLA

### Approach

Instead of intercepting the network requests, which can be done by mockttp or sinon libraries, I instead went for mocking the modules approach. 

**Advantages:**

- Ability to request the response from api in whatever way that the test demands
- Mocks can be extracted to separate package
- Reusable for react-native sdk in future

### Jest Module Mock

Jest allows us to mock the modules following way ([ref](https://jestjs.io/docs/en/es6-class-mocks#automatic-mock))
```js
import someModule from 'someModule';
jest.mock('someModule')

someModule.funcCall.mockResolvedValue(50);
```
In component, if you import the module - `someModule` and call `funcCall()` on it, it will return 50.

In similar way we are going to mock the axios here, and functions on it - get, post, put, delete
Check this file - https://github.com/GetStream/stream-chat-react/pull/224/files#diff-8290d0448eb9afb7b412b06ee37bae50

**Note** - For most of the UI tests we won't need api or event mocks. It will be mostly in MessageList, ChannelList

### Directory structure

All the mocks related code will be under `src/__mocks__` folder, and tests will import everything from this folder.
`__mocks__`  has following 3 subdirectories:

- **generator** - contains all the mock data generator (nothing to do with ES6 generators) functions e.g., channel, message

e.g., 
```js
const user1 = generateUser();
const message1 = generateMessage({ user: user1 });
const message2 = generateMessage({ user: user1 });

const channel1 = generateChannel();
const channel2 = generateChannel({
    messages: [message1]
})
```
- **api** - contains all the api response functions such as queryChannelsApi, sendMessageApi, getOrCreateChannelsApi. These functions simply build a api response out of the data given to it as param.

e.g.
```js, 
const channel1 = generateChannel();
const channel2 = generateChannel();

const apiResponse = queryChannelsApi([channel1, channel2]);

// apiResponse will be something like following
{
  type: 'get',
  response: {
    data: {
      channels: [{ ...}, { ...}],
      duration: 0.01
    }
  }
}
```

You can also mock api with error. All the errored api mocks are in `__mocks__/api/error.js`.

```js
const erroredResponse = erroredPostApi();
```

- **event** - contains all the mocked event dispatchers
e.g., dispatchMessageNewEvent. These functions will simply dispatch a fake event on client.

Screenshot 2020-04-30 at 10 30 07

### Example

When you write a test, you should know what apis are going to be called, and accordingly hook in the mocked api responses for that test:

Lets look at one example of a test:

_Note_: we are using [testing-library](https://testing-library.com/) for dom testing.

```js
import {
  getTestClient,
  useMockedApis,
  queryChannelsApi,
  generateMessage,
  generateChannel,
  generateUser,
  getOrCreateChannelApi,
  dispatchMessageNewEvent,
  dispatchNotificationAddedToChannelEvent,
} from '../../../__mocks__';

it('should move the channel to top if new message is received', async () => {
    // Here we want to test the ChannelList component with 3 channels.
    // So lets quickly create 3 channels. When no params are given,
    // it will generate a fake channel with empty messages and members.
    // But you can request channel containing some messages or members as well. 
    const channel1 = generateChannel();
    const channel2 = generateChannel();
    const channel3 = generateChannel();

    // When ChannelList gets rendered, it will make a call to queryChannels api.
    // So lets hook in that api response.
    // For this purpose we use `useMockedApis` function, which basically
    // mocks the axios module to return fake response when queryChannels
    // api gets called. If you are expecting more api calls, then include its responses
    // if the array.
    useMockedApis(axios, [queryChannelsApi([channel1, channel2, channel3])]);

    chatClientVishal = await getTestClient({ id: 'vishal' });

    const { getByText, getByRole, getAllByRole } = render(
        <Chat client={chatClientVishal}>
        <ChannelList
            filters={{}}
            Preview={ChannelPreviewComponent}
            List={ChannelListComponent}
            options={{ state: true, watch: true, presence: true }}
        />
        </Chat>,
    );

    // Wait for list of channels to load in DOM. I have used a custom `List` component
    // which has `role="list"` on its dom element.
    await waitFor(() => {
        expect(getByRole('list')).toBeInTheDocument();
    });

    // Generate a new message with given user.
    const newMessage = generateMessage({
        user: generateUser(),
    });

    // Dispatch the event on client, so ChannelList will react to that event
    dispatchMessageNewEvent(chatClientVishal, newMessage, channel3.channel);

    await waitFor(() => {
        expect(getByText(newMessage.text)).toBeInTheDocument();
    });

    const items = getAllByRole('listitem');

    // Get the closes listitem to the channel that received new message.
    const channelPreview = getByText(newMessage.text).closest(
        '[role="listitem"]',
    );
    expect(channelPreview.isEqualNode(items[0])).toBeTruthy();
});
```


src/__mocks__/index.js
```diff
@@ -0,0 +1,24 @@
+import { StreamChat } from 'stream-chat';
+import { v4 as uuidv4 } from 'uuid';
+
+const apiKey = 'API_KEY';
+const apiSecret = 'API_SECRET';
+
+export async function getTestClient(user) {
```
mahboubii, line 7:
```suggestion
export function getTestClient(user) {
```

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+  useMockedApis,
+  queryChannelsApi,
+  generateMessage,
+  generateChannel,
+  generateUser,
+  getOrCreateChannelApi,
+  dispatchMessageNewEvent,
+  dispatchNotificationAddedToChannelEvent,
+} from '../../../__mocks__';
+import axios from 'axios';
```
mahboubii, line 19: this should be moved to the top, eslint could catch these

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+  getTestClient,
+  useMockedApis,
+  queryChannelsApi,
+  generateMessage,
+  generateChannel,
+  generateUser,
+  getOrCreateChannelApi,
+  dispatchMessageNewEvent,
+  dispatchNotificationAddedToChannelEvent,
+} from '../../../__mocks__';
```
mahboubii, line 18: @jaapbakker88 @vishalnarkhede we can setup babel module resolver to get rid of all these `../../..`, at least for tests

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+    <div>{channel.data.name}</div>
+    <div>{latestMessage}</div>
+  </div>
+);
+
+const ChannelListComponent = (props) => {
+  const { error, loading, LoadingErrorIndicator, LoadingIndicator } = props;
+  if (error) {
+    return <LoadingErrorIndicator type="Connection Error" />;
+  } else if (loading) {
```
mahboubii, line 41: else if not needed

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+    return <LoadingIndicator />;
+  } else {
```
mahboubii, line 43: redundant else

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+  let chatClientVishal;
+
+  it('should move the channel to top if new message is received', async () => {
+    const channel1 = generateChannel();
+    const channel2 = generateChannel();
+    const channel3 = generateChannel();
+
+    useMockedApis(axios, [queryChannelsApi([channel1, channel2, channel3])]);
+
+    chatClientVishal = await getTestClient({ id: 'vishal' });
```
mahboubii, line 60: not sure if these tests are running in parallel atm, but I see getTestClient is called twice. does it cause an issue with the mocks if it's moved to `describe` scope?

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,141 @@
...
+    const channel1 = generateChannel();
+    const channel2 = generateChannel();
+    const channel3 = generateChannel();
+
+    useMockedApis(axios, [
+      queryChannelsApi([channel1, channel2]),
+      getOrCreateChannelApi(channel3),
+    ]);
+
+    chatClientVishal = await getTestClient({ id: 'vishal' });
```
mahboubii, line 107:
```suggestion
    chatClientVishal = getTestClient({ id: 'vishal' });
```

src/components/InfiniteScrollPaginator/ReverseInfiniteScroll.js
```diff
@@ -224,6 +224,8 @@ class ReverseInfiniteScroll extends Component {
       this.scrollComponent = node;
     };
 
+    props['data-testid'] = 'message-list';
```
mahboubii, line 227: this could be confusing since we also have a MessageList component

vishalnarkhede: Oh yeah, sorry!! That needs to be fixed ... I forgot to fix it. THanks!!

vicnicius:
I wonder if adding those test helpers to the `__mocks__` namespace wouldn't be confusing, as jest uses the same namespace to implement [manual mocks](https://jestjs.io/docs/en/manual-mocks).

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -0,0 +1,172 @@
...
+  generateChannel,
+  generateUser,
+  getOrCreateChannelApi,
+  dispatchMessageNewEvent,
+  dispatchNotificationAddedToChannelEvent,
+  getTestClientWithUser,
+  erroredGetApi,
+} from '__mocks__';
+
+// eslint-disable-next-line no-undef
```
vicnicius, line 24: Not sure if it would be on the scope of this PR, but we can add jest globals to eslint using https://github.com/jest-community/eslint-plugin-jest.

src/components/MessageList/__tests__/MessageList.test.js
```diff
@@ -0,0 +1,72 @@
...
+      ],
+      members: [
+        generateMember({ user: user1 }),
+        generateMember({ user: user1 }),
+      ],
+    });
+
+    useMockedApis(axios, [getOrCreateChannelApi(mockedChannel)]);
+
+    chatClientVishal = await getTestClientWithUser({ id: 'vishal' });
```
vicnicius, line 46:
```suggestion
    const chatClientVishal = await getTestClientWithUser({ id: 'vishal' });
```

src/components/ChannelList/__tests__/ChannelList.test.js
mahboubii: agree #230

vishalnarkhede: We can change the name? Any suggestions? How about `__mock-builders__`? Since we only export functions from this directory and not mock objects.

vishalnarkhede:
> I wonder if adding those test helpers to the `__mocks__` namespace wouldn't be confusing, as jest uses the same namespace to implement [manual mocks](https://jestjs.io/docs/en/manual-mocks).

We can change the name? Any suggestions? How about `__mock-builders__`? Since we only export functions from this directory and not mock objects.

@mahboubii @jaapbakker88 @ath92

mahboubii:
> > I wonder if adding those test helpers to the `__mocks__` namespace wouldn't be confusing, as jest uses the same namespace to implement [manual mocks](https://jestjs.io/docs/en/manual-mocks).
> 
> We can change the name? Any suggestions? How about `__mock-builders__`? Since we only export functions from this directory and not mock objects.
> 
> @mahboubii @jaapbakker88 @ath92

or `mock-stream` maybe since this mock is only specific to stream client

jaapbakker88:
Let's do `mock-builders`, if we split this out into a library we can call it `mock-stream-chat`

vishalnarkhede:
> Let's do `mock-builders`, if we split this out into a library we can call it `mock-stream-chat`

Sounds good. Will update the PR soon :)

vishalnarkhede:
> Let's do mock-builders, if we split this out into a library we can call it mock-stream-chat

@vicnicius @jaapbakker88 @mahboubii @ath92 done - https://github.com/GetStream/stream-chat-react/pull/224/commits/e5bcbdf4a09d9e4f28d87f3dc35804b1ccdd8252

PR #226, opened 2020-04-30
CRS-233 read count tooltip

ath92:
# Submit a pull request

## CLA


ath92:
Just added that. Does it make sense to separate that `formatArray`. function out so we don't have to do double maintenance?

jaapbakker88:
> Just added that. Does it make sense to separate that `formatArray`. function out so we don't have to do double maintenance?

yeah, that makes total sense, good function to put in utils.js maybe, also naming could be a bit more descriptive then.

ath92:
Alright, just pushed that too. Can you check if naming etc. makes sense?

PR #229, opened 2020-05-01
Add MessageDeleted prop

ath92:
# Submit a pull request

## CLA

Adds `MessageDeleted` prop to all Message components. If left undefined, what happens if a message has been deleted is left up to the component itself. This means that for `MessageSimple` and `MessageCommerce` a default "message deleted" is rendered, whereas for the other Message components nothing is shown.

### DIFF
--- a/src/components/Message.js
+++ b/src/components/Message.js
@@ -126,7 +126,7 @@ class Message extends Component {
      * The component that will be rendered if the message has been deleted.
      * All props are passed into this component.
      */
-    MessageDeleted: PropTypes.func,
+    MessageDeleted: PropTypes.elementType,
   };
 
   static defaultProps = {
--- a/src/components/Message.js
+++ b/src/components/Message.js
@@ -122,6 +122,11 @@ class Message extends Component {
      * Available props - https://getstream.github.io/stream-chat-react/#messageinput
      * */
     additionalMessageInputProps: PropTypes.object,
+    /**
+     * The component that will be rendered if the message has been deleted.
+     * All props are passed into this component.
+     */
+    MessageDeleted: PropTypes.func,
   };
 
   static defaultProps = {
--- a/src/components/MessageCommerce.js
+++ b/src/components/MessageCommerce.js
@@ -8,7 +8,7 @@ import { MessageRepliesCountButton } from './MessageRepliesCountButton';
 
 import PropTypes from 'prop-types';
 
-import { isOnlyEmojis, renderText } from '../utils';
+import { isOnlyEmojis, renderText, smartRender } from '../utils';
 import { withTranslationContext } from '../context';
 
 /**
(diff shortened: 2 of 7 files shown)


src/components/MessageCommerce.js
```diff
@@ -86,10 +86,32 @@ class MessageCommerce extends PureComponent {
...
+          <span
+            key={message.id}
+            className={`${messageClasses} str-chat__message--deleted`}
+          >
+            {t('This message was deleted...')}
+          </span>
+          <div className="clearfix" />
+        </React.Fragment>
+      );
+    },
```
jaapbakker88, line 114: Probably good to split this out into a separate component

src/components/Message.js
```diff
@@ -122,6 +122,11 @@ class Message extends Component {
      * Available props - https://getstream.github.io/stream-chat-react/#messageinput
      * */
     additionalMessageInputProps: PropTypes.object,
+    /**
+     * The component that will be rendered if the message has been deleted.
+     * All props are passed into this component.
+     */
+    MessageDeleted: PropTypes.func,
```
jaapbakker88, line 129: This should be `PropTypes. elementType `

src/components/MessageCommerce.js
```diff
@@ -220,27 +243,14 @@ class MessageCommerce extends PureComponent {
...
     );
 
-    if (
-      message.type === 'message.read' ||
-      message.deleted_at ||
-      message.type === 'message.date'
-    ) {
-      return null;
+    if (message.deleted_at) {
+      return <MessageDeleted {...this.props} />; // MessageDeleted is always defined because it has a default value
```
jaapbakker88, line 247: For these renders you probably want to use this helper function:
https://github.com/GetStream/stream-chat-react/blob/985a7b235156942a51718769adc40e897f71ebc5/src/utils.js#L205

src/components/MessageCommerce.js
ath92: I split this out, even though I realized while doing so that the block of code for the deleted message in `MessageCommerce` was redundant, and it would actually always return null if a message had been deleted in what's currently on master.

For now, I've made `MessageDeleted` a separate component, even though it's only used in `MessageSimple`. I think it makes sense to keep `MessageSimple`. a bit smaller, but let me know if you prefer it to stay inside `MessageSimple`.

src/components/MessageCommerce.js
ath92: That's useful, didn't know that existed in the utils.

ath92:
@jaapbakker88 I think I've addressed all your points, let me know if there are other things that need fixing

src/components/MessageDeleted.js
```diff
@@ -0,0 +1,31 @@
...
+      >
+        <div className="str-chat__message--deleted-inner">
+          {t('This message was deleted...')}
+        </div>
+      </div>
+    );
+  }
+}
+
+export { MessageDeleted };
```
vishalnarkhede, line 31: better to make it consumer of translation context explicitely.

```js
const MessageDeleted = withTranslationContext(MessageDeleted)
export { MessageDeleted }
```

PR #233, opened 2020-05-06
CRS-16 avatar handlers

ath92:
# Submit a pull request

## CLA

This adds `onUserClick` event handler props to the `Message` components.


wmadden:
@ath92 @jaapbakker88 it looks like this code never worked. 

The `Avatar` accepts an `onClick` event handler, but it's never set to anything. `MessageSimple` and the other `Message` variants expect an `onUserClick` prop to be passed to them (by `MessageList`) but it isn't. `ChannelHeader` and `ChannelPreview`, which also use the `Avatar`, similarly never set the `onUserClick` property. And `onUserClick` isn't present in the `ChannelContext`

PR #236, opened 2020-05-06
Tests and Hooks: ChannelList

vishalnarkhede:
# Submit a pull request

## CLA

## Tasks

- [X] Tests for ChannelList, ChannelListMessenger, ChannelListTeam components
- [X] Converted ChannelListMessenger, ChannelListTeam to functional components
- [X] Moved `@testing-library/react` and `react-test-renderer` to devDependencies from dependencies
- [X] Fixed some documentation for ChannelList component.
- [x] Convert `ChannelList` to functional component and start using hooks


src/components/ChannelList/ChannelListMessenger.js
```diff
@@ -10,60 +10,54 @@ import { LoadingChannels } from '../Loading';
...
+   */
+  LoadingIndicator: PropTypes.elementType,
+  /**
+   * Error indicator UI Component. It will be displayed if `error` prop is true
+   *
+   * Defaults to and accepts same props as:
+   * [ChatDown](https://github.com/GetStream/stream-chat-react/blob/master/src/components/ChatDown.js)
+   *
+   */
+  LoadingErrorIndicator: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
vicnicius, line 54: Should this one also be `PropTypes.elementType`?

src/components/ChannelList/ChannelListTeam.js
```diff
@@ -13,94 +13,94 @@ import chevrondown from '../../assets/str-chat__icon-chevron-down.svg';
...
+   */
+  LoadingIndicator: PropTypes.elementType,
+  /**
+   * Error indicator UI Component. It will be displayed if `error` prop is true
+   *
+   * Defaults to and accepts same props as:
+   * [ChatDown](https://github.com/GetStream/stream-chat-react/blob/master/src/components/ChatDown.js)
+   *
+   */
+  LoadingErrorIndicator: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
vicnicius, line 98: Same as above. Should this be `PropTypes.elementType`?

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -74,89 +138,591 @@ describe('ChannelList', () => {
...
+        await waitFor(() => {
+          expect(getByRole('list')).toBeInTheDocument();
+        });
+
+        const latestMessageNode = getByText(message2.text);
+
+        dispatchChannelTruncatedEvent(chatClientVishal, channel1.channel);
+
+        await waitFor(() => {
+          expect(getNodeText(latestMessageNode)).not.toBe(message2.text);
```
vicnicius, line 700:
```suggestion
          expect(latestMessageNode).not.toHaveTextContent(message2.text);
```

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -1,22 +1,30 @@
 import React from 'react';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import axios from 'axios';
-import { cleanup, render, waitFor } from '@testing-library/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { getNodeText } from '@testing-library/dom';
```
vicnicius, line 5: If https://github.com/GetStream/stream-chat-react/pull/236/files#r430243870 works we can get rid of this.

src/components/ChannelList/ChannelList.js
```diff
@@ -486,58 +131,196 @@ class ChannelList extends PureComponent {
...
+   * Available built-in options (also accepts the same props as):
+   *
+   * 1. [LoadMorePaginator](https://github.com/GetStream/stream-chat-react/blob/master/src/components/LoadMorePaginator.js)
+   * 2. [InfiniteScrollPaginator](https://github.com/GetStream/stream-chat-react/blob/master/src/components/InfiniteScrollPaginator.js)
+   */
+  Paginator: PropTypes.elementType,
+  /**
+   * Function that overrides default behaviour when new message is received on channel that is not being watched
+   *
+   * @param {Component} setChannels Setter for channels value in state.
```
vishalnarkhede, line 238: This will be a breaking change. So far we used to pass the `this` reference of component to these event handlers. But with functional component, that won't be possible. So instead we will pass `setChannels` function so channels can be altered from these handlers.

@jaapbakker88 @mahboubii @ath92 @vicnicius any comments?

src/components/ChannelList/__tests__/ChannelList.test.js
```diff
@@ -1,22 +1,37 @@
 import React from 'react';
```
vishalnarkhede, line 1: ### Tests

✓ should re-query channels when filters change (233ms)
✓ should render `LoadingErrorIndicator` when queryChannels api throws error (27ms)
✓ when queryChannels api returns no channels, `EmptyStateIndicator` should be rendered (22ms)
- mobile navigation
    ✓ should call `closeMobileNav` prop function, when clicked outside ChannelList (573ms)
    ✓ should not call `closeMobileNav` prop function on click, if ChannelList is collapsed (226ms)
- Default and custom active channel
    ✓ should call `setActiveChannel` prop function with first channel as param (64ms)
    ✓ should call `setActiveChannel` prop function with channel (which has `customActiveChannel` id)  as param (48ms)
    ✓ should render channel with id `customActiveChannel` at top of the list (217ms)
- Event handling
    - message.new
    ✓ should move channel to top of the list (287ms)
    ✓ should not alter order if `lockChannelOrder` prop is true (360ms)
    - notification.message_new
    ✓ should move channel to top of the list by default (250ms)
    ✓ should call `onMessageNew` function prop, if provided (183ms)
    - notification.added_to_channel
    ✓ should move channel to top of the list by default (308ms)
    ✓ should call `onAddedToChannel` function prop, if provided (269ms)
    - notification.removed_from_channel
    ✓ should remove the channel from list by default (275ms)
    ✓ should call `onRemovedFromChannel` function prop, if provided (307ms)
    - channel.updated
    ✓ should update the channel in list, by default (157ms)
    ✓ should call `onChannelUpdated` function prop, if provided (169ms)
    - channel.deleted
    ✓ should remove channel from list, by default (179ms)
    ✓ should call `onChannelDeleted` function prop, if provided (145ms)
    ✓ should unset activeChannel if it was deleted (150ms)
    - connection.recovered
    ✓ should rerender the list (183ms)
    - channel.truncated
    ✓ should remove latest message (233ms)
    ✓ should call `onChannelTruncated` function prop, if provided (161ms)

src/components/ChannelList/ChannelList.js
jaapbakker88: I think it makes sense, this way. let's just create a minor release when we do.

src/components/ChannelList/ChannelList.js
```diff
@@ -1,543 +1,361 @@
...
-        this.props.onAddedToChannel(this, e);
-      } else {
-        const channel = await this.getChannel(e.channel.type, e.channel.id);
-        this.setState((prevState) => ({
-          channels: uniqBy([channel, ...prevState.channels], 'cid'),
-        }));
-      }
-    }
+  // Set a channel with id {customActiveChannel} as active and move it to the top of the list.
+  // Of customActiveChannel prop is absent, then set the first channel in list as active channel.
```
vicnicius, line 38:
```suggestion
  // If customActiveChannel prop is absent, then set the first channel in list as active channel.
```

src/components/ChannelList/ChannelList.js
```diff
@@ -1,543 +1,361 @@
...
-          );
-          return {
-            channels,
-          };
-        });
-      }
+    if (
+      !channels ||
+      channels.length === 0 ||
+      channels.length > (options.limit || MAX_QUERY_CHANNELS_LIMIT)
```
vicnicius, line 51: Why do we need this condition here?

vishalnarkhede: Its to ensure we execute this only on first page load. But maybe makes sense to move this logic to hook (usePaginatedChannels) itself

src/components/ChannelList/ChannelList.js
```diff
@@ -1,543 +1,361 @@
...
-        channels[channelIndex].data = Immutable(e.channel);
-
-        this.setState({
-          channels: [...channels],
-          channelUpdateCount: this.state.channelUpdateCount + 1,
-        });
+    if (customActiveChannel) {
+      const _customActiveChannel = channels.filter(
+        (channel) => channel.id === customActiveChannel,
+      )[0];
```
vicnicius, line 59:
```suggestion
      const _customActiveChannel = channels.find(
        (channel) => channel.id === customActiveChannel,
      );
```

src/components/ChannelList/ChannelList.js
vicnicius: Got it, thanks. And I agree with you about moving it to the paginated channels hook 👍

src/components/ChannelList/ChannelList.js
```diff
@@ -1,543 +1,361 @@
...
+    props.filters,
+    props.sort,
+    props.options,
+    activeChannelHandler,
+  );
+
+  useMobileNavigation(channelListRef, props.navOpen, props.closeMobileNav);
+
+  // All the event listeners
+  !props.lockChannelOrder && useMessageNewListener(setChannels);
```
vicnicius, line 102: This breaks the first rule of Hooks https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level.

src/components/ChannelList/ChannelList.js
```diff
@@ -1,543 +1,361 @@
...
+  useNotificationMessageNewListener(setChannels, props.onMessageNew);
```
vicnicius, line 103: Not something to be handled on this PR, but looking at the code here I had an idea:
I was wondering if just extracting the listeners instead wouldn't make things a bit more explicit and easier to maintain. Like:
```js
  const channelDeleteListener = useCallback(() => {getChannelDeleteListener(a, b)}, [a, b]);
  const channelUpdatedListener = useCallback(() => {...}, [...]);
  const channelTruncatedListener = useCallback(() => {...}, [...]);
  useEffect(() => {
    client.on('channel.deleted', channelDeleteListener);
    client.on('channel.truncated', channelDeleteListener);
    client.on('channel.updated', channelDeleteListener);
    return () => {
      client.off('channel.deleted');
      client.off('channel.updated');
      client.off('channel.truncated');
    };
  }, [client]);
```
The benefits I see are that the listener functions could be tested in isolation more efficiently as they would be simple functions. We would possibly have a more clear way of understanding the events handled by this component, which is a key part of it.

src/components/ChannelList/hooks/useNotificationRemovedFromChannelListener.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useContext } from 'react';
+import { ChatContext } from '../../../context';
+
+export const useNotificationRemovedFromChannelListener = (
+  setChannels,
+  customHandler,
```
vicnicius, line 6: Without listing this `customHandler` as a dependency on the `useEffect` hook, we make it no longer possible to use dynamic custom handlers. Looking at [the old implementation](https://github.com/GetStream/stream-chat-react/pull/236/files#diff-23769ba2a0759e37502ac29ef02b9795L298) it seems like this was possible before.

vicnicius: The same is true for other listeners in this PR.

src/components/ChannelList/hooks/usePaginatedChannels.js
```diff
@@ -0,0 +1,78 @@
...
+) => {
+  const [channels, setChannels] = useState(Immutable([]));
+  const [loadingChannels, setLoadingChannels] = useState(true);
+  const [refreshing, setRefreshing] = useState(true);
+  const [error, setError] = useState(false);
+  const [offset, setOffset] = useState(0);
+  const [hasNextPage, setHasNextPage] = useState(true);
+
+  const queryChannels = async (queryType) => {
+    // const { setActiveChannelOnMount } = this.props;
```
vicnicius, line 21: Hanging comment 🔪

vicnicius:
Great work here, man! A big one  and looks way better than before.

I made some comments and shared some ideas, but the two key questions I have regard https://github.com/GetStream/stream-chat-react/pull/236#discussion_r434477382 and https://github.com/GetStream/stream-chat-react/pull/236#discussion_r434445168.

src/components/ChannelList/ChannelList.js
vishalnarkhede: Sounds good!! Lets handle it separately :)

PR #239, opened 2020-05-07
Unset channel on delete if it is active

vicnicius:

Unset channel on delete if the deleted channel was the active one and there's no custom `onChannelDelete` handler.

### DIFF
--- a/src/components/ChannelList/ChannelList.js
+++ b/src/components/ChannelList/ChannelList.js
@@ -416,6 +416,11 @@ class ChannelList extends PureComponent {
           channels: [...channels],
           channelUpdateCount: this.state.channelUpdateCount + 1,
         });
+
+        // Unset active channel if it was the deleted one
+        if (e.channel.cid === this.props.channel.cid) {
+          this.props.setActiveChannel({});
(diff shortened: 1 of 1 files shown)


src/components/ChannelList/ChannelList.js
```diff
@@ -416,6 +416,11 @@ class ChannelList extends PureComponent {
           channels: [...channels],
           channelUpdateCount: this.state.channelUpdateCount + 1,
         });
+
+        // Unset active channel if it was the deleted one
+        if (e.channel.cid === this.props.channel.cid) {
+          this.props.setActiveChannel(undefined);
```
mahboubii, line 422: I think it's best to use `{}` instead of undefined, had this issue before that some components might not check for undef before using the object properties

vicnicius: 👍 [fixed](https://github.com/GetStream/stream-chat-react/pull/239/commits/35c832cbfae704c61a010523b8d37cd25a54e24c#diff-23769ba2a0759e37502ac29ef02b9795R422), thanks.

PR #240, opened 2020-05-07
CRS-242 Thread layout rework

ath92:
# Submit a pull request

## CLA

This commit makes the whole thread scrollable, instead of having the first message stick to the top of the thread.

It also fixes some spacing issues around the not-first messages in the thread (there was too much padding) and the "Start of a new thread" separator (too much / too little padding)

### DIFF
--- a/src/components/Thread/Thread.js
+++ b/src/components/Thread/Thread.js
@@ -211,13 +211,13 @@ class ThreadInner extends React.PureComponent {
             Message={this.props.Message}
             {...this.props.additionalMessageListProps}
           />
-          {smartRender(this.props.MessageInput, {
-            MessageInputSmall,
-            parent: this.props.thread,
-            focus: this.props.autoFocus,
-            ...this.props.additionalMessageInputProps,
-          })}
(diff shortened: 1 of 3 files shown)


src/components/Thread/Thread.js
```diff
@@ -211,13 +211,13 @@ class ThreadInner extends React.PureComponent {
...
-            focus: this.props.autoFocus,
-            ...this.props.additionalMessageInputProps,
-          })}
         </div>
+        {smartRender(this.props.MessageInput, {
+          MessageInputSmall,
+          parent: this.props.thread,
+          focus: this.props.autoFocus,
+          ...this.props.additionalMessageInputProps,
+        })}
```
mahboubii, line 220: whitespace change?

ath92: The message input was moved outside of the parent div. That ways it doesn't scroll with the rest of the thread

mahboubii: oh right! been staring at screen for so long

PR #244, opened 2020-05-07
Tests/command item

mahboubii:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/CommandItem/__tests__/CommandItem.test.js
+++ b/src/components/CommandItem/__tests__/CommandItem.test.js
@@ -1,0 +1,65 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { cleanup, render, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import CommandItem from '../CommandItem';
+
+afterEach(cleanup); // eslint-disable-line
+
+describe('commandItem', () => {
+  it('should render component with empty entity', () => {
+    const tree = renderer.create(<CommandItem entity={{}} />).toJSON();
+    expect(tree).toMatchInlineSnapshot(`
+      <div
+        className="str-chat__slash-command"
+      >
+        <span
+          className="str-chat__slash-command-header"
+        >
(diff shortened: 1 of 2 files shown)


src/components/CommandItem/__tests__/CommandItem.test.js
```diff
@@ -0,0 +1,65 @@
...
+import '@testing-library/jest-dom';
+
+import CommandItem from '../CommandItem';
+
+afterEach(cleanup); // eslint-disable-line
+
+describe('commandItem', () => {
+  it('should render component with empty entity', () => {
+    const tree = renderer.create(<CommandItem entity={{}} />).toJSON();
+    expect(tree).toMatchInlineSnapshot(`
```
vishalnarkhede, line 13: Can we keep snapshots in separate files? similar to this - https://github.com/GetStream/stream-chat-react/pull/236/files#diff-9f8593107b8d0b64260cd6c820bef430R39

mahboubii: they suggest using inline snapshots whenever the snapshot is short but fine with me if you think it's better to put them in a separate file 
https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-inline-snapshots.md

src/components/CommandItem/__tests__/CommandItem.test.js
```diff
@@ -0,0 +1,65 @@
...
+      </div>
+    `);
+  });
+
+  it('should render component with custom entity prop', async () => {
+    const entity = { name: 'name', args: 'args', description: 'description' };
+    const Component = <CommandItem entity={entity} />;
+
+    const { getByText } = render(Component);
+    await waitFor(() => {
```
ath92, line 36: Do you need the waitFor here? I was under the impression that `render` is synchronous

mahboubii: you're right it's not needed

PR #250, opened 2020-05-07
CRS-244 Fix Message UI component type definitions

ath92:
# Submit a pull request

## CLA

Currently the typescript definitions (i.e. interfaces) for Message components are set up as follows:

 - MessageProps, which contains all the props for the high level Message component
 - MessageUIComponentProps, which extends MessageProps, and adds some props specific to the UI components.

However, the high level Message component receives several props which the UI components ignore (and which aren’t defined in their PropTypes definitions). I refactored it so it accurately reflects the actual React components.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -459,13 +459,12 @@ export interface AttachmentUIComponentProps {
   ): void;
 }
 
+// MessageProps are all props shared between the Message component and the Message UI components (e.g. MessageSimple)
 export interface MessageProps extends TranslationContextValue {
   /** The message object */
   message?: Client.MessageResponse;
   /** The client connection object for connecting to Stream */
   client?: Client.StreamChat;
-  /** The current channel this message is displayed in */
-  channel?: Client.Channel;
   /** A list of users that have read this message **/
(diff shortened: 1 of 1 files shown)


jaapbakker88:
@vishalnarkhede @ath92 are we sure we can delete all these types, we are passing them all to the message.

ath92:
Hmm, I see we're indeed passing everything down. My reasoning was that if the actual Message UI components (e.g. `MessageSimple`) don't use some of the props for `Message`, and they also don't have them in their PropTypes, then the typescript definitions also shouldn't contain those props.

Note that the definitions for `Message` have only changed in the sense that they no longer contain some of the props that only applied to the UI components (e.g. `onMentionsClickMessage` is now only defined for the UI components, but `onMentionsClick` is still part of the props for the `Message` component).

So essentially I just split things out so that the UI components don't just get all the props that are defined on the high level component, because I think it was not accurate.

jaapbakker88:
sounds good to me, @vishalnarkhede can you also take a look at this?

vishalnarkhede:
Will give it a look in some time!!

PR #256, opened 2020-05-11
CRS-157 Add tooltip snapshot test (for completeness)

ath92:
# Submit a pull request

## CLA

Just a very simple snapshot test, so this component i covered.

### DIFF
--- a/src/components/Tooltip/__tests__/Tooltip.test.js
+++ b/src/components/Tooltip/__tests__/Tooltip.test.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import Tooltip from '../Tooltip';
+
+describe('Tooltip', () => {
+  it('should render as expected', () => {
+    const tree = renderer.create(<Tooltip></Tooltip>).toJSON();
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/__tests__/Tooltip.test.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import Tooltip from '../Tooltip';
+
+describe('Tooltip', () => {
+  it('should render as expected', () => {
+    const tree = renderer.create(<Tooltip></Tooltip>).toJSON();
```
vicnicius, line 8: Quite a small component, but what do you think of adding a child to the component, so we see it wrapping it?

```suggestion
    const tree = renderer.create(<Tooltip>tooltip text</Tooltip>).toJSON();
```

ath92: Yeah makes sense :) will add

vicnicius:
We could also squash those two commits here, to keep history cleaner.

PR #270, opened 2020-05-14
CRS-139 MessageInput tests

ath92:
# Submit a pull request

## CLA

Note that these tests don't really try to mock out child components. This means that some of the functionality in these tests relies on the implementation of its children. With that said, I tried to stick to a declarative style as much as possible, so that the actual implementation of those children doesn't really matter, as long as they do what is expected for MessageInput to work properly. In that sense it follows what's in this article: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering/

### DIFF
--- a/src/components/MessageInput/__tests__/MessageInput.test.js
+++ b/src/components/MessageInput/__tests__/MessageInput.test.js
@@ -1,15 +1,8 @@
 import React from 'react';
-import {
-  cleanup,
-  render,
-  waitFor,
-  fireEvent,
-  findAllByPlaceholderText,
-} from '@testing-library/react';
+import { cleanup, render, waitFor, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import axios from 'axios';
 import MessageInput from '../MessageInput';
-import MessageInputLarge from '../MessageInputLarge';
 import { Chat } from '../../Chat';
 import { Channel } from '../../Channel';
 import {
@@ -28,6 +21,9 @@ let chatClient, channel;
 // mock i18n
 const t = (t) => t;
 
+// silence console.warn because some components log large warnings in catch statements
+console.warn = () => {};
+
 // MessageInput components rely on ChannelContext.
 // ChannelContext is created by Channel component,
 // Which relies on ChatContext, created by Chat component.
@@ -42,9 +38,12 @@ const renderComponent = (props = {}) =>
 
 describe('MessageInput', () => {
   const inputPlaceholder = 'Type your message';
+  const username = 'username';
+  const userid = 'userid';
+
   // First, set up a client and channel, so we can properly set up the context etc.
   beforeAll(async () => {
-    const user1 = generateUser();
+    const user1 = generateUser({ name: username, id: userid });
     const message1 = generateMessage({ user: user1 });
     const mockedChannel = generateChannel({
       messages: [message1],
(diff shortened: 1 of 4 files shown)


src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -0,0 +1,342 @@
...
+    const fileUploadUrl = 'http://www.getstream.io'; // real url, because ImagePreview will try to load the image
+
+    const mockUploadApi = () =>
+      jest.fn().mockResolvedValue(
+        Promise.resolve({
+          file: fileUploadUrl,
+        }),
+      );
+
+    // const mockFaultyUploadApi = () => jest.fn().mockResolvedValue(
```
vicnicius: 🔪

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -0,0 +1,342 @@
...
+      const formElement = await findByPlaceholderText(inputPlaceholder);
+      dropFile(
+        new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' }),
+        formElement,
+      );
+
+      await waitFor(() => expect(doImageUploadRequest).toHaveBeenCalled());
+    });
+
+    // it('Should call error handler if an image failed to upload', async () => {
```
vicnicius: Should we uncomment this one?

src/components/MessageInput/__tests__/MessageInput.test.js
vicnicius: It was removed on a later commit. 👍

src/components/MessageInput/__tests__/MessageInput.test.js
vicnicius: It was uncommented on a later commit. :+1:

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -0,0 +1,479 @@
...
+} from '../../../mock-builders';
+
+jest.mock('axios');
+let chatClient, channel;
+
+// mock i18n
+const t = (t) => t;
+
+// silence console.warn because some components log large warnings in catch statements
+console.warn = () => {};
```
vicnicius, line 25: I think it is better to do it inside the test statements that output those large warnings. E.g: `jest.spyOn(console, 'warn').mockImplementationOnce(() => null);` This way we prevent ignoring warns that would be relevant.

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -0,0 +1,479 @@
...
+          expect.any(Object),
+        );
+        expect(doFileUploadRequest).toHaveBeenCalledWith(
+          file,
+          expect.any(Object),
+        );
+      });
+    });
+
+    // TODO: Check if pasting plaintext is not prevented -> tricky because recreating exact event is hard
```
vicnicius, line 292: Minor: we could use `it.todo('Check if...');` here instead. The nice thing is that it shows on the test report:
image

ath92: Ah I had no idea this existed, will add :)

src/components/MessageInput/__tests__/MessageInput.test.js
ath92: Totally agree, will fix

src/components/MessageInput/__tests__/MessageInput.test.js
ath92: Also, agreed about the errors. Those actually don't show up locally for me, so I didn't realise they were in the Travis console output.

I'll look into the best way to get rid of them. One way would be to mock out the `ImagePreview` component (like we discussed). That introduces other difficulties though, because as far as I could tell I could only mock the entire module, which means I would have to reimplement certain things like the dropzone in order to get some of these tests to work. 

Maybe I can just mock out the child-child dependency that actually does the image loading though. I'll investigate a bit more :)

ath92:
@vicnicius I just pushed some changes that get rid of the console errors, and clean up the console warn suppression. I think that's all, right?

Issue #212, opened 2020-04-16
Feature request: Chat NavOpen prop


JJ-Kidd:
Can you please add the ability to accept navOpen as a prop on the Chat component. By default this is set to true in the state. Overriding it was a little fiddly and not very clean, it would be much nicer if I could just override this by default instead of managing my own state and setting a timeout. Please see an example of what I have and my suggestion below

```javascript
const [displayNav, setDisplayNav] = useState(false);

<Chat client={chatClient} theme="messaging light">
      <ChannelList
        navOpen={displayNav}
        closeMobileNav={() => {
          setDisplayNav(false);
        }}
        userId={userId}
        filters={directFilter}
      />
      <Channel channel={channel}>
        <Window>
          <ChannelHeader
            openMobileNav={() => {
              setTimeout(() => {
                setDisplayNav(true);
              }, 100);
            }}
          />
          <MessageList />
          <MessageInput />
        </Window>
        <Thread />
      </Channel>
    </Chat>
```

```javascript
<Chat client={chatClient} theme="messaging light" navOpen={false}>
 ...
</Chat>
```

PR #284, opened 2020-05-19
added custom navOpen prop

mksglu:
To say the truth about this, I don't think a custom prop will be needed. If we do, I think this would be like this. #212


mahboubii:
fixed in https://github.com/GetStream/stream-chat-react/pull/306

PR #286, opened 2020-05-20
CRS-216 MessageInput

ath92:
MessageInput using hooks

Still missing:
 - TS definitions need updating + comments need to be added
 - Test branch was merged into this, so will need to clean up git history later


vicnicius:
Except for some types still failing, LGTM 👍

src/components/MessageInput/MessageInputState.js
```diff
@@ -0,0 +1,700 @@
...
+        fileUploads: state.fileUploads.without(action.id),
+      };
+    case 'reduceNumberOfUploads': // TODO: figure out if we can just use uploadOrder instead
+      return { ...state, numberOfUploads: state.numberOfUploads - 1 };
+    case 'addMentionedUser':
+      return {
+        ...state,
+        mentioned_users: state.mentioned_users.concat(action.userId),
+      };
+    default:
```
vicnicius, line 217: Maybe we could add a warning if the reducer defaults.

ath92:
Yeah the types failing should be resolved once my commit to the js client is merged :)

src/components/MessageInput/MessageInputState.js
```diff
@@ -0,0 +1,700 @@
...
+ *  type: 'reduceNumberOfUploads',
+ * }} ReduceNumberOfUploadsAction
+ * @typedef {{
+ *  type: 'addMentionedUser',
+ *  userId: string,
+ * }} AddMentionedUserAction
+ *
+ * MessageInput state reducer
+ * @param {State} state
+ * @param {SetEmojiPickerIsOpenAction | SetTextAction | ClearAction | SetImageUploadAction | SetFileUploadAction | RemoveImageUploadAction | RemoveFileUploadAction | ReduceNumberOfUploadsAction | AddMentionedUserAction} action
```
vicnicius, line 147: Maybe define a separate type for those actions and use it here:
```typescript
/**
 * @typedef {
 * | SetEmojiPickerIsOpenAction
 * | SetTextAction
 * | ClearAction
 * | SetImageUploadAction
 * | SetFileUploadAction
 * | RemoveImageUploadAction
 * | RemoveFileUploadAction
 * | ReduceNumberOfUploadsAction
 * | AddMentionedUserAction
 * } MessageInputStateAction
 *
 * MessageInput state reducer
 * @param {State} state
 * @param {MessageInputStateAction} action
 * @returns {State}
 */
```

src/components/MessageInput/MessageInputTypes.ts
```diff
@@ -0,0 +1,52 @@
+type SetEmojiPickerIsOpenAction = {
```
vicnicius, line 1: I definitely think that it is way better to extract the types to their own files like this. 👍 I only wonder if we should go with one per component or if we could have one type file per "namespace": Like `src/components/MessageInput/types.ts`. But that's something we can discuss and adjust along the way.

ath92: Ah yes that also makes sense, I wasn't really sure what to call this so if we just go with types.ts, I won't have to care about naming anymore :)

PR #303, opened 2020-05-28
Fix some typos

ferhatelmas:
While checking a SO question, saw some typos.


ferhatelmas:
BTW, type problem is on master.

PR #307, opened 2020-05-29
Add tests and improve <Message /> component

vicnicius:

- Add tests to the <Message /> component.

Depends on: https://github.com/GetStream/stream-chat-js/pull/344


vishalnarkhede:
Is it ready for review?

vicnicius:
> Is it ready for review?

Yeah... it can be reviewed, but types will only be green after https://github.com/GetStream/stream-chat-js/pull/344 is released.

PR #309, opened 2020-06-02
CRS-256 mentioned users

ath92:
# Submit a pull request

## CLA

This allows users to remove mentions by just getting rid of (part of) the text in the input.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -445,7 +445,7 @@ export interface MessageInputState {
   fileUploads: SeamlessImmutable.ImmutableObject<{ [id: string]: FileUpload }>;
   emojiPickerIsOpen: boolean;
   // ids of users mentioned in message
-  mentioned_users: string[];
+  mentioned_users: { id: string; name: string }[];
   numberOfUploads: number;
 }
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -445,7 +445,7 @@ export interface MessageInputState {
   fileUploads: SeamlessImmutable.ImmutableObject<{ [id: string]: FileUpload }>;
   emojiPickerIsOpen: boolean;
   // ids of users mentioned in message
-  mentioned_users: { id: string; name: string }[];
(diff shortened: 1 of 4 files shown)


types/index.d.ts
```diff
@@ -106,7 +106,6 @@ export interface ChatProps {
   // | 'livestream dark'
   theme?: string;
   i18nInstance?: Streami18n;
-  initialNavOpen?: boolean;
```
ath92, line 109: Didn't mean to remove this, will fix

types/index.d.ts
```diff
@@ -445,7 +445,7 @@ export interface MessageInputState {
   fileUploads: SeamlessImmutable.ImmutableObject<{ [id: string]: FileUpload }>;
   emojiPickerIsOpen: boolean;
   // ids of users mentioned in message
-  mentioned_users: string[];
```
jaapbakker88, line 448: I think name should be an optional field. The only required field on a user is `id`. Also true for other areas. Probably want to check if a user has a name, then use it otherwise use id.

ath92: Hmm, I see, thanks :). I didn't realise name is an optional field.

I think I might change the type of that state variable to `Client.UserResponse[]`, so it's consistent with the Message object, and it makes it easier to check if a user is still mentioned, because I'll have access to both name and id.

jaapbakker88: Sounds good!

vicnicius: I was looking into something that incidentally touches this, and I think we get `Client.User[]`, and not `Client.UserResponse[]` there, but I'm not sure. I wasn't looking into that field specifically. It would be nice to double-check, though.

ath92: I double checked, it seems like we get `UserResponse`, both from the parent (as part of the `message` prop), and from the actual input, as part of `ChatAutoCompleteProps` which contains `users: Client.UserResponse[];`.

I just added the changes, should be all good now.

PR #315, opened 2020-06-05
Tests/CRS-99/event component

mahboubii:


src/context/TranslationContext.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
 import Dayjs from 'dayjs';
+import LocalizedFormat from 'dayjs/plugin/localizedFormat';
+
+const extendedDayjs = Dayjs.extend(LocalizedFormat);
```
vishalnarkhede, line 5: It should be in `Streami18n.js` I think!! @mahboubii

mahboubii: yea it is there if you wrap the component with translation context, but the context default value here is not supporting the localized format, causing issues when we use LT format in components

PR #318, opened 2020-06-05
CRS-60 infinite scroll

ath92:
# Submit a pull request

## CLA

This converts the InfinteScroll component to hooks. It also merges the implementation ReverseInfiniteScroll into InfiniteScroll, so there's no more code duplication.


src/components/InfiniteScrollPaginator/InfiniteScroll.js
```diff
@@ -1,232 +1,156 @@
...
-      }
+      loadMore();
     }
-  }
+  }, [useWindow, isReverse, threshold, listenToScroll, loadMore]);
 
-  calculateOffset(el, scrollTop) {
-    if (!el) {
-      return 0;
+  useEffect(() => {
```
vicnicius, line 95: On this `useEffect` hook we don't list dependencies, but on the following one, we do. Shouldn't we do it for both?

src/components/InfiniteScrollPaginator/InfiniteScroll.js
```diff
@@ -1,232 +1,156 @@
...
-      (el.offsetHeight - scrollTop - window.innerHeight)
-    );
-  }
+    scrollEl.addEventListener('scroll', scrollListener, useCapture);
+    scrollEl.addEventListener('resize', scrollListener, useCapture);
 
-  calculateTopPosition(el) {
-    if (!el) {
-      return 0;
+    if (initialLoad) {
```
vicnicius, line 104: Could we get rid of the `initialLoad` prop (by adding an extra `useEffect(scrollListener)`)? I can't think of a reason for exposing this functionality.

src/components/MessageList/MessageList.js
```diff
@@ -786,7 +786,8 @@ class MessageList extends PureComponent {
           {!elements.length ? (
             <EmptyStateIndicator listType="message" />
           ) : (
-            <ReverseInfiniteScroll
```
vicnicius, line 789: We could still use the ReverseInfiniteScroll, couldn't we?

types/index.d.ts
```diff
@@ -786,6 +786,8 @@ export interface InfiniteScrollProps {
   element?: React.ElementType;
   loader?: React.ReactNode;
   threshold?: number;
+  listenToScroll?(standardOffset: number, reverseOffset: number): any;
```
vicnicius, line 789:
```suggestion
  listenToScroll?(standardOffset: number, reverseOffset: number): void;
```
I see different definitions for the return of this callback in different places. I think we could settle with `void`. What you think?

PR #326, opened 2020-06-08
Message input tests

ath92:
# Submit a pull request

## CLA

What's in this pr:
 - Got rid of small differences between MessageInput presentation components:
   - Added some <title> elements to svgs
   - Got rid of double emoji picker in EditMessageForm
   - 'Pick your emoji' text now uses i18n in all components
 - Moved EditMessageForm into MessageInput folder because it is almost exactly the same as the other MessageInputs, aside from some minor css/markup differences.
 - MessageInput tests are now reused for all presentation components. In most cases (MessageInputFlat, MessageInputSmall, EditMessageForm) this should cover (almost) all actual user interactions.
 - Add tests for MessageInputLarge typing string


src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -31,439 +36,440 @@ const t = (key) => key;
...
+      const user1 = generateUser({ name: username, id: userid });
+      const message1 = generateMessage({ user: user1 });
+      const mockedChannel = generateChannel({
+        messages: [message1],
+        members: [generateMember({ user: user1 })],
+      });
+      useMockedApis(axios, [getOrCreateChannelApi(mockedChannel)]);
+      chatClient = await getTestClientWithUser({ id: user1.id });
+      channel = chatClient.channel('messaging', mockedChannel.id);
+      // channel.editMessage = jest.fn();
```
vicnicius, line 82: 🔪

src/components/EditMessageForm/EditMessageForm.js
```diff
@@ -150,38 +144,20 @@ class EditMessageForm extends React.Component {
...
-              ? true
-              : false
-          }
+          disabled={this.props.numberOfUploads >= this.props.maxNumberOfFiles}
           handleFiles={this.props.uploadNewFiles}
         >
           <form onSubmit={this.props.handleSubmit}>
             {this.renderEmojiPicker()}
             {this.renderUploads()}
-            {this.props.emojiPickerIsOpen && (
```
vicnicius, line 163: No style side-effects because of the order those components are rendered?

src/components/EditMessageForm/EditMessageForm.js
```diff
@@ -123,24 +119,22 @@ class EditMessageForm extends React.Component {
...
+        className="str-chat__small-message-input-emojipicker"
+        ref={this.props.emojiPickerRef}
+      >
+        <Picker
+          native
+          emoji="point_up"
+          title={t('Pick your emoji')}
+          onSelect={this.props.onSelectEmoji}
+          color="#006CFF"
+          showPreview={false}
```
vicnicius, line 133: What does this property do? I see we didn't have this set on the second emoji picker we had before, but we do here.

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -31,439 +36,440 @@ const t = (key) => key;
...
+      fireEvent.click(emojiIcon);
 
-    expect(queryByText('Pick your emoji…')).toBeInTheDocument();
+      expect(queryByText('Pick your emoji')).toBeInTheDocument();
 
-    const emoji = '💯';
-    const emojiButton = queryByText(emoji);
-    expect(emojiButton).toBeInTheDocument();
+      const emoji = '💯';
+      const emojiButton = queryAllByText(emoji)[0];
```
vicnicius, line 140: Even after https://github.com/GetStream/stream-chat-react/pull/326/commits/9877314e03a2dd066edd7c564c4236fdad7edb0b#diff-4c02f7c9b62e8fcc93a2a781681a4b82L163, can we still have more than one of this emoji on the dom?

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -31,439 +36,440 @@ const t = (key) => key;
...
+            ]),
+          }),
+        );
       });
-      const submitButton = await findByTitle('Send');
 
-      const formElement = await findByPlaceholderText(inputPlaceholder);
-      const file = new File(['Message in a bottle'], 'the-police.mp3', {
-        type: 'audio/mp3',
+      it('should an audio attachment if an audio file is dropped into the input', async () => {
```
vicnicius, line 436:
```suggestion
      it('should upload an audio attachment if an audio file is dropped into the input', async () => {
```

src/components/MessageInput/EditMessageForm.js
```diff
@@ -150,38 +144,20 @@ class EditMessageForm extends React.Component {
...
-              ? true
-              : false
-          }
+          disabled={this.props.numberOfUploads >= this.props.maxNumberOfFiles}
           handleFiles={this.props.uploadNewFiles}
         >
           <form onSubmit={this.props.handleSubmit}>
             {this.renderEmojiPicker()}
             {this.renderUploads()}
-            {this.props.emojiPickerIsOpen && (
```
mahboubii, line 163: does this mean we were rendering emojiPicker twice?

src/components/MessageInput/MessageInputFlat.js
```diff
@@ -196,14 +197,18 @@ class MessageInputFlat extends PureComponent {
                     height="28"
                     xmlns="http://www.w3.org/2000/svg"
                   >
+                    <title>{t('Open emoji picker')}</title>
```
mahboubii, line 200: making sure these new texts are included in the i18n translation files

src/components/MessageInput/MessageInputLarge.js
```diff
@@ -171,7 +172,7 @@ class MessageInputLarge extends PureComponent {
       //example: "bob, joe, and sam"
       outStr = t('{{ commaSeparatedUsers }} and {{ lastUser }} are typing...', {
         commaSeparatedUsers: arr3.slice(0, -1).join(', '),
-        lastUser: arr3.slice(-1),
+        lastUser: arr3.slice(-1)[0],
```
mahboubii, line 175:
```suggestion
        lastUser: arr3[arr3.length - 1],
```

mahboubii: or even better if you deconstruct the array using `[firstActor, ...commaSeparatedUsers] = arr3` and remove the slice parts

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -25,445 +29,441 @@ jest.mock('axios');
...
-    chatClient = await getTestClientWithUser({ id: user1.id });
-    channel = chatClient.channel('messaging', mockedChannel.id);
-    // channel.editMessage = jest.fn();
-  });
+[
+  { InputComponent: MessageInputLarge, name: 'MessageInputLarge' },
+  { InputComponent: MessageInputSmall, name: 'MessageInputSmall' },
+  { InputComponent: MessageInputFlat, name: 'MessageInputFlat' },
+  { InputComponent: EditMessageForm, name: 'EditMessageForm' },
+].forEach(({ InputComponent, name }) => {
```
mahboubii, line 40: great job unifying all these! 👍

src/components/EditMessageForm/EditMessageForm.js
ath92: The emoji-picker component we use has an option to show a slightly larger preview of the emoji with its name while you hover it. I don't think it really adds value, and we don't use it for the other MessageInput components. My guess is that it's just some leftover (like the double picker). I think this makes things a little bit more consistent across the board.

src/components/EditMessageForm/EditMessageForm.js
ath92: I double checked: doesn't look like it, the emoji picker is always positioned outside of the rest of the EditMessageForm, the order in which it's rendered doesn't change anything here. (it uses absolute positioning with a z-index)

src/components/MessageInput/__tests__/MessageInput.test.js
ath92: Good point, forgot to change it back to what it was :)

src/components/MessageInput/EditMessageForm.js
ath92: Yep :)

src/components/MessageInput/MessageInputFlat.js
ath92: They are, they just weren't used in all the MessageInput components (this makes it more consistent)

src/components/MessageInput/MessageInputLarge.js
ath92: Agreed. I refactored this whole function to make it more readable. (but I think I'll make a separate PR for that, since it also includes all of the conversion to hooks, etc.)

src/components/MessageInput/__tests__/MessageInput.test.js
vicnicius: Oh, I see. Mocking localStorage just because of this seems a bit too much. I think it is fine to leave it as is. Or choose another emoji (I imagine this one is part of a default frequently used set 🤔 ).

ath92: I actually picked this emoji because it isn't part of the standard frequently used set :P That worked fine when I ran all of these tests once, but now that they're repeated for all UI components the localstorage thing is messing things up again :)

PR #332, opened 2020-06-11
WIP: message components hooks

vicnicius:
# Submit a pull request

⚠️ WIP ⚠️ 

Refactoring all message components.

**I will split this into smaller PRs, but keeping it here for a big picture perspective.**

PR #336, opened 2020-06-15
Task/message components refactor part ii

vicnicius:

~To be merged after https://github.com/GetStream/stream-chat-react/pull/335~

This adds a set of custom hooks that will be used by the message UI components when transforming them in Functional components. They're abstractions of functionality that now resides in the message HOC.

More context on this change here: https://github.com/GetStream/stream-chat-react/pull/332.


src/components/Message/hooks/useHandleAction.js
```diff
@@ -0,0 +1,38 @@
...
+   *@type {import('types').ChannelContextValue}
+   */
+  const { channel, updateMessage, removeMessage } = useContext(ChannelContext);
+  return async (name, value, event) => {
+    event.preventDefault();
+    /**
+     *@type {import('types').ChannelContextValue}
+     */
+    if (!message || !updateMessage || !removeMessage || !channel) {
+      console.log(!!message, !!updateMessage, !!removeMessage, !!channel);
```
vishalnarkhede, line 21: Do we need this log statement? Its just bunch of true/false, doesn't say much!

src/components/Message/hooks/useHandleMute.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @typedef {{
+ *   notify?: import('types').MessageComponentProps['addNotification'],
+ *   getSuccessNotification?: import('types').MessageComponentProps['getMuteUserSuccessNotification'],
+ *   getErrorNotification?: import('types').MessageComponentProps['getMuteUserErrorNotification'],
+ * }} NotificationArg
+ * @type {(message: import('stream-chat').MessageResponse, notification: NotificationArg) => (event: React.MouseEvent<HTMLElement>) => Promise<void>}
+ */
+export const useHandleMute = (message, notifications) => {
```
vishalnarkhede, line 14: maybe `useMuteHandler` makes more sense? What do you think?

src/components/Message/hooks/useHandleMute.js
```diff
@@ -0,0 +1,89 @@
...
+      getSuccessNotification,
+      getErrorNotification,
+    } = notifications;
+
+    if (!t || !message || !notify || !client) {
+      return;
+    }
+    if (!isUserMuted(message, mutes)) {
+      try {
+        if (!message.user) {
```
vishalnarkhede, line 39: This condition can be put before parent if-else!! And thus can be removed from `else` as well

src/components/Message/hooks/useReactionHandler.js
```diff
@@ -0,0 +1,79 @@
...
+          );
+        }
+      });
+    }
+
+    const originalMessage = message;
+    let reactionChangePromise;
+
+    /*
+    - Add the reaction to the local state
```
vishalnarkhede, line 51: Lets remove this particular line, since we are not adding it to local state before API call.

src/components/Message/hooks/useHandleMute.js
vicnicius: Makes total sense 👍

src/components/Message/hooks/useHandleMute.js
vicnicius: 👌 I moved it to the first check.

src/components/Message/hooks/useHandleAction.js
vicnicius: Some debugging leftover x)

PR #338, opened 2020-06-15
<MessageDeleted /> hooks

vicnicius:
# Submit a pull request

~To be merged after https://github.com/GetStream/stream-chat-react/pull/336.~

Making MessageDeleted a functional component.

### DIFF
--- a/src/components/Message/MessageDeleted.js
+++ b/src/components/Message/MessageDeleted.js
@@ -1,38 +1,45 @@
 // @ts-check
-import React, { PureComponent } from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
-import { withTranslationContext } from '../../context';
+import { TranslationContext } from '../../context';
+import { useUserRole } from './hooks/useUserRole';
+import { MessagePropTypes } from './utils';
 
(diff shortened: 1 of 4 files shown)


src/components/Message/MessageDeleted.js
```diff
@@ -1,38 +1,38 @@
...
-  }
-}
+    </div>
+  );
+};
 
-export default withTranslationContext(MessageDeleted);
+MessageDeleted.propTypes = {
+  /** The [message object](https://getstream.io/chat/docs/#message_format) */
+  // @ts-ignore
```
vicnicius, line 32: I'll try to make this one work without ignoring it

jaapbakker88: what's the error you're getting from this one?

vicnicius: Typescript was matching its definition of the message object against the more loose `propType` one and complaining about it. Using a [more strict](https://github.com/GetStream/stream-chat-react/pull/338/files#diff-91e6dd3754c70f9237d1963e304a426bR110) definition fixed it. I amended the commit with the fix.

PR #342, opened 2020-06-15
CRS-191: add types to ChatDown component

jaapbakker88:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/ChatDown/ChatDown.js
+++ b/src/components/ChatDown/ChatDown.js
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types';
 import { LoadingChannels } from '../Loading';
 import { TranslationContext } from '../../context';
 
-// @ts-ignore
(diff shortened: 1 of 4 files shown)


src/components/ChatDown/ChatDown.js
```diff
@@ -1,17 +1,20 @@
+// @ts-check
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 
 import { LoadingChannels } from '../Loading';
 import { TranslationContext } from '../../context';
 
+// @ts-ignore
```
vicnicius, line 8: We can add something like
```typescript
declare module "*.png" {
   const value: string;
   export = value;
}
```
to our type definitions to avoid having to ignore asset imports.

jaapbakker88: ah yeah, nice. adding..

PR #351, opened 2020-06-16
CRS-65: Loading Components tests/ts/hooks

jaapbakker88:
# Submit a pull request

## CLA


src/components/Loading/LoadingErrorIndicator.js
```diff
@@ -25,7 +29,7 @@ LoadingErrorIndicator.propTypes = {
   error: PropTypes.oneOfType([
     PropTypes.shape({ message: PropTypes.string }),
     PropTypes.bool,
```
vishalnarkhede, line 31: You can get rid of this, if you fix this `setError` statement - https://github.com/GetStream/stream-chat-react/blob/master/src/components/ChannelList/hooks/usePaginatedChannels.js#L56 This is the only place where we set error!!

jaapbakker88: let's do this after, too annoying to refactor right now @vishalnarkhede

PR #360, opened 2020-06-17
Make <MessageSimple /> a functional component

vicnicius:
# Submit a pull request

Refactors the <MessageSimple /> component. Creates some new custom hooks.

### To-do:

- [x] Check component API diffs for breaking changes
- [x] QA
- [x] Clean-up

### The vision

The general idea is to decouple all message UI components from the Message HOC, replacing the logic that now is handled by the later in the form of custom hooks. When migrating the other message UI components, we'll be able to abstract more of the logic that now lives inside the component itself, and that's shared between all of them.

### DIFF
--- a/src/components/Message/MessageDeleted.js
+++ b/src/components/Message/MessageDeleted.js
@@ -37,6 +37,8 @@ const MessageDeleted = (props) => {
 
 MessageDeleted.propTypes = {
   /** The [message object](https://getstream.io/chat/docs/#message_format) */
+  // @ts-ignore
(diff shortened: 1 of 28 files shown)


vicnicius:
> LGTM
> 
> I am sure you have done it already, but just putting it here. Make sure to test all the following cases:
> 
> * message with text
> * message with reactions
> * message with attachment/s
> * message with thread
> * some markdown testing
> * deleted messages
> * read indicators
> * mobile view
> * edit message functionality

Yeah, I did a lot of testing while building it but I'll do one more thorough QA test before merging.

PR #372, opened 2020-06-24
CRS-50 ChatAutoComplete

ath92:
# Submit a pull request

## CLA

Refactor for ChatAutocomplete.
Touched some other things too:
 - Fixed types for commands
 - Also removed `char` prop from EmoticonItem because it was never used, and properly setting PropTypes resulted in errors being thrown. (this component is only used as part of ChatAutoComplete)

Note: relies on type fix in stream-chat-js (https://github.com/GetStream/stream-chat-js/pull/358), so TSC will complain until that's released


src/components/ChatAutoComplete/__tests__/ChatAutocomplete.test.js
```diff
@@ -0,0 +1,180 @@
...
+        id: user.id,
+      }),
+    );
+  });
+
+  it('should use the queryMembers API for mentions if a channel has many members', async () => {
+    const users = Array(100).fill().map(generateUser);
+    const members = users.map((u) => generateMember({ user: u }));
+    const messages = [generateMessage({ user: users[0] })];
+    const queryMembers = jest.fn(() => Promise.resolve({ members }));
```
vishalnarkhede, line 125: I will suggest, lets stick to `useMockedApis` instead of mocking functions manually on js client classes.

src/components/ChatAutoComplete/ChatAutoComplete.js
```diff
@@ -1,134 +1,76 @@
...
-  });
+        const users = response.members.map((m) => m.user);
+        if (onReady) onReady(users);
+      },
+      200,
+    ),
+    [queryMembers],
+  );
 
-  // dataProvider accepts `onReady` function, which will executed once the data is ready.
```
vishalnarkhede, line 122: Why are we removing this comment?

jaapbakker88:
Some conflicts here @ath92

src/components/ChatAutoComplete/ChatAutoComplete.js
ath92: Not intentional, will add it back :)

src/components/ChatAutoComplete/__tests__/ChatAutocomplete.test.js
ath92: I changed the tests to do this. It did mean I had to use the `Chat` and `Channel` component to provide the right context, and I had to write a little helper to set the active channel. It seems that this changes some state in the stream-chat client/channel that is needed for everything to work properly.

Can you please check if what I did makes sense to you? I feel like there might be an easier way to do this, but I'm not totally sure if that's actually the case.

src/components/ChatAutoComplete/__tests__/ChatAutocomplete.test.js
```diff
@@ -1,36 +1,49 @@
...
-const renderComponent = async (
-  props = {},
-  channelContext = channelContextMock,
-) => {
+const renderComponent = async (props = {}, activeChannel = channel) => {
   const placeholderText = props.placeholder || 'placeholder';
   const renderResult = render(
-    <ChannelContext.Provider value={channelContext}>
-      <ChatAutoComplete {...props} placeholder={placeholderText} />
-    </ChannelContext.Provider>,
```
jaapbakker88, line 31: What's the reason for removing this?

src/context/ChatContext.js
```diff
@@ -6,7 +6,7 @@ import { StreamChat } from 'stream-chat';
  * @typedef {import('types').ChatContextValue} ChatContext
  */
 export const ChatContext = React.createContext(
-  /** @type {ChatContext} */ { client: new StreamChat('') },
```
jaapbakker88, line 9: this one's wrong but we'll fix it once we merge all the type fixes in

src/components/MessageInput/MessageInputSmall.js
```diff
@@ -42,7 +42,6 @@ const MessageInputSmall = (props) => {
           <EmojiPicker {...messageInput} small />
           <div className="str-chat__small-message-input--textarea-wrapper">
             <ChatAutoComplete
-              users={messageInput.getUsers()}
```
jaapbakker88, line 45: Can we get rid of this function altogether?

PR #390, opened 2020-06-30
clean up the message list a bit

mahboubii:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/MessageList/MessageList.js
+++ b/src/components/MessageList/MessageList.js
@@ -184,7 +184,7 @@ class MessageList extends PureComponent {
     else if (target === 'bottom')
       scrollTop = containerEl.scrollHeight - containerEl.offsetHeight;
 
-    if (scrollTop) containerEl.scrollTop = scrollTop; // eslint-disable-line no-param-reassign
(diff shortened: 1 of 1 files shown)


src/components/MessageList/MessageList.js
```diff
@@ -306,34 +173,28 @@ class MessageList extends PureComponent {
...
+      Object.prototype.toString.call(target) === '[object Number]';
+
+    let scrollTop;
+    if (isElement) scrollTop = target.offsetTop;
+    else if (isNumber) scrollTop = target;
+    else if (target === 'top') scrollTop = 0;
+    else if (target === 'bottom')
+      scrollTop = containerEl.scrollHeight - containerEl.offsetHeight;
+
+    if (scrollTop) containerEl.scrollTop = scrollTop; // eslint-disable-line no-param-reassign
```
ath92, line 187: `scrollTop` will be implicitly cast to `false` if it equals 0 (i.e. what you set it to if `target === 'top'`, so this will behave differently from the previous code.

You can resolve this by checking `if (scrollTop !== undefined)`.

mahboubii: good catch! pushed a fix

PR #400, opened 2020-07-02
add coverage script

mahboubii:
seems to be working
 [codecov](https://codecov.io/gh/GetStream/stream-chat-react)


package.json
```diff
@@ -140,6 +141,7 @@
   },
   "scripts": {
     "test": "rm -rf built && jest",
+    "coverage": "yarn test --collectCoverage && codecov",
```
vicnicius, line 144: Minor: what do you think of having just the `codecov` call here? Then on CI we can change the tests call to add the `--collectCoverage` flag. Just to keep those scripts atomic.

mahboubii: I actually like that better but it's traditional for the coverage command to actually run the tests along generating the report, @ferhatelmas what do you think?

ferhatelmas: IMHO it seems like a good idea at first but we run coverage only on CI so fine as it is.

PR #420, opened 2020-07-07
refactor edit message logic out of MessageList component

mahboubii:


src/components/Message/Message.js
```diff
@@ -160,13 +155,15 @@ class Message extends Component {
...
   };
 
-  /** @type {(nextProps: Props) => boolean} Typescript syntax */
-  shouldComponentUpdate(nextProps) {
-    return shouldMessageComponentUpdate(nextProps, this.props);
+  /** @type {(nextProps: Props, nextState: State) => boolean} Typescript syntax */
+  shouldComponentUpdate(nextProps, nextState) {
+    return (
+      shouldMessageComponentUpdate(nextProps, this.props) ||
+      !deepequal(nextState, this.state)
```
vicnicius, line 165: What you think of just having the explicit comparison here: `nextState.editing !== this.state.editing`?

mahboubii: yea that makes a lot more sense

mahboubii: actually let's keep this as it is since it might cause confusion in future

PR #422, opened 2020-07-08
CRS-105 Channel

ath92:
# Submit a pull request

## CLA

Refactor Channel component to use hooks


src/components/AutoCompleteTextarea/types.ts
```diff
@@ -17,7 +17,7 @@ export interface TriggerMap {
       q: string,
       text: string,
       onReady: (data: any[], token: string) => void,
-    ) => Promise<Array<Object | string>> | Array<Object | string>;
+    ) => Promise<any> | Array<Object | string>;
```
vicnicius, line 20:
```suggestion
    ) => Promise<void> | Array<Object | string>;
```

src/components/Channel/Channel.js
```diff
@@ -1,266 +1,457 @@
...
 } from '../Loading';
+import useMentionsHandlers from './hooks/useMentionsHandlers';
+import useEditMessageHandler from './hooks/useEditMessageHandler';
+import { channelReducer, initialState } from './channelState';
+
+/** @type {React.FC<import('types').ChannelProps>}>} */
+const Channel = ({ EmptyPlaceholder = null, ...props }) => {
+  const { channel: contextChannel } = useContext(ChatContext);
+  const channel = props.channel || contextChannel;
+  // const { channel } = useContext(ChatContext);
```
vicnicius, line 31: 🔪

src/components/Channel/Channel.js
```diff
@@ -1,266 +1,457 @@
...
+        } catch (e) {
+          dispatch({ type: 'setError', error: e });
+          errored = true;
+        }
+      }
+      done = true;
+      originalTitle.current = document.title;
+      if (!errored) {
+        dispatch({ type: 'initStateFromChannel', channel });
+        if (channel.countUnread() > 0) channel.markRead();
```
vicnicius, line 220: Why do we call `markRead()` above and `channel.markRead()` directly here?

ath92: Good question, I'm not quite sure either. This was how the old component handled things, so I just kept it for the time being.

But since the `markRead` callback does almost the same if called from here with some additional safeguards, it makes sense to just call that instead. I changed that in the latest commit.

PR #423, opened 2020-07-08
Replace lodash.truncate with simple utility

ath92:
# Submit a pull request

## CLA

Replace lodash.truncate module with a simple utility function.
Should save some bytes (see https://github.com/lodash/lodash/blob/master/truncate.js).
Also removes the last dependency on the lodash package (i.e. not `lodash.<something>`, but `lodash/<something>`)


mahboubii:
@ath92 could you add some simple tests for it too?

mahboubii:
@ath92 @jaapbakker88 it appears that `lodash.uniqby` also has limited usage and can be removed, this way we can remove lodash completely

ath92:
@mahboubii this PR combined with https://github.com/GetStream/stream-chat-react/pull/422 should get rid of all uses of the full lodash package (I think). I think the reduced bundle size here might be just from getting rid of lodash.truncate

PR #499, opened 2020-08-31
Ts chat fixes for upgrade to stream-chat@2.x.x

nhannah:
# First level integration of stream-chat-js with typescript

## CLA

Correct current type issues with stream js client, prior to full integration with generics

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ The [styleguidist docs for stream-chat-react](https://getstream.github.io/stream
 
 The React components are created using the [stream-chat-js](https://github.com/getstream/stream-chat-js) library. If you're customizing the components, it's important to learn how the Chat Server API works. You'll want to review our [JS chat API docs](https://getstream.io/chat/docs/js/).
 
+**Note:** The [stream-chat-js](https://github.com/getstream/stream-chat-js) library allows for fully typed responses using generics, currently the React SDK does not allow for user defined types via generics, so custom fields will be returned with type `unknown` and need to be ignored using `@ts-ignore` when using custom components in typescript.
+
 ## Commands
 
 - yarn docs-server
@@ -62,9 +64,9 @@ The React components are created using the [stream-chat-js](https://github.com/g
 
 ## Component Reusability
 
-1.  If a component implements a ton of logic, it's helpful if you split it out into two parts: The top-level component, which handles all the logic, and a lower level component, which handles rendering. That makes it easy to change the rendering without having to touch the other stuff. Have a look at Message and MessageTeam to see how this approach works.
+1. If a component implements a ton of logic, it's helpful if you split it out into two parts: The top-level component, which handles all the logic, and a lower level component, which handles rendering. That makes it easy to change the rendering without having to touch the other stuff. Have a look at Message and MessageTeam to see how this approach works.
 
-2.  Make things configurable via the props where possible. Sometimes an even better approach is to use the props.children. Have a look at how flexible the channel layout is due to this approach:
+2. Make things configurable via the props where possible. Sometimes an even better approach is to use the props.children. Have a look at how flexible the channel layout is due to this approach:
 
 ```jsx
 <Channel>
@@ -131,7 +133,7 @@ Note that the PureComponent uses a shallow diff to determine if a component shou
 The regular component simply always rerenders when there is a state change.
 
 You can read more about PureComponents and common gotchas here:
-https://codeburst.io/when-to-use-component-or-purecomponent-a60cfad01a81
+<https://codeburst.io/when-to-use-component-or-purecomponent-a60cfad01a81>
 
 You want the shallow diff only to be true if something changed.
 Common mistakes that hurt performance are:
@@ -146,13 +148,13 @@ Common mistakes that hurt performance are:
 Instance of class `Streami18n` should be provided to the Chat component to handle translations.
 Stream provides the following list of in-built translations for components:
 
-1.  English (en)
-2.  Dutch (nl)
-3.  Russian (ru)
-4.  Turkish (tr)
-5.  French (fr)
-6.  Italian (it)
-7.  Hindi (hi)
+1. English (en)
+2. Dutch (nl)
+3. Russian (ru)
+4. Turkish (tr)
+5. French (fr)
+6. Italian (it)
+7. Hindi (hi)
 
 The default language is English. The simplest way to start using chat components in one of the in-built languages would be the following:
 
@@ -178,7 +180,7 @@ const i18n = new Streami18n({
 });
 ```
 
-You can find all the available keys here: https://github.com/GetStream/stream-chat-react/tree/master/src/i18n
+You can find all the available keys here: <https://github.com/GetStream/stream-chat-react/tree/master/src/i18n>
 
 They are also exported as a JSON object from the library.
 
@@ -195,7 +197,7 @@ import {
 } from 'stream-chat-react';
 ```
 
-Please read this docs on i18n for more details and further customizations - https://getstream.github.io/stream-chat-react/#section-streami18n
+Please read this docs on i18n for more details and further customizations - <https://getstream.github.io/stream-chat-react/#section-streami18n>
 
 ## Contributing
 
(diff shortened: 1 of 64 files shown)


vishalnarkhede (changes requested):
Have some small comments :) Other than that LGTM

README.md
```diff
@@ -54,6 +54,8 @@ The [styleguidist docs for stream-chat-react](https://getstream.github.io/stream
 
 The React components are created using the [stream-chat-js](https://github.com/getstream/stream-chat-js) library. If you're customizing the components, it's important to learn how the Chat Server API works. You'll want to review our [JS chat API docs](https://getstream.io/chat/docs/js/).
 
+**Note:** The [stream-chat-js](https://github.com/getstream/stream-chat-js) library allows for fully typed responses using generics, currently the React SDK does not allow for user defined types via generics, so custom fields will be returned with type `unknown` and need to be ignored using `@ts-ignore` when using custom components in typescript.
+
```
vishalnarkhede, line 58:
```suggestion
## Typescript

**Note:** The [stream-chat-js](https://github.com/getstream/stream-chat-js) library allows for fully typed responses using generics, currently the React SDK does not allow for user defined types via generics, so custom fields will be returned with type `unknown` and need to be ignored using `@ts-ignore` when using custom components in typescript.

```

vishalnarkhede:
@nhannah tests are failing for some reason in ci

src/components/ChatAutoComplete/ChatAutoComplete.js
```diff
@@ -147,24 +147,26 @@ const ChatAutoComplete = (props) => {
               return [];
             }
             const selectedCommands = commands.filter(
-              (c) => c.name.indexOf(q) !== -1,
+              (c) => c.name?.indexOf(q) !== -1,
```
mahboubii, line 150: commands should always have a name right?

src/components/EventComponent/EventComponent.js
```diff
@@ -30,9 +30,9 @@ const EventComponent = ({ message }) => {
 
   if (
     type === 'channel.event' &&
-    (event.type === 'member.removed' || event.type === 'member.added')
+    (event?.type === 'member.removed' || event?.type === 'member.added')
```
mahboubii, line 33: not sure if we should use all these optional chainings just to silence the ts compiler

src/components/Message/hooks/useActionHandler.js
```diff
@@ -19,13 +19,14 @@ export const useActionHandler = (message) => {
...
+    if (messageID) {
+      const data = await channel.sendAction(messageID, formData);
 
-    const data = await channel.sendAction(messageID, formData);
-
-    if (data && data.message) {
-      updateMessage(data.message);
-    } else {
-      removeMessage(message);
+      if (data && data.message) {
```
mahboubii, line 25:
```suggestion
      if (data?.message) {
```

src/components/Message/hooks/useReactionHandler.js
```diff
@@ -48,28 +48,29 @@ export const useReactionHandler = (message) => {
...
-    if (userExistingReaction) {
-      reactionChangePromise = channel.deleteReaction(
-        message.id,
-        // @ts-ignore Typescript doesn't understand that the userExistingReaction variable might have been mutated inside the foreach loop
-        userExistingReaction.type,
-      );
-    } else {
-      // add the reaction
-      const messageID = message.id;
+    if (message.id) {
```
mahboubii, line 51: same goes for this extra check

types/index.d.ts
```diff
@@ -706,12 +780,7 @@ export interface MessageDeletedProps extends TranslationContextValue {
 }
 
 export interface ThreadProps {
-  Message?: React.ElementType<MessageUIComponentProps>;
```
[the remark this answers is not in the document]
nhannah: Let me look, I took it off cause it's in `@typedef { import('types').ThreadProps & import('types').ChannelContextValue & import('types').TranslationContextValue} Props` under `ChannelContextValue`.

src/components/EventComponent/EventComponent.js
nhannah: Moved to just a single event check, but I think this is needed until we are in real TS, then we could cast or type guard based on the `type` but in this JSDocs format it get's messy quick.

src/components/Message/hooks/useReactionHandler.js
nhannah: Yeah what is optional and not could use another sweep through in the stream chat js, this one is cause when a message is not a response id is optional and they get intermingled in use. Dan and I chose to be conservative where we weren't 100% and list things as optional to reduce the risk of exceptions.

src/components/ChatAutoComplete/ChatAutoComplete.js
nhannah: This is the question I was posing in Slack, we need to confirm with the backend, will stay conservative for now.

src/components/ChannelList/ChannelListTeam.js
```diff
@@ -45,19 +45,19 @@ const ChannelListTeam = ({
         <div className="str-chat__channel-list-team__header">
           <div className="str-chat__channel-list-team__header--left">
             <Avatar
-              image={client.user.image}
-              name={client.user.name || client.user.id}
+              image={client.user?.image}
```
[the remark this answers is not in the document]
nhannah: I think the overall problem with this is it breaks typescript safety if we start ignoring the types via ts-ignore based on assumptions of their current state. e.g. At some point in the future some new bit is added to the js client that would make the user undefined at a point they were on this screen. No type error would be caused as we ignored it under the assumption it would never be undefined at this point, but it would then cause a crash as we try to access the property image of undefined. I think the cleanest solution to the addition of multiple opt chains is something like `const { image, name, id, status } = client.user || {};`

src/components/ChannelHeader/ChannelHeader.js
```diff
@@ -32,26 +32,26 @@ const ChannelHeader = ({ title, live }) => {
...
             <span className="str-chat__header-livestream-left--livelabel">
               {t('live')}
             </span>
           )}
         </p>
-        {channel && channel.data.subtitle && (
+        {channel && channel.data?.subtitle && (
           <p className="str-chat__header-livestream-left--subtitle">
-            {channel.data.subtitle}
+            {channel.data?.subtitle}
```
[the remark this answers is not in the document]
nhannah: It was inside of a check for channel already, changed some of this up though.

src/components/Channel/channelState.js
```diff
@@ -59,7 +59,9 @@ export const channelReducer = (state, action) => {
       if (!state.thread) return state;
       return {
         ...state,
-        threadMessages: channel.state.threads[state.thread.id],
+        threadMessages: state.thread?.id
+          ? channel.state.threads[state.thread.id]
+          : Immutable([]),
```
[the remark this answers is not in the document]
nhannah: I'll look at this, @vishalnarkhede made the change I think so need to dig a little.

src/components/Message/hooks/useReactionHandler.js
nhannah: Staying conservative it seems...

src/components/Channel/channelState.js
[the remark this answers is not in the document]
nhannah: OK looks like we should always have a fallback for threadMessages to empty Immutable Array, it starts as that and is set to that on close so I have changed it so it will gall back to that.

types/index.d.ts
```diff
@@ -208,37 +267,37 @@ export interface ChannelListProps {
   lockChannelOrder?: boolean;
   onMessageNew?(
     thisArg: React.Dispatch<React.SetStateAction<Client.Channel[]>>,
-    e: Client.Event<string>,
+    e: Client.Event,
```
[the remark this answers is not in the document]
nhannah: Updated the change log https://github.com/GetStream/stream-chat-js/pull/419

types/index.d.ts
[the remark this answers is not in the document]
nhannah: Ok just matched the hand written types to the component types

vishalnarkhede (approved):

PR #501, opened 2020-09-04
[CRS-281] MML integration

gumuz:
Integrating the mml lib into, will not work correctly without these: https://github.com/GetStream/mml-react/pull/1


src/components/Message/hooks/useActionHandlerMML.js
```diff
@@ -0,0 +1,36 @@
+// @ts-check
```
mahboubii, line 1: this is basically the same as `useActionHandler`
https://github.com/GetStream/stream-chat-react/pull/562/files#diff-aa50df8b4b857ab266247a60ceb2e55f

src/components/Message/MessageText.js
```diff
@@ -58,6 +61,12 @@ const MessageTextComponent = (props) => {
     return null;
   }
 
+  const renderedText = message?.mml ? (
+    <MML source={message.mml} onSubmit={handleActionMML} />
+  ) : (
+    messageText
+  );
```
mahboubii, line 68: IMO, we should stick to the attachment type, mixing up the text render with the mml could make things unnecessary complicated

mahboubii (changes requested):
@gumuz are we going to support mml in `message.mml` and `message.attachments[{type: mml}]` ?

gumuz:
> @gumuz are we going to support mml in `message.mml` and `message.attachments[{type: mml}]` ?

Honestly, I'm not sure. I guess you can already use attachment types in any way you want at the moment, does not need any backend support afaik. Thoughts @tbarbugli ?

ps. this PR was a proof of concept, feel free to use/discard it however you want.

PR #514, opened 2020-09-10
fix loadMore

jaapbakker88:
# Submit a pull request

## CLA


src/components/Channel/Channel.js
```diff
@@ -136,6 +138,7 @@ const ChannelInner = ({
   const originalTitle = useRef('');
   const lastRead = useRef(new Date());
   const chatContext = useContext(ChatContext);
+  const [online, setOnline] = useState(true);
```
mahboubii, line 141: we can use ref instead of state here to avoid causing re-render since it's not necessary

mahboubii (approved):
🌮

mahboubii:
@jaapbakker88 if you think it's necessary you can also add a simple test for this change

PR #518, opened 2020-09-15
Fix/fix prop types

vicnicius:
This avoids `channel` and `client` PropType warnings in development mode. The custom PropType validator relies on a simple comparison of constructor names, but this should be good enough. When it fails the custom validator will do so with the following message: 
image


mahboubii (approved):
🌮

PR #519, opened 2020-09-15
Fix dependencies and improve bundle configuration

vicnicius:

Currently, the way we setup the SDK, we expect consumers of it to have stream-chat as a dependency already. This was not explicit in our package.json, so this PR fixes it. While at it, I also took the opportunity to clean up the rollup configurations a bit. This should make the work of reducing the bundle size slightly easier.

To-do:
- [x] - Test compressed bundle in different browsers.


package.json
```diff
@@ -41,13 +41,13 @@
...
     "seamless-immutable": "^7.1.4",
-    "stream-chat": "^2.2.2",
     "textarea-caret": "^3.1.0",
     "uuid": "^7.0.3"
   },
   "peerDependencies": {
     "react": "^16.8.x",
-    "react-dom": "^16.8.x"
+    "react-dom": "^16.8.x",
+    "stream-chat": "^2.2.x"
```
mahboubii, line 50:
```suggestion
    "stream-chat": "2.2.x"
```
I'd suggest we lock in the version to only patches

package.json
```diff
@@ -145,6 +146,7 @@
   "scripts": {
     "test": "rm -rf built && jest",
     "coverage": "yarn test --collectCoverage && codecov",
+    "bundle": "rollup -c",
     "start": "mkdir -p dist/assets assets && yarn build-translations && rollup -c -w",
     "build": "mkdir -p dist/assets assets && yarn build-translations && rollup -c",
```
mahboubii, line 151:
```suggestion
    "start": "mkdir -p dist/assets assets && yarn build-translations && yarn bundle -w",
    "build": "mkdir -p dist/assets assets && yarn build-translations && yarn bundle",
```

mahboubii:
@vicnicius did you try using the bundled output in an example app in the browser, the size reduction is huge and I have a feeling it's broken

vicnicius:
> @vicnicius did you try using the bundled output in an example app in the browser, the size reduction is huge and I have a feeling it's broken

I have the same feeling. I'll test with different setups to make sure everything is good 👍

mahboubii (approved):

jaapbakker88:
@vicnicius we'll need to fix this tutorial after we publish this https://getstream.io/chat/react-chat/tutorial/

PR #530, opened 2020-09-21
Optimize emoji-mart

vicnicius:

This PR introduces some improvements to how we handle `emoji-mart` that should lead to more flexibility for clients and a smaller default bundle for us. It achieves this by doing the following:

**1. Create a smaller default set of emojis for us to use:**

A big chunk of the emoji-mart library comes from the JSON file that contains the emoji data. While analyzing this data, I noticed that the data that is provided for handling the skin color feature takes a large amount of space. So with the help of a tiny script, I removed all the data related to skin colors in the emojis, based on the Facebook set provided by emoji-mart. Because the feature wouldn't work anymore without this data, it was disabled. With this change, we reduced the size of the compressed emoji data JSON from ~597KB to ~183KB.

Screenshot 2020-09-23 at 14 39 48

**2. Making sure we only use the "Nimble" version of the emoji-mart components, and always from their `dist-modern` bundle:**

Because we now use our own emoji data source, it is important to make sure we only use the components EmojiMart calls "Nimble," else we would end up importing the default dataset anyways. Together with this change, we can work with the not-transpiled version of their library because we transpile the javascript ourselves, so all imports are now pointing to their dist-modern bundle.

**3. Allowing for an easy way to overwrite both the emoji dataset and the underlying components:**

By providing prop overrides and injecting the emojiConfig object into the ChannelContext, we allow clients to have more control over how they want to use emojis with the SDK. Whether they want to keep the previous behavior or customize their own EmojiPicker, this should now be easily achievable. Internally, we should always rely on the ChannelContext emojiConfig to do anything emoji related.

Default Picker before and after:


vishalnarkhede:
@vicnicius I can see `dist` folder being committed. Is that intentional?

vicnicius:
> @vicnicius I can see `dist` folder being committed. Is that intentional?

Yes, I was doing some tests. I'll rebase and remove the temporary commit. 👍

vicnicius (commented):
LGTM. I can't approve it since I opened the PR in the first place.

DanC5 (approved):

PR #548, opened 2020-09-28
Feature/new messages indicator

### LINKED ISSUE #154: New messages
Is it possible to show new messages DateSeparator and have a flag on each new message that is new? 

Example: I have a new message on a channel which i didn't open. When I open it I want to see the 'New messages' DateSeparator and those new messages to have a red background. (Similar to Slack's new messages separator + a flag for each message that is considered new)

mahboubii:
fixes #154

### DIFF
--- a/src/components/MessageList/MessageListInner.js
+++ b/src/components/MessageList/MessageListInner.js
@@ -35,7 +35,8 @@ const getReadStates = (messages, read) => {
   return readData;
 };
 
-const insertDates = (messages) => {
+const insertDates = (messages, lastRead, userID) => {
+  let unread = false;
   const newMessages = [];
   for (let i = 0, l = messages.length; i < l; i += 1) {
     const message = messages[i];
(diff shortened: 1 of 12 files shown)


src/components/MessageList/MessageListInner.js
```diff
@@ -49,7 +50,21 @@ const insertDates = (messages) => {
       prevMessageDate = messages[i - 1].created_at.toDateString();
     }
 
-    if (i === 0 || messageDate !== prevMessageDate) {
+    if (!unread) {
+      unread = lastRead.getTime() < message.created_at.getTime();
```
vicnicius, line 54: is `lastRead` always defined?

vicnicius: It looks like it isn't: https://github.com/GetStream/stream-chat-js/blob/ca4f5a587d7d10c6a424b82eba0a359605f0023a/src/channel.ts#L1110

```suggestion
      unread = lastRead && lastRead.getTime() < message.created_at.getTime();
```

mahboubii: logically it should not be undefined at this stage, but won't hurt to have an extra check 👍

jaapbakker88 (approved):
LGTM

vicnicius (approved):

src/components/MessageList/MessageListInner.js
vicnicius: And will make TS less angry in the future x)

PR #562, opened 2020-10-06
Feature/mml

mahboubii:
based on #104

TODO: 
- [x] upgrade mml-react to fix missing dependencies, right now you have to link local version 
- [x] tests
- [x] fix types


src/styles/Message.scss
```diff
@@ -986,3 +986,61 @@
...
+    }
+  }
+
+  // expand carousel to the edge of the screen, the horizontal padding is theme
+  // specific in stream-chat-react
+  &.messaging {
+    // as .messaging.str-chat .str-chat__list:
+    @media screen and (max-width: 960px) {
+      .mml-align-left .mml-carousel {
+        margin-right: -10px;
```
kuus, line 1039: sorry this was a mistake of mine I fixed in a later commit, should be `margin-left` here @mahboubii

kuus (changes requested):

src/styles/Message.scss
mahboubii: thanks for the heads up

src/components/Message/FixedHeightMessage.js
```diff
@@ -91,6 +93,15 @@ const FixedHeightMessage = ({ message, groupedByUser }) => {
 
         <div className="str-chat__virtual-message__text" data-testid="msg-text">
           {renderedText}
+
+          {message.mml && (
```
DanC5, line 97: what type is `message.mml`?

src/components/Message/MessageSimple.js
```diff
@@ -206,6 +207,15 @@ const MessageSimple = (props) => {
                 reactionSelectorRef={reactionSelectorRef}
               />
             )}
+
+            {message.mml && (
```
DanC5, line 211: not sure it matters but some optional chained and some not

types/index.d.ts
```diff
@@ -765,9 +777,9 @@ export interface BaseAttachmentUIComponentProps {
 		Examples include canceling a \/giphy command or shuffling the results.
 		*/
   actionHandler?(
-    name: string,
-    value: string,
-    event: React.BaseSyntheticEvent,
+    name: string | Record<string, any>,
```
DanC5, line 780: not a huge deal but these will prob need to be unknown rather than any when we go to TS

src/components/Message/FixedHeightMessage.js
mahboubii: string|undef

src/components/Message/MessageSimple.js
mahboubii: yea just to suppress TS errors, types are not accurate

DanC5 (approved):

PR #566, opened 2020-10-09
CRS-292 Prevent state updates on unmounted Channel component

ath92:
# Description of the pull request
This prevents React from logging a warning to the console if the `Channel` component finishes its `loadMore` callback after it has already been unmounted.

Steps to reproduce:
1. Trigger Channel's `loadMore` function
2. Unmount Channel component (e.g. by switching to a different channel) while `loadMore` is still fetching
3. When the request resolves, Channel will try to call dispatch on its state reducer, even though that's a noop. React complains with a warning:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
in ChannelInner (created by Channel)
```

Note: I spent quite some time trying to write a test case for this, but it seems that it's very hard to figure out if a component tries to update its state after it has been unmounted. For example, React prevents the actual reducer from being called, so you cannot just spy on that to know if there was an attempted state update.


jaapbakker88 (approved):
LGTM

src/components/Channel/Channel.js
```diff
@@ -138,6 +138,7 @@ const ChannelInner = ({
   const lastRead = useRef(new Date());
   const chatContext = useContext(ChatContext);
   const online = useRef(true);
+  const isMounted = useRef(true);
```
vishalnarkhede, line 141: I have seen this weird issue with our previous components as well. LGTM

@ath92 Maybe can we extract it to separate hook? 

```js
const isMounted = useIsMounted();
```

ath92: I suppose this hook would basically be a combination of `useRef` for storing the flag + `useEffect` for setting it to true/false?

vishalnarkhede: yeah

ath92: Just did this, also allowed me to write a test quite easily :)

vicnicius:
Would be nice if we could also get rid of the root cause of the warning. In this case, cancel the promise or abort the request.

ath92:
> Would be nice if we could also get rid of the root cause of the warning. In this case, cancel the promise or abort the request.

It seems `stream-chat-js` gives access to the axios instance, and lets me pass options into the actual axios call used by `loadMore`, so I think it should be possible to cancel the request that way. 

But I don't think that would actually get rid of state updates; `axios` would reject its Promise, and we'd end up in the `catch` block where there's another state update. 

I think in this case cancelling the request will add little benefit: it would require additional stateful logic (i.e. we'd have to keep the axios cancelToken somewhere, and make sure the request actually gets cancelled on unmount). If the server handles cancellation requests, I think in most cases this request would arrive at the server too late to make a difference.

jaapbakker88:
@ath92 think this will do for now, let's get this merged and released.

PR #567, opened 2020-10-09
Adding livestream messagelist

vishalnarkhede:
# Submit a pull request

## CLA


src/components/MessageList/MessageList.js
```diff
@@ -32,6 +32,7 @@ class MessageList extends PureComponent {
     this.state = {
```
vishalnarkhede, line 32: Adding this file only to make it easier to see diff. This file will be reverted after the review

src/components/MessageList/LivestreamMessageList.js
```diff
@@ -0,0 +1,502 @@
...
+      this.setState({
+        slicedMessages: this.props.messages.slice(
+          -(this.state.slicedMessages.length + this.props.messageLimit),
+        ),
+      });
+      return;
+    }
+
+    // Otherwise fetch from server.
+    if (this.props.messageLimit) {
```
tatianainama, line 260: Not really familiar with the code, but is this condition necessary? You end up calling `props.loadMore` with an `undefined` value for `messageLimit` anyways

mahboubii: passing `undefined` down will replace the `loadMore(limit = 100)` default value, good call though, we can simplify this a bit

vishalnarkhede: Makes sense ... have fixed it by adding default value to messageLimit prop and removing this if condition

src/components/MessageList/LivestreamMessageList.js
```diff
@@ -0,0 +1,502 @@
...
+
+    if (scrollTop !== undefined) containerEl.scrollTop = scrollTop; // eslint-disable-line no-param-reassign
+  };
+
+  goToNewMessages = () => {
+    this.scrollToBottom();
+    this.setState({ newMessagesNotification: false });
+  };
+
+  userScrolledUp = () => this.scrollOffset > 310;
```
tatianainama: Maybe a nitpick, but what does this `310` means?

src/components/MessageList/LivestreamMessageList.js
```diff
@@ -0,0 +1,502 @@
...
+  /**
+   * target - target to scroll to (DOM element, scrollTop Number, 'top', or 'bottom'
+   * containerEl - DOM element for the container with scrollbars
+   * source: https://stackoverflow.com/a/48429314
+   */
+  scrollToTarget = (target, containerEl) => {
+    // Moved up here for readability:
+    const isElement = target && target.nodeType === 1;
+    const isNumber =
+      Object.prototype.toString.call(target) === '[object Number]';
```
tatianainama: Using `toString()` to check types is unreliable (the `toStringTag` can be changed to anything). How about using `typeof` instead?

jaapbakker88 (approved):

PR #571, opened 2020-10-13
feat: VirtualList support message grouping

mahboubii:


src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -34,6 +34,8 @@ const VirtualizedMessageList = ({
   loadingMore,
   messageLimit = 100,
   overscan = 200,
+  noGroupByUser = false,
```
vicnicius, line 37: What you think of:
```suggestion
  shouldGroupByUser = true,
```
Just to avoid the double negation `!noGroupByUser` later

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -101,9 +107,18 @@ const VirtualizedMessageList = ({
...
         return smartRender(MessageDeleted, { message }, null);
 
-      return <Message message={message} />;
+      return (
+        <Message
+          message={message}
+          groupedByUser={
+            !noGroupByUser &&
+            i > 0 &&
+            message.user.id === messageList[i - 1].user.id
```
vicnicius, line 116: Can we always assume messages will have users here? I was looking into https://github.com/GetStream/stream-chat-js/blob/a68c23ed1bdbd8dbc71241ec6d7f66b086836a47/src/types.ts#L1366 and I'm not sure.

src/components/MessageList/VirtualizedMessageList.js
mahboubii: I was going the same direction as MessageList component but this makes more sense, will update

src/components/MessageList/VirtualizedMessageList.js
mahboubii: yea the returned messages from API will always have a user AFAIK, the null/undef type in there is not accurate since it's a shared type I guess

vicnicius (approved):

Issue #565, opened 2020-10-09
Limit the number of attachments selected


alexlopezit:
Is there is any way we can prevent the user to select more than 10 attachments as it seems that is the limit when sending a message to the server, at the moment you can select more than that and then send the message which will fail, then clicking the link `Message failed. Click to try again.` will simply keep failing in the background and the users does not know why, also there is no way to remove this failed message from the UI witch is a bit confusing, unless you refresh the page.

![image](https://user-images.githubusercontent.com/1654135/95539858-d1d56d80-0a32-11eb-9073-5cdf6295ec07.png)

gz#6534

mahboubii:
Thanks for reporting this, currently API limits the number of attachments to a maximum of 10 for each message, we will put a reflect this limit in our components soon.

alexlopezit:
Thanks @mahboubii 

I just found this prop we can use on the Channel component `maxNumberOfFiles`, which basically disables the the `ImageDropzone` and the `FileUploadButton`, which is what we are looking for, however would be good to get a visual indication or display a message somewhere so the user knows why it's been disabled.

I would suggest to perhaps changing this text https://github.com/GetStream/stream-chat-react/blob/30771692fcec5057eb4d76f17208e714e5a38cbf/src/components/MessageInput/MessageInputLarge.js#L122 to something along the lines `You have reached the max number of attachments per message` or something like that.

Also if possible could the FileUploadButton https://github.com/GetStream/react-file-utils/blob/master/src/components/FileUploadButton.js have a class when disabled, so we can perhaps change the color of the icon to make it look disabled too.

And looks like there is a bug with `ImageDropzone` as if I set the `maxNumberOfFiles={4}` but the I drag 5 or more files, they all get accepted.

jaapbakker88:
@alexlopezit problem is that it disables when you reach the limit, but as long as you haven't reached the limit it allows you to add multiple files that will get you across that limit. We're looking into a solution for this. 

Visual indication is something we will definitely add as well.

mahboubii:
@alexlopezit fixed in [v3.1.5](https://github.com/GetStream/stream-chat-react/releases/tag/v3.1.5)

alexlopezit:
Hi @jaapbakker88 and @mahboubii , I just checked this on v3.1.5 and though it was not working until I manually added the prop `maxNumberOfFiles` to the channel, perhaps you want to make set the default value if not present?

mahboubii:
[...]

PR #577, opened 2020-10-14
CRS-293 Max attachments

ath92:
# Submit a pull request

Draft: relies on https://github.com/GetStream/react-file-utils/pull/22. Relates to https://github.com/GetStream/stream-chat-react/issues/565

This PR uses `channelContext.maxNumberOfFiles` to prevent the user from uploading too many files. It does so by doing two things:
 - setting the `maxNumberOfFiles` prop in `ImageDropzone` (part of `react-file-utils`)
 - disabling the file upload button if needed

Other fixes:
 - Removing while uploading no longer throws
 - if `channelContext.multipleUploads` is set to false, you can now only upload a single file (and its default value is now `true`)

To do:
 - Get translations for tooltip ("You've reached the maximum number of files")
 - UI indication in ImageDropzone, when attempting to drop more than the allowed number of files (in `react-file-utils`)

### DIFF
--- a/src/components/Channel/Channel.js
+++ b/src/components/Channel/Channel.js
@@ -35,6 +35,10 @@ const Channel = ({ EmptyPlaceholder = null, ...props }) => {
   return <ChannelInner {...props} channel={channel} key={channel.cid} />;
 };
 
+Channel.defaultProps = {
+  multipleUploads: true,
+};
+
 Channel.propTypes = {
   /** Which channel to connect to, will initialize the channel if it's not initialized yet */
   channel: PropTypes.instanceOf(StreamChannel),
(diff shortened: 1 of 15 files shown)


src/components/Channel/Channel.js
```diff
@@ -96,7 +100,7 @@ Channel.propTypes = {
    * @param {User} user   Target [user object](https://getstream.io/chat/docs/#chat-doc-set-user) which is hovered
    */
   onMentionsHover: PropTypes.func,
-  /** Weather to allow multiple attachment uploads */
```
jaapbakker88, line 99: 🤣

jaapbakker88 (approved):

PR #579, opened 2020-10-15
feat: TypingIndicator renders null if typing_events is disabled

mahboubii:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/TypingIndicator/TypingIndicator.js
+++ b/src/components/TypingIndicator/TypingIndicator.js
@@ -10,9 +10,10 @@ import { Avatar } from '../Avatar';
  * @type {React.FC<Props>}
  */
 const TypingIndicator = ({ avatarSize = 32 }) => {
-  const { typing, client } = useContext(ChannelContext);
+  const { typing, client, channel } = useContext(ChannelContext);
 
-  if (!typing || !client) return null;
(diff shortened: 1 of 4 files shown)


jaapbakker88 (approved):

vicnicius:
Shouldn't we also (only?) check here: https://github.com/GetStream/stream-chat-react/blob/cde4d4642323a6491b83c27a4febb4560feafcab/src/components/MessageList/VirtualizedMessageList.js#L143 and here https://github.com/GetStream/stream-chat-react/blob/2c2180ed10d0b8afaa90d17779e09e1de5bca010/src/components/MessageList/MessageListInner.js#L309 ?

mahboubii:
> Shouldn't we also (only?) check here:
> 
> https://github.com/GetStream/stream-chat-react/blob/cde4d4642323a6491b83c27a4febb4560feafcab/src/components/MessageList/VirtualizedMessageList.js#L143
> 
> and here
> https://github.com/GetStream/stream-chat-react/blob/2c2180ed10d0b8afaa90d17779e09e1de5bca010/src/components/MessageList/MessageListInner.js#L309
> 
> ?

The issue arises when someone uses a custom TypingIndicator that does not respect `typing_events` (probably not often) but adding these checks to message list components increase their complexity, which is already too much.

vicnicius (approved):

PR #580, opened 2020-10-15
feat: Thread renders null if replies is disabled

mahboubii:
pending on #579 for tests to pass


mahboubii:
> @mahboubii can we also remove the reply button from the UI components?

ofc, already on the todo list.

jaapbakker88 (approved):

PR #584, opened 2020-10-19
feat: disable Input upload if disabled in channel config

mahboubii:


src/components/MessageInput/hooks/messageInput.js
```diff
@@ -665,8 +665,11 @@ export default function useMessageInputState(props) {
     [uploadNewFiles, insertText],
   );
 
+  const isUploadEnabled = channel?.getConfig?.()?.uploads !== false;
```
vicnicius, line 668: So we want it to be considered enabled if for some reason we can't get the config explicitly, is that it?

mahboubii: correct, in case the config is undefined we assume the features are supported

mahboubii:
closing due to merge conflicts

PR #585, opened 2020-10-19
feat: Input upload disabled based on channel config

mahboubii:


src/components/MessageInput/EditMessageForm.js
```diff
@@ -39,13 +38,13 @@ const EditMessageForm = (props) => {
         accept={channelContext.acceptedFiles}
         multiple={channelContext.multipleUploads}
         disabled={
-          channelContext.maxNumberOfFiles !== undefined &&
-          messageInput.numberOfUploads >= channelContext.maxNumberOfFiles
+          !messageInput.isUploadEnabled || messageInput.maxFilesLeft === 0
         }
+        maxNumberOfFiles={messageInput.maxFilesLeft}
```
mahboubii, line 43: EditMessageForm changed regarding max attachments was missing from #577

vicnicius (approved):

src/components/MessageInput/hooks/messageInput.js
```diff
@@ -687,8 +687,11 @@ export default function useMessageInputState(props) {
 
   const maxFilesLeft = maxFilesAllowed - numberOfUploads;
 
+  const isUploadEnabled = channel?.getConfig?.()?.uploads !== false;
```
vicnicius, line 690: A small suggestion, as I think this line is a little bit tricky to read. Feel free to ignore it.

```suggestion
  const uploadConfig = channel?.getConfig?.()?.uploads;
  const isUploadEnabled = typeof uploadConfig === 'undefined' || uploadConfig === true;
```

vicnicius: or it could be null as well 😅  maybe just adding a comment would do, so we're sure the undefined !== false was intended. x)

mahboubii: agree, this is easier to read, but there are few more instances of `channel?.getConfig?.().X !== false` for other properties across the codebase, so let's keep this one similar to them. we can simplify them all together later

PR #591, opened 2020-10-27
Fix: Escape special characters in mentioned users names

jaapbakker88:
# Submit a pull request

## CLA


ath92 (approved):
LGTM. One note though: if a username contains any of the characters we're now escaping, it won't properly highlight the mention, right? Should this be documented somewhere?

mahboubii (approved):

mahboubii:
> LGTM. One note though: if a username contains any of the characters we're now escaping, it won't properly highlight the mention, right? Should this be documented somewhere?

it uses the name and if missing fallback to id, this change actually solves the issue of not highlighting or breaking where a name has one of those special chars.

PR #596, opened 2020-10-29
Fix #595 Handling upload errors

### LINKED ISSUE #595: Preventing file upload for files over 20mb still renders an empty thumbnail preview
We are currently trying to prevent the user from selecting images bigger than 20mb as it looks like is the limit on the API.

I tried to use the `doImageUploadRequest` where I can check the size of the image, the idea is to notify the user when the file is bigger than 20mb and then not upload that image, here is the code I am using.

```
  const doImageUploadRequest = (file: File) => {
    const { size } = file
    const sizeInMB = size / (1024 * 1024)

    if (sizeInMB > 20) {
      // We can show a toast notification here letting the user know
      return channel.sendImage(null)
    }

    return channel.sendImage(file)
  }

...

<MessageInput doImageUploadRequest={doImageUploadRequest} />
```

The problem is that even when I try to not upload that large image, a thumbnail preview is generated which contains a retry svg icon, is there a way to prevent the rendering of this "retry thumbnail"?

cc @tbarbugli 

gz#6954

vishalnarkhede:
# Submit a pull request

## CLA


src/components/MessageInput/hooks/messageInput.js
```diff
@@ -560,6 +573,7 @@ export default function useMessageInputState(props) {
         if (!imageUploads[id]) {
           alreadyRemoved = true;
         } else {
+          console.warn('>>');
```
mahboubii, line 576:
```suggestion
```

mahboubii (approved):

ath92 (approved):

PR #611, opened 2020-11-17
CRS-323 Virtualized dynamic heights

### LINKED ISSUE #603: scroll does not stick to bottom on message height change.
VirtualizedMessageList does not seem to stick to the bottom when the message height is changed. In our situation users can add emoji as reactions to messages. when this is done we simply display the emoji reaction at the bottom of the message. This causes the message height to increase. this causes the react-virtuoso to lose track of isAtBottom. this is because followOutput breaks when a item is resized. 

```
followOutput?: boolean

If set to true, the list will automatically scroll to bottom if the total count is changed. See the Stick to Bottom scenario for a working example.
```

gz#7172

ath92:
# Submit a pull request
Uses a `ResizeObserver` to automatically scroll to the bottom of the list when needed.

## CLA

fixes #603


src/components/MessageList/__tests__/VirtualizedMessageList.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { cleanup } from '@testing-library/react';
 import renderer from 'react-test-renderer';
+import './ResizeObserver.mock';
```
mahboubii, line 4: should the file name end in .js?

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -51,6 +101,7 @@ const VirtualizedMessageList = ({
   );
   const mounted = useRef(false);
   const atBottom = useRef(false);
+  const lastMessageIndex = useRef(0);
```
mahboubii, line 104: maybe better to initialize with `messages.length` instead of zero?

mahboubii (approved):

src/components/MessageList/__tests__/VirtualizedMessageList.test.js
ath92: Actually named it like this deliberately because jest will run any file ending with .js and then complain that there are no tests in the file. Might be a little weird but this file is really only needed for jest

petyosi:
This PR is stale once we merge #694 - I have a few other "cheap" ideas on how to address the behavior, which we can discuss. Let's close.

PR #614, opened 2020-11-19
fix: prevent firing load more when user is offline

mahboubii:
not the most brilliant approach, but our WS is a little bit behind detecting network is down or not
well supported: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine


jaapbakker88 (approved):

ath92 (approved):
LGTM (I'm slightly weirded out by that camelCase for `onLine`)

PR #617, opened 2020-11-24
Feature: support channel search

rlundquist3:
## CLA

This is a workaround for `stream-chat`'s `channelQuery` filters not matching partial strings (see https://getstream.io/chat/docs/query_syntax/?language=js). It is a simple search that provides a bit more flexibility than the existing channel query capability in `stream-chat`. It could be used in the `ChannelSearch` component as well.
- adds `searchStr` prop to the `ChannelList` component
- `usePaginatedChannels` is passed this `searchStr` prop, and will only return channels that have a property in the channel data that matches the search string (case-insensitive, checks with and without punctuation)


mahboubii:
@rlundquist3 Thanks a lot for the contribution! we are going to add support for a more general custom filter function which should support your use case. /cc @DanC5

mahboubii:
hey @rlundquist3, please check out [v3.5.3](https://github.com/GetStream/stream-chat-react/releases/tag/v3.5.3). You can pass in a `channelRenderFilterFn` prop to ChannelList to filter or sort the channels.

PR #670, opened 2021-01-13
textarea-paste-limit

DanC5:
# Submit a pull request

## CLA


src/components/MessageInput/hooks/messageInput.js
```diff
@@ -699,14 +699,24 @@ export default function useMessageInput(props) {
...
+
         // fallback to regular text paste
         if (plainTextPromise) {
-          const s = await plainTextPromise;
-          insertText(s);
+          const { maxLength } = additionalTextareaProps;
+          const pastedText = await plainTextPromise;
+
+          if (pastedText.length > maxLength) {
+            dispatch({
```
ath92, line 709: Because we're not using the `insertText` callback here, any text that is already in the textarea at the time of pasting will be removed if the to-be-pasted text is longer than `maxLength`. I think that's kind of unexpected behavior; I'd expect whatever is pasted to just get truncated, but if there was any text in front of my cursor while pasting, it should still be there.

I think it might be better to move the logic that truncates the text into the `insertText` callback, and leave this function as is.

yetieaterxb1 (approved):

ath92 (approved):

PR #675, opened 2021-01-20
delay closeEmojiPicker event listener

DanC5:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/MessageInput/hooks/messageInput.js
+++ b/src/components/MessageInput/hooks/messageInput.js
@@ -355,6 +355,7 @@ export default function useMessageInput(props) {
       value: true,
     });
 
+    // prevents race condition where event listener picks up the click that opens the picker
     setTimeout(() =>
       document.addEventListener('click', closeEmojiPicker, false),
     );
--- a/src/components/MessageInput/hooks/messageInput.js
+++ b/src/components/MessageInput/hooks/messageInput.js
@@ -343,23 +343,27 @@ export default function useMessageInput(props) {
           type: 'setEmojiPickerIsOpen',
           value: false,
         });
-        document.removeEventListener('click', closeEmojiPicker, false);
       }
     },
     [emojiPickerRef],
   );
 
-  const openEmojiPicker = useCallback(() => {
+  const openEmojiPicker = useCallback((event) => {
     dispatch({
       type: 'setEmojiPickerIsOpen',
       value: true,
     });
 
-    // prevents race condition where event listener picks up the click that opens the picker
-    setTimeout(() =>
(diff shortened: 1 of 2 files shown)


src/components/MessageInput/hooks/messageInput.js
```diff
@@ -354,7 +354,10 @@ export default function useMessageInput(props) {
       type: 'setEmojiPickerIsOpen',
       value: true,
     });
-    document.addEventListener('click', closeEmojiPicker, false);
+
+    setTimeout(() =>
```
vishalnarkhede, line 358: Can we add a comment here as to why we need timeout?

DanC5: good call, added!

vishalnarkhede: How about useEffect?

```js
useEffect(() => {
    if (emojiPickerIsOpen) {
        document.addEventListener('click', closeEmojiPicker, false);
    }

    return () => {
        document.removeEventListener('click', closeEmojiPicker, false);
    }
}, [emojiPickerIsOpen])
```

ath92 (commented):
I tried this locally, but for me the current changes only allow me to open the emoji picker once (without the changes in this PR, it doesn't open at all).

I tried @vishalnarkhede 's solution with `useEffect`. I like it because it ensures that the event handler always gets cleaned up, which the current code doesn't. You can see this happening when you run a development server; as you make changes to the code while the picker is open, multiple event handlers are added to the document, because they're only cleaned up when you close the picker. Unfortunately, Vishal's solution doesn't seem to solve the race for me, and the picker doesn't open with only `useEffect`.

I'm still not quite sure how this used to work, but I think I found a clean solution:
```
  const closeEmojiPicker = useCallback(
    (e) => {
      if (
        emojiPickerRef.current &&
        !emojiPickerRef.current.contains(e.target)
      ) {
        dispatch({
          type: 'setEmojiPickerIsOpen',
          value: false,
        });
      }
    },
    [emojiPickerRef],
  );

  const openEmojiPicker = useCallback((e) => {
    dispatch({
      type: 'setEmojiPickerIsOpen',
      value: true,
    });
    // Prevent event from bubbling to document, so the close handler is never called for this event
    e.stopPropagation();
  }, []);

  useEffect(() => {
    if (state.emojiPickerIsOpen) {
      document.addEventListener('click', closeEmojiPicker, false);
    }
    return () => {
      document.removeEventListener('click', closeEmojiPicker, false);
    };
  }, [closeEmojiPicker, state.emojiPickerIsOpen]);
```

ath92 (approved):

PR #678, opened 2021-01-22
Custom channellist filter

yetieaterxb1:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/ChannelList/ChannelList.js
+++ b/src/components/ChannelList/ChannelList.js
@@ -118,8 +118,8 @@ const ChannelList = (props) => {
     activeChannelHandler,
   );
 
-  const loadedChannels = props.channelListFilterFunction
-    ? props.channelListFilterFunction(channels)
+  const loadedChannels = props.channelRenderFilterFn
+    ? props.channelRenderFilterFn(channels)
     : channels;
 
   useMobileNavigation(channelListRef, navOpen, closeMobileNav);
@@ -362,7 +362,7 @@ ChannelList.propTypes = {
    * @param {Array} channels
    * @returns {Array} channels
    * */
-  channelListFilterFunction: /** @type {PropTypes.Validator<(channels: import('stream-chat').Channel[]) => import('stream-chat').Channel[]>} */ (PropTypes.func),
+  channelRenderFilterFn: /** @type {PropTypes.Validator<(channels: import('stream-chat').Channel[]) => import('stream-chat').Channel[]>} */ (PropTypes.func),
   /**
    * Object containing query filters
    * @see See [Channel query documentation](https://getstream.io/chat/docs/query_channels/?language=js) for a list of available fields for filter.
--- a/src/components/ChannelList/ChannelList.js
+++ b/src/components/ChannelList/ChannelList.js
@@ -357,6 +357,10 @@ ChannelList.propTypes = {
    * */
   onChannelDeleted: PropTypes.func,
   /**
+   * Optional function for custom filtering channels.
+   * */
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -322,6 +322,8 @@ export interface ChannelListProps {
     e: Client.Event,
   ): void;
   setActiveChannelOnMount?: boolean;
+  /** Optional channel list filter */
```
DanC5, line 325:
```suggestion
  /** Optional function to filter channels prior to loading in the DOM */
```

src/components/ChannelList/ChannelList.js
```diff
@@ -352,6 +356,10 @@ ChannelList.propTypes = {
    * @param {Event}     event       [Event object](https://getstream.io/chat/docs/event_object/?language=js) corresponding to `channel.deleted` event
    * */
   onChannelDeleted: PropTypes.func,
+  /**
+   * Optional function for custom filtering channels.
```
DanC5, line 360:
```suggestion
   * Optional function to filter channels prior to loading in the DOM. Do not use any complex or async logic here that would significantly delay the loading of the ChannelList. We recommend array methods like filter/sort/reduce.
```

types/index.d.ts
```diff
@@ -322,6 +331,8 @@ export interface ChannelListProps {
...
+  filterFunction?: (channels: Array<Client.Channel>) => Array<Client.Channel>;
```
DanC5, line 335: you can make these StreamChatChannel now that we have that type with the generics

types/index.d.ts
```diff
@@ -55,6 +55,15 @@ export type StreamChatReactClient = Client.StreamChat<
   AnyType,
   StreamChatReactUserType
 >;
+export type StreamChatChannel = Client.Channel<
```
DanC5, line 58: since we're using the generic in the filterFunction type we no longer need this new type

src/components/ChannelList/ChannelList.js
```diff
@@ -352,6 +356,12 @@ ChannelList.propTypes = {
    * @param {Event}     event       [Event object](https://getstream.io/chat/docs/event_object/?language=js) corresponding to `channel.deleted` event
    * */
   onChannelDeleted: PropTypes.func,
+  /**
+   * Optional function to filter channels prior to loading in the DOM. Do not use any complex or async logic here that would significantly delay the loading of the ChannelList. We recommend array methods like filter/sort/reduce.
```
DanC5, line 360: you can prob newline this comment so it's a bit easier for us to read

mahboubii: also mention this function should be a pure function without any data mutations

types/index.d.ts
```diff
@@ -322,6 +331,8 @@ export interface ChannelListProps {
     e: Client.Event,
   ): void;
   setActiveChannelOnMount?: boolean;
+  /** Optional function to filter channels prior to loading in the DOM */
```
mahboubii, line 334: should be the same doc as the prop

types/index.d.ts
```diff
@@ -322,6 +331,8 @@ export interface ChannelListProps {
...
+  filterFunction?: <T>(channels: T) => T;
```
mahboubii, line 335: this should not be a generic IMO, what is passed in is an array of Immturable Channels and the return should also be the same type

mahboubii, line 335: can we think of a better name for this function? `filterFunction is not very intuitive`

DanC5 (approved):

PR #694, opened 2021-02-05
CRS-384 Upgrade react-virtuoso to v1

ath92:
# Submit a pull request

## CLA

Upgrades react-virtuoso to the latest version available. A refactor of `VirtualizedMessageList` was required to adapt to the breaking changes introduced with this new version. In particular, some extra work was needed to handle the new way virtuoso deals with prepending messages. 

Improvements:
- VirtualMessageList works well with `Message` and `FixedHeightMessage`
- Margins in the item content are handled
- New reactions keep the list at the bottom (if already there)

Not included:
- Images with no specified sizing displace the timeline. The same happens in non-virtualized mode.

Squash the commits before merging, a lot of back and forth.

### DIFF
--- a/examples/messaging/src/App.js
+++ b/examples/messaging/src/App.js
@@ -10,7 +10,7 @@ import {
   InfiniteScrollPaginator,
   MessageInput,
   MessageInputFlat,
-  MessageList as TheMessageList,
+  MessageList,
   Thread,
   Window,
 } from 'stream-chat-react';
@@ -61,7 +61,7 @@ const App = () => (
     <Channel>
       <Window>
         <ChannelHeader />
-        <TheMessageList />
+        <MessageList />
         <MessageInput Input={MessageInputFlat} focus />
       </Window>
       <Thread />
--- a/examples/messaging/src/App.js
+++ b/examples/messaging/src/App.js
@@ -10,7 +10,7 @@ import {
   InfiniteScrollPaginator,
   MessageInput,
   MessageInputFlat,
-  MessageList,
+  VirtualizedMessageList as TheMessageList,
   Thread,
   Window,
 } from 'stream-chat-react';
@@ -61,7 +61,7 @@ const App = () => (
     <Channel>
       <Window>
         <ChannelHeader />
-        <MessageList />
+        <TheMessageList />
         <MessageInput Input={MessageInputFlat} focus />
       </Window>
       <Thread />
--- a/examples/messaging/src/App.js
+++ b/examples/messaging/src/App.js
@@ -10,7 +10,7 @@ import {
   InfiniteScrollPaginator,
   MessageInput,
   MessageInputFlat,
-  VirtualizedMessageList as TheMessageList,
+  MessageList as TheMessageList,
   Thread,
   Window,
 } from 'stream-chat-react';
--- a/src/components/MessageList/VirtualizedMessageList.js
+++ b/src/components/MessageList/VirtualizedMessageList.js
@@ -40,6 +40,7 @@ const VirtualizedMessageList = ({
   TypingIndicator = null,
   LoadingIndicator = DefaultLoadingIndicator,
   EmptyStateIndicator = DefaultEmptyStateIndicator,
+  stickToBottomScrollBehavior = 'smooth',
 }) => {
   const { t } = useContext(TranslationContext);
 
@@ -144,10 +145,10 @@ const VirtualizedMessageList = ({
         overscan={overscan}
         followOutput={(isAtBottom) => {
           if (shouldForceScrollToBottom()) {
-            return isAtBottom ? 'smooth' : 'auto';
+            return isAtBottom ? stickToBottomScrollBehavior : 'auto';
           }
           // a message from another user has been received - don't scroll to bottom unless already there
-          return isAtBottom ? 'smooth' : false;
+          return isAtBottom ? stickToBottomScrollBehavior : false;
         }}
         itemContent={(i) => messageRenderer(messages, i)}
         components={virtuosoComponents}
--- a/src/components/MessageList/VirtualizedMessageList.js
+++ b/src/components/MessageList/VirtualizedMessageList.js
@@ -15,7 +15,7 @@ import { usePrependedMessagesCount } from './hooks/usePrependMessagesCount';
 import { useShouldForceScrollToBottom } from './hooks/useShouldForceScrollToBottom';
 import MessageNotification from './MessageNotification';
 
-const prependOffset = 10000;
+const PREPEND_OFFSET = 10 ** 7;
 
 /**
  * VirtualizedMessageList - This component renders a list of messages in a virtual list. Its a consumer of [Channel Context](https://getstream.github.io/stream-chat-react/#channel)
@@ -64,7 +64,7 @@ const VirtualizedMessageList = ({
   const messageRenderer = useCallback(
     (messageList, virtuosoIndex) => {
       const streamMessageIndex =
-        virtuosoIndex + numItemsPrepended - prependOffset;
+        virtuosoIndex + numItemsPrepended - PREPEND_OFFSET;
       // use custom renderer supplied by client if present and skip the rest
       if (customMessageRenderer) {
         return customMessageRenderer(messageList, streamMessageIndex);
@@ -152,7 +152,7 @@ const VirtualizedMessageList = ({
         }}
         itemContent={(i) => messageRenderer(messages, i)}
         components={virtuosoComponents}
-        firstItemIndex={prependOffset - numItemsPrepended}
+        firstItemIndex={PREPEND_OFFSET - numItemsPrepended}
         startReached={() => {
           if (hasMore) {
             loadMore(messageLimit);
--- a/src/components/MessageList/VirtualizedMessageList.js
+++ b/src/components/MessageList/VirtualizedMessageList.js
@@ -113,11 +113,19 @@ const VirtualizedMessageList = ({
       );
 
     /**
-     * setting 'overflow: auto' traps CSS margins of the item elements, preventing incorrect item measurements.
+     * using 'display: inline-block' traps CSS margins of the item elements, preventing incorrect item measurements.
      * @type {import('react-virtuoso').Components['Item']}
      */
     const Item = (props) => {
-      return <div {...props} style={{ overflow: 'auto' }} />;
+      return (
+        <div
+          {...props}
+          style={{
+            display: 'inline-block',
+            width: '100%',
+          }}
+        />
+      );
     };
 
     const Footer = () => {
@@ -143,6 +151,7 @@ const VirtualizedMessageList = ({
         ref={virtuoso}
         totalCount={messages.length}
         overscan={overscan}
+        style={{ overflowX: 'hidden' }}
         followOutput={(isAtBottom) => {
           if (shouldForceScrollToBottom()) {
             return isAtBottom ? stickToBottomScrollBehavior : 'auto';
@@ -150,7 +159,9 @@ const VirtualizedMessageList = ({
           // a message from another user has been received - don't scroll to bottom unless already there
           return isAtBottom ? stickToBottomScrollBehavior : false;
         }}
-        itemContent={(i) => messageRenderer(messages, i)}
+        itemContent={(i) => {
+          return messageRenderer(messages, i);
+        }}
         components={virtuosoComponents}
         firstItemIndex={PREPEND_OFFSET - numItemsPrepended}
         startReached={() => {
--- a/src/components/MessageList/VirtualizedMessageList.js
+++ b/src/components/MessageList/VirtualizedMessageList.js
@@ -1,27 +1,24 @@
 // @ts-check
-import React, {
-  useContext,
-  useState,
-  useEffect,
-  useCallback,
-  useRef,
-} from 'react';
+import React, { useCallback, useContext, useMemo, useRef } from 'react';
 import { Virtuoso } from 'react-virtuoso';
-
-import { smartRender } from '../../utils';
-import MessageNotification from './MessageNotification';
 import { ChannelContext, TranslationContext } from '../../context';
+import { smartRender } from '../../utils';
+import { EmptyStateIndicator as DefaultEmptyStateIndicator } from '../EmptyStateIndicator';
 import { EventComponent } from '../EventComponent';
 import { LoadingIndicator as DefaultLoadingIndicator } from '../Loading';
-import { EmptyStateIndicator as DefaultEmptyStateIndicator } from '../EmptyStateIndicator';
 import {
-  FixedHeightMessage,
+  FixedHeightMessage as DefaultMessage,
   MessageDeleted as DefaultMessageDeleted,
 } from '../Message';
+import { useNewMessageNotification } from './hooks/useNewMessageNotification';
+import { usePrependedMessagesCount } from './hooks/usePrependMessagesCount';
+import { useShouldForceScrollToBottom } from './hooks/useShouldForceScrollToBottom';
+import MessageNotification from './MessageNotification';
+
+const prependOffset = 10000;
 
 /**
  * VirtualizedMessageList - This component renders a list of messages in a virtual list. Its a consumer of [Channel Context](https://getstream.github.io/stream-chat-react/#channel)
- * It is pretty fast for rendering thousands of messages but it needs its Message component to have fixed height
  * @example ../../docs/VirtualizedMessageList.md
  * @type {React.FC<import('types').VirtualizedMessageListInternalProps>}
  */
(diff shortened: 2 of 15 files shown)


src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -108,15 +104,46 @@ const VirtualizedMessageList = ({
...
-            message.user.id === messageList[i - 1].user.id
+            streamMessagesIndex > 0 &&
+            message.user.id === messageList[streamMessagesIndex - 1].user.id
           }
         />
       );
     },
-    [MessageDeleted, customMessageRenderer, shouldGroupByUser],
+    [
+      MessageDeleted,
```
DanC5, line 114: can probably remove MessageDeleted from dep array, it shouldn't change

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -108,15 +104,46 @@ const VirtualizedMessageList = ({
...
+      TypingIndicator ? <TypingIndicator avatarSize={24} /> : <></>;
+
+    return {
+      EmptyPlaceholder,
+      ScrollSeekPlaceHolder: scrollSeekPlaceHolder,
+      Header,
+      Footer,
+    };
+  }, [
+    EmptyStateIndicator,
```
DanC5, line 141: same with EmptyState and Typing

DanC5:
@ath92 I'll admit I don't have much background on this feature but it looks pretty good, will wait for Amin to check it out too. Though in the meantime can you replace @ts-ignore with @ts-expect-error? This will help us when we get to converting this file and will notify you if there's not a type error to follow.

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -78,23 +79,18 @@ const VirtualizedMessageList = ({
...
 
+  const [numItemsPrepended, setNumItemsPrepended] = useState(0);
   const messageRenderer = useCallback(
-    (messageList, i) => {
+    (messageList, virtuosoIndex) => {
+      const streamMessagesIndex = virtuosoIndex + numItemsPrepended;
       // use custom renderer supplied by client if present and skip the rest
-      if (customMessageRenderer) return customMessageRenderer(messageList, i);
+      if (customMessageRenderer)
+        return customMessageRenderer(messageList, streamMessagesIndex);
```
DanC5, line 91: lets wrap in brackets when an inline return gets newlined

mahboubii:
looks like the new `scrollTo` function also accepts `smooth` behavior

src/components/MessageList/VirtualizedMessageList.js
ath92: Not sure if I'd remove this. Even if it shouldn't change this, it technically could, and it doesn't really hurt to have it in the dependencies array.

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -125,27 +152,14 @@ const VirtualizedMessageList = ({
...
-          ) : (
-            <></>
-          )
-        }
-        footer={() =>
-          TypingIndicator ? <TypingIndicator avatarSize={24} /> : <></>
-        }
+        itemContent={(i) => messageRenderer(messages, i)}
+        components={virtuosoComponents}
+        firstItemIndex={-numItemsPrepended}
```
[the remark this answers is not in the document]
ath92: This one's a bit funny, I'm not sure what's the best way to approach this. The virtuoso examples for this all have a large, mostly empty array as the initial state. As more items are backfilled (i.e. our loadMore function), the examples add these prepended items to the empty array, and the `firstItemIndex` is set to the first index in this mostly empty array that has actual messages. See https://virtuoso.dev/prepend-items/

I'd rather not keep a separate copy of our `messages` array with a bunch of empty slots at the start as state here, that feels very wasteful. I can increase the `firstItemIndex` by some arbitrary amount `x`, only to subtract `x` from the index when we use it to access the `messages` array, but I don't know why that's needed aside from the warning in the virtuoso docs.

To clarify, here's what I'm suggesting we could do:
```
const prependOffset = 10000;

// ....


  const messageRenderer = useCallback(
    (messageList, virtuosoIndex) => {
      const streamMessagesIndex = virtuosoIndex + numItemsPrepended - prependOffset;

// ....

      <Virtuoso
        itemContent={(i) => messageRenderer(messages, i)}
        components={virtuosoComponents}
        firstItemIndex={prependOffset - numItemsPrepended}

```
This would keep the index above 0 for the first 10000 messages that are preloaded. But I'm not sure what the difference would be; it seems to work exactly the same way as the current code.

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -125,27 +152,14 @@ const VirtualizedMessageList = ({
         totalCount={messages.length}
         overscan={overscan}
         followOutput={true}
-        maxHeightCacheSize={2000} // reset the cache once it reaches 2k
-        scrollSeek={scrollSeekPlaceHolder}
```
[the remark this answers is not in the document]
ath92: Hmm, so I updated the `components` prop to include the `scrollSeekPlaceholder`, but the `ScrollSeekConfiguration` is new. Should we allow SDK users to override this virtuoso prop? Should we provide a default value?

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
mahboubii: Interesting, I wonder why it specifically says to use positive values here, maybe there are some gotchas we're missing.
but having sth like `const [firstItemIndex, setFirstItemIndex] = useState(50000);` and reducing this number everytime we load more should be good for now

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
mahboubii: we don't have a default one for now, I'd just check and break the user passed prop into UI and config until we go with a major release and update the API

src/components/MessageList/VirtualizedMessageList.js
DanC5: I think this raises a larger question of how we handle dependency arrays. On the react native side we only include the data that should actually trigger a new render during an app session, so we don't include things like prop UI components that don't change mid-session. Obviously native is a little different, but we definitely saw performance improvements and decreased re-renders upon getting stricter with the dependency arrays. @mahboubii do you have thoughts on whether or not to keep UI components in dependency arrays?

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
petyosi: Sorry for the inconvenience on that. Internal boundaries are affected by a negative firstItemIndex :(. I will give this some thought, but in the meantime, using an arbitrary large number should work fine.

src/components/MessageList/VirtualizedMessageList.js
mahboubii: There could be customers who change the UI props conditionally in their app but the performance gain from removing this dependency check is worth it IMO and we should move toward that path. 
In the meantime since changing these effect dependencies are potentially a breaking change, let's leave it as is until the next major release.
we should make it clear for customers that UI prop overrides should not change after the first render of the app

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
mahboubii: btw great job on v1 @petyosi 👍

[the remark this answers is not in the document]
ath92: For sure! I'm not against using a positive index, it just wasn't quite clear to me from the docs why it's needed. I'd love to get a deeper understanding of Virtuoso btw, looks like there's some cool stuff going on under the hood but I find it a little difficult to dig through the source code at the moment because I've never done much with rx-like architectures.

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -125,27 +152,14 @@ const VirtualizedMessageList = ({
         totalCount={messages.length}
         overscan={overscan}
         followOutput={true}
```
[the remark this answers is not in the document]
ath92: `alignToBottom` seems nice, but it's not consistent with our other messagelists?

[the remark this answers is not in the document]
ath92: Also, `smooth` would be cool in most cases but the virtuoso docs mention that it may not be able to keep up with many messages coming in quickly. Might be risky for livestream customers?

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
ath92: I've updated the code now, thanks for pointing this out because I didn't realise there was so much going on in that prop. Somehow the typescript definitions also got a little messed up so tsc didn't complain that I missed a bunch of stuff there. Should be better now, but I do think the code looks a bit ugly. Definitely something we'll want to fix the next time we make breaking changes.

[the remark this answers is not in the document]
mahboubii: awesome thanks, can you also leave a todo comment so we won't forget about it later

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
mahboubii: we can put in the default and allow customers to override them if it's not working for them

src/components/MessageList/VirtualizedMessageList.js
ath92: Yeah, I added this to the component:
```
// TODO: split scrollSeekPlaceholder into two props (e.g. ScrollSeekPlaceholder and scrollSeekConfiguration) when making breaking changes
```

and this to the typedefs:
```
*  Note: virtuoso has broken out the placeholder value and instead includes it in its components prop.
*  TODO: break out placeholder when making other breaking changes.
```

src/components/MessageList/VirtualizedMessageList.js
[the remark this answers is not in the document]
ath92: Yeah, I added this to the component:
```
// TODO: split scrollSeekPlaceholder into two props (e.g. ScrollSeekPlaceholder and scrollSeekConfiguration) when making breaking changes
```

and this to the typedefs:
```
*  Note: virtuoso has broken out the placeholder value and instead includes it in its components prop.
*  TODO: break out placeholder when making other breaking changes.
```

ath92:
I think I've addressed the comments in this review so far, but I'm running into an issue: sometimes the `startReached` callback is called twice (one after the other, so after the first promise resolves, the function is called again to load more items). This causes the list to jump to the latest set of items that has been loaded (because `firstItemIndex` is changed when the last set of messages is loaded).

I'm not sure why this happens - as far as I can tell we're not doing anything different from the examples provided on the Virtuoso website. Perhaps @petyosi has some idea on why this could be happening? (let me know if it makes more sense to try and create a more minimal example of the problem and open an issue in the virtuoso repo instead)

petyosi:
> I think I've addressed the comments in this review so far, but I'm running into an issue: sometimes the `startReached` callback is called twice (one after the other, so after the first promise resolves, the function is called again to load more items). This causes the list to jump to the latest set of items that has been loaded (because `firstItemIndex` is changed when the last set of messages is loaded).
> 
> I'm not sure why this happens - as far as I can tell we're not doing anything different from the examples provided on the Virtuoso website. Perhaps @petyosi has some idea on why this could be happening? (let me know if it makes more sense to try and create a more minimal example of the problem and open an issue in the virtuoso repo instead)

I have been struggling with this problem in this issue: https://github.com/petyosi/react-virtuoso/issues/281 - check if debouncing the callback handle won't resolve it. If not, a repro will help me. Notice that the OS matters - if you reproduce it on Mac, then you are probably facing something different.

ath92:
> I have been struggling with this problem in this issue: [petyosi/react-virtuoso#281](https://github.com/petyosi/react-virtuoso/issues/281) - check if debouncing the callback handle won't resolve it. If not, a repro will help me. Notice that the OS matters - if you reproduce it on Mac, then you are probably facing something different.

Got some time to look into this again: The issue I was seeing was unrelated to the issue you mentioned here (and unrelated to virtuoso), but caused by how we handle prepending items to our channel state. 

In short: the promise that is returned by `loadMore` contains the amount of new messages, but those new messages aren't added to the channel state at exactly the moment the promise resolves: the actual state update is performed by a debounced `loadMoreFinished` function in the Channel component. In some cases, this caused the list to be re-rendered with new messages, while `firstItemIndex` was not yet updated, or the other way around (`firstItemIndex` already updated, but no new messages yet). 

I pushed a workaround for this that doesn't depend on the timing of the `loadMore` promise resolving, but I'm not super happy with it because it's kinda complex. I don't immediately see an easier way to fix it though, but open to suggestions.

petyosi:
@mahboubii @DanC5 @virdesai please take a look

src/components/MessageList/hooks/usePrependMessagesCount.js
```diff
@@ -0,0 +1,44 @@
+// @ts-check
+import { useRef, useMemo } from 'react';
+
+/**
+ * @param {Array<{ id: string }> | undefined} messages,
+ */
+export const usePrependedMessagesCount = (messages) => {
+  const firstMessageId = useRef(messages && messages.length && messages[0].id);
```
mahboubii, line 8:
```suggestion
  const firstMessageId = useRef(messages?.[0]?.id);
```

src/components/MessageList/hooks/usePrependMessagesCount.js
```diff
@@ -0,0 +1,44 @@
...
+  const earliestMessageId = useRef(
+    messages && messages.length && messages[0].id,
+  );
```
mahboubii, line 11:
```suggestion
  const earliestMessageId = useRef(messages?.[0]?.id);
```

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -1,27 +1,24 @@
...
-  FixedHeightMessage,
+  FixedHeightMessage as DefaultMessage,
   MessageDeleted as DefaultMessageDeleted,
 } from '../Message';
+import { useNewMessageNotification } from './hooks/useNewMessageNotification';
+import { usePrependedMessagesCount } from './hooks/usePrependMessagesCount';
+import { useShouldForceScrollToBottom } from './hooks/useShouldForceScrollToBottom';
+import MessageNotification from './MessageNotification';
+
+const prependOffset = 10000;
```
mahboubii, line 18: should we increase this to 50k just to be on the safe side? what happens if we prepend more than 10k?

mahboubii (approved):

petyosi:
@DanC5 let's give this a review when possible

examples/messaging/src/App.js
```diff
@@ -10,7 +10,7 @@ import {
   InfiniteScrollPaginator,
   MessageInput,
   MessageInputFlat,
-  MessageList,
+  MessageList as TheMessageList,
```
DanC5, line 13: do we need this name reassignment?

DanC5 (approved):

examples/messaging/src/App.js
petyosi: No, not really, it was a convenience for testing. I reverted the messaging app tweaks.

DanC5: great looks good, will merge

src/components/MessageList/VirtualizedMessageList.js
```diff
@@ -32,91 +29,117 @@ const VirtualizedMessageList = ({
...
   loadingMore,
   messageLimit = 100,
-  overscan = 200,
+  overscan = 0,
   shouldGroupByUser = false,
   customMessageRenderer,
+  // TODO: refactor to scrollSeekPlaceHolderConfiguration and components.ScrollSeekPlaceholder, like the Virtuoso Component
   scrollSeekPlaceHolder,
-  Message = FixedHeightMessage,
+  Message = DefaultMessage,
```
kainosnoema, line 37: Seeing that a fixed height message might not be required anymore, is it possible to use the `MessageTeam` component in the virtualized list now?

Looks like this Message property requires a type with `FixedHeightMessageProps` still, so not sure how to do it...

DanC5: @petyosi can you speak to the feasibility of using our other Message UI components in the VL now?

petyosi: @kainosnoema indeed, using "any" messages in the virtual list is something which we are looking into, but not yet ready to introduce as the default configuration. The part which does not work as smoothly as we wish is messages which contain `img` elements without a specified size. Such posts may cause the virtual list to "jump" when the image is loaded.

This being said, you can ignore the typing for now and try passing the plain message - we would love to get your feedback on the performance of that.

kainosnoema: @petyosi thanks for the update. We're building our own custom message component and will be having photo galleries with a fixed height (when attachments are present), so don't think we'll run into the jumping issue. We'll give this a shot!

PR #708, opened 2021-02-12
TS/channel-context

DanC5:
# Submit a pull request

## CLA


src/context/ChannelContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+export type ChannelContextValue<
+  At extends UnknownType = DefaultAttachmentType,
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+> = ChannelState<At, Ch, Co, Ev, Me, Re, Us> & {
+  client: StreamChat<At, Ch, Co, Ev, Me, Re, Us>;
```
virdesai, line 145: why is client on ChannelContext? we should keep it only in ChatContext. it'll get confusing if we're listening to events on 2 potentially different clients

DanC5: don't know why it's passed in there, but didn't want to pull out and potentially cause a break. figured we'd just rip out when we reorg context after TS

src/context/ChannelContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+  StreamChat,
+  UpdateMessageAPIResponse,
+  UserResponse,
+} from 'stream-chat';
+import type {
+  Data as EmojiMartData,
+  NimbleEmojiIndex,
+  NimbleEmojiProps,
+  NimblePickerProps,
+} from 'emoji-mart';
```
virdesai, line 19: on RN the only things that were out of abc order were imports from react/react-native and those went first, everything else, including our stream-chat stuff, was abc ordered

DanC5: oh got it, thought we were giving stream stuff precedence but i'll swap em so the libs are consistent

src/context/ChannelContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+        type: string;
+      }
+    | {
+        asset_url: string;
+        file_size: number;
+        mime_type: string;
+        title: string;
+        type: string;
+      }
+  >;
```
virdesai, line 91: what's up with this? should these be thrown on the default attachment type as optionals instead of done like this?

virdesai (approved):

src/context/ChannelContext.tsx
DanC5: looks like only ones not on the Attachment type are mime_type and file_size, will just add these as optional and extend the default type

PR #713, opened 2021-02-17
typescript CRS-336, CRS-337, CRS-338, CRS-345

virdesai:
# Submit a pull request

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices
- [ ] Code changes are tested (add some information if not applicable)

## Description of the pull request

Tickets Addressed:
- [CRS-336](https://stream-io.atlassian.net/browse/CRS-336)
- [CRS-337](https://stream-io.atlassian.net/browse/CRS-337)
- [CRS-338](https://stream-io.atlassian.net/browse/CRS-338)
- [CRS-345](https://stream-io.atlassian.net/browse/CRS-345)

Typescript'd the following components:
- Avatar
- ChannelList
- ChannelPreview
- ChannelSearch
- ChatDown
- EmptyStateIndicator
- InfiniteScrollPaginator
- Loading
- LoadMore
- UserItem

PR #716, opened 2021-02-18
TS/Channel

DanC5:
# Submit a pull request

## CLA


src/components/Channel/hooks/useMentionsHandlers.ts
```diff
@@ -0,0 +1,51 @@
...
+) =>
+  useCallback(
+    (event: MouseEvent<HTMLElement>, mentioned_users: UserResponse<Us>[]) => {
+      if (!onMentionsHover && !onMentionsClick) return;
+
+      const target = event.target;
+      // @ts-expect-error
+      const tagName = target?.tagName?.toLowerCase() as string;
+      // @ts-expect-error
+      const textContent = target?.innerHTML?.replace('*', '') as string;
```
petyosi, line 23:
```suggestion
      if (!(event.target instanceof HTMLElement)) return;
      
      const target = event.target;
      const tagName = target.tagName.toLowerCase() as string;
      const textContent = target.innerHTML.replace('*', '') as string;
```

I like that approach to narrow down the target type. Throwing an error instead of `return` also works.

src/components/Channel/Channel.tsx
```diff
@@ -197,8 +400,8 @@ const ChannelInner = ({
     }
   }, [state.messages, state.thread]);
 
-  // Message
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+  /** MESSAGE */
+
```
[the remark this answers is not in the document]
DanC5: oh good catch, I missed the removal of that one, will update

src/components/Channel/hooks/useMentionsHandlers.ts
DanC5: I like this better as well, also no longer need the string casting

src/components/Channel/emojiData.ts
```diff
@@ -1,20 +1,22 @@
-export const emojiSetDef = {
+import type { Data as EmojiData } from 'emoji-mart';
+
+import type { CommonEmoji, EmojiSetDef, MinimalEmoji } from '../../context';
```
virdesai, line 3: this is with the intention that we get rid of a lot of these BS index.js/ts files that make VSCode command searching for files a little annoying
```suggestion
import type { CommonEmoji, EmojiSetDef, MinimalEmoji } from '../../context/ChannelContext';
```

src/components/Channel/hooks/index.ts
```diff
@@ -0,0 +1,2 @@
+export { useEditMessageHandler as useChannelEditMessageHandler } from './useEditMessageHandler';
+export { useMentionsHandlers as useChannelMentionsHandler } from './useMentionsHandlers';
```
virdesai, line 2: why not just rename our files and exports to this in general then?

src/components/Channel/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Channel';
+export * from './hooks';
```
virdesai, line 2: call each of the hooks individually like https://github.com/GetStream/stream-chat-react/pull/713/files#diff-7fd43b0b162a214d655a8176a345b983bfa23ff95be8e4059e587cf9db3fc51f so we can get rid of a bunch of `index.ts/ts` files eventually

src/components/Channel/channelState.ts
```diff
@@ -1,5 +1,117 @@
-/** @type {import('./types').ChannelStateReducer} */
-export const channelReducer = (state, action) => {
+import type { Reducer } from 'react';
+import type {
+  Channel,
+  MessageResponse,
+  ChannelState as StreamChannelState,
+} from 'stream-chat';
+
+import type { ChannelState } from '../../context';
```
virdesai, line 8:
```suggestion
import type { ChannelState } from '../../context/ChannelContext';
```

src/components/Channel/channelState.ts
```diff
@@ -111,7 +234,6 @@ export const channelReducer = (state, action) => {
   }
 };
 
-/** @type {import('./types').ChannelState} */
 export const initialState = {
```
virdesai, line 237: should this have `initialState: ChannelState`?

src/components/Channel/hooks/useEditMessageHandler.ts
```diff
@@ -0,0 +1,50 @@
+import { useChatContext } from '../../../context';
```
virdesai, line 1:
```suggestion
import { useChatContext } from '../../../context/ChatContext';
```

src/components/Channel/hooks/useMentionsHandlers.ts
```diff
@@ -0,0 +1,52 @@
+import { MouseEvent, useCallback } from 'react';
+import type { UserResponse } from 'stream-chat';
```
virdesai, line 2: also does `MouseEvent` need to be imported here individually because `React` default isn't being pulled off of this import?
```suggestion
import { MouseEvent, useCallback } from 'react';

import type { UserResponse } from 'stream-chat';
```

src/components/Channel/Channel.tsx
```diff
@@ -14,29 +14,208 @@ import DefaultEmojiIndex from 'emoji-mart/dist-modern/utils/emoji-index/nimble-e
...
-
-import { ChannelContext, ChatContext, TranslationContext } from '../../context';
+import {
+  ChannelContextValue,
+  ChannelProvider,
+  MessageAttachments,
+  MessageToSend,
+  useChatContext,
+  useTranslationContext,
+} from '../../context';
```
virdesai, line 59: split this up into the individual imports for each `Channel/Chat/Translation` context so we can get rid of the `index.ts/js` files

src/components/Channel/Channel.tsx
```diff
@@ -14,29 +14,208 @@ import DefaultEmojiIndex from 'emoji-mart/dist-modern/utils/emoji-index/nimble-e
...
+   *
+   * Defaults to and accepts same props as: [Attachment](https://github.com/GetStream/stream-chat-react/blob/master/src/components/Attachment.js)
+   * */
+  Attachment?: React.ComponentType<unknown>; // TODO - add generic when Attachment pr merged
+  /** The connected and active channel */
+  channel?: StreamChannel<At, Ch, Co, Ev, Me, Re, Us>;
+  /**
+   * Override mark channel read request (Advanced usage only)
+   *
+   * @param {Channel} channel object
```
virdesai, line 102: this param stuff is no longer necessary

src/components/Channel/Channel.tsx
```diff
@@ -14,29 +14,208 @@ import DefaultEmojiIndex from 'emoji-mart/dist-modern/utils/emoji-index/nimble-e
...
+   * */
+  doMarkReadRequest?: (
+    channel: StreamChannel<At, Ch, Co, Ev, Me, Re, Us>,
+  ) => Promise<MessageResponse<At, Ch, Co, Me, Re, Us>> | void;
+  /** Override send message request (Advanced usage only)
+   *
+   * @param {String} channelId full channel ID in format of `type:id`
+   * @param {Object} message
```
virdesai, line 110: this param stuff is no longer necessary

src/components/Channel/Channel.tsx
```diff
@@ -14,29 +14,208 @@ import DefaultEmojiIndex from 'emoji-mart/dist-modern/utils/emoji-index/nimble-e
...
+  doSendMessageRequest?: (
+    channelId: string,
+    message: Message<At, Me, Us>,
+  ) => ReturnType<
+    StreamChannel<At, Ch, Co, Ev, Me, Re, Us>['sendMessage']
+  > | void;
+  /** Override update(edit) message request (Advanced usage only)
+   *
+   * @param {String} channelId full channel ID in format of `type:id`
+   * @param {Object} updatedMessage
```
virdesai, line 121: this param stuff is no longer necessary

src/components/Channel/Channel.tsx
```diff
@@ -14,29 +14,208 @@ import DefaultEmojiIndex from 'emoji-mart/dist-modern/utils/emoji-index/nimble-e
...
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+>(
+  props: PropsWithChildren<ChannelProps<At, Ch, Co, Ev, Me, Re, Us>>,
+) => {
+  const { EmptyPlaceholder = null } = props;
```
virdesai, line 208: I see us using `props.channel` or other stuff later referenced off of `props`. Let's just dereference props up here for the values that are used within this file so we can save some on the lookup cost during runtime from doing `props.someVariable`

src/components/Channel/Channel.tsx
```diff
@@ -45,25 +224,46 @@ const Channel = ({ EmptyPlaceholder = null, ...props }) => {
...
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+>(
+  props: PropsWithChildren<ChannelProps<At, Ch, Co, Ev, Me, Re, Us>> & {
+    key: string;
+  },
```
virdesai, line 238: not gonna be properly formatted with prettier but think we should do it this way
```suggestion
  props: PropsWithChildren<ChannelProps<At, Ch, Co, Ev, Me, Re, Us> & {
    key: string;
  }>,
```

src/components/Channel/Channel.tsx
```diff
@@ -142,52 +355,55 @@ const ChannelInner = ({
...
         client.on('connection.recovered', handleEvent);
         channel.on(handleEvent);
       }
     })();
+
     return () => {
       if (errored || !done) return;
       document.removeEventListener('visibilitychange', onVisibilityChange);
-      channel.off(handleEvent);
+      channel?.off(handleEvent);
```
virdesai, line 399:
```suggestion
      if (channel) {
        channel.off(handleEvent);
      }
```

src/components/Channel/Channel.tsx
```diff
@@ -219,17 +434,21 @@ const ChannelInner = ({
   );
 
   const loadMore = useCallback(
-    async (limit = 100) => {
-      if (!online.current || !window.navigator.onLine) return 0;
+    // eslint-disable-next-line @typescript-eslint/no-inferrable-types
```
virdesai, line 437: ?

src/components/Channel/Channel.tsx
```diff
@@ -321,13 +555,20 @@ const ChannelInner = ({
...
     [client.user, client.userID],
   );
 
   const sendMessage = useCallback(
-    async ({ text, attachments = [], mentioned_users = [], parent }) => {
+    async ({
+      text = '',
+      attachments = [],
+      mentioned_users = [],
+      parent = undefined,
```
virdesai, line 568: abc 😆 
```suggestion
      attachments = [],
      mentioned_users = [],
      parent = undefined,
      text = '',
```

src/components/Channel/hooks/index.ts
DanC5: I believe it's to differentiate between the edit and mentions handlers that are exported from Message

src/components/Channel/channelState.ts
DanC5: since ChannelState needs generics throwing this on there led to typing errors

src/components/Channel/hooks/useMentionsHandlers.ts
DanC5: yep no jsx so no need for React

virdesai: yeah I got the lack of need for React but I meant is importing MouseEvent necessary because the default isn't being imported because otherwise it's a natural type that is present

src/components/Channel/Channel.tsx
DanC5: some weirdness happening with the useCallback, it will infer the type as number but then later in the file it's referenced as any

petyosi (approved):

virdesai (approved):

PR #723, opened 2021-02-23
Mobile Options onClick

DanC5:
The MessageText component triggers the display of the MessageOptions (react, reply, actions) onHover. This created some issues on mobile where hover events do not exist. We now disable hover on mobile and trigger an onClick event to show the MessageOptions when a user presses the MessageText component.

### DIFF
--- a/src/components/Message/MessageText.js
+++ b/src/components/Message/MessageText.js
@@ -12,7 +12,11 @@ import {
   useReactionClick,
   useMentionsUIHandler,
 } from './hooks';
-import { messageHasReactions, messageHasAttachments } from './utils';
+import {
+  handleMobilePress,
+  messageHasReactions,
+  messageHasAttachments,
+} from './utils';
 import MessageOptions from './MessageOptions';
 
 /**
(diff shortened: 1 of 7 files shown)


petyosi:
Testing this on iPhone, I see that activating an option for one message does not hide the existing ones:
image

Also, (not a critical problem, and maybe not a good idea to disable it), the text selection is on, so my taps sometimes select the text.

DanC5:
@petyosi I did notice both of those issues, but this gets to a bigger conversation around the nature of our Message UI components. We've seen that Message UI components by far require more customization than any other part of a chat app. So for this reason we've created four sample UI components for clients to use for MVP purposes and to give them a roadmap for building their own Message UI component. The idea is once the client has handled all their integration steps and has their app working, then they will take on the task of building their own Message via the prop on the MessageList. I was speaking with a client on this very issue yesterday and told him we can't incorporate every bit of feedback into our sample components as they're simply meant to be guides. That said, giving it more thought I think it does make sense to have something like an outside click listener that closes the MessageOptions component when a click is registered outside its parent message. I'll look into that after taking on a task for IBM.

virdesai (approved):
no clue what options onClick is but okay

DanC5:
@petyosi added a new hook to handle the mobile press function and added a click off listener in there as well, all seems to work pretty well in toggling the message options

PR #777, opened 2021-03-12
Fix an error when a hyphen causes a false positive

felixclack:
The `search` function on `NimbleEmojiIndex` can return an array with
a null value. This in turn causes an error when attempting to
destructure that element later on.

I wasn't able to write a test that highlights this error. I found it
difficult to replace the dependency on NimbleEmojiIndex with a mock.

Open to pointers on how to do that so I can complete this with an actual
failing test.

In the meantime, ensuring that `found` only contains non-null values
will fix the error.

## CLA


DanC5:
@felixclack thanks for the contribution! We're actually about to make a TypeScript release of our library and this quick fix has already been implemented. I'm going to close this PR as a new release will be made within a week or so, thanks.

PR #785, opened 2021-03-17
Support for date separators in virtual message list

petyosi:
This PR introduces the `disableDateSeparator` property for the virtual message list. To avoid breaking changes, the default value of the property is `true`.


DanC5:
closing in favor of the code already merged to conversion branch

PR #786, opened 2021-03-17
Support date separators in virtual message list

petyosi:
Adds disableDateSeparator property for the virtual message list. To avoid breaking changes, the default value of the property is `true`.

### DIFF
--- a/src/components/MessageList/VirtualizedMessageList.tsx
+++ b/src/components/MessageList/VirtualizedMessageList.tsx
@@ -11,7 +11,9 @@ import { useNewMessageNotification } from './hooks/useNewMessageNotification';
 import { usePrependedMessagesCount } from './hooks/usePrependMessagesCount';
 import { useShouldForceScrollToBottom } from './hooks/useShouldForceScrollToBottom';
 import { MessageNotification } from './MessageNotification';
+import { insertDates } from './utils';
 
+import type { DateSeparatorProps } from '../DateSeparator/DateSeparator';
 import {
   EmptyStateIndicator as DefaultEmptyStateIndicator,
   EmptyStateIndicatorProps,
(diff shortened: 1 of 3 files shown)


src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -327,6 +361,8 @@ export function VirtualizedMessageList<
 
   return (
     <VirtualizedMessageListWithContext
+      //@ts-expect-error
```
DanC5, line 364: assuming this is because channel is optionally undefined?

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -159,6 +174,18 @@ const VirtualizedMessageListWithContext = <
     TypingIndicator = null,
   } = props;
 
+  const { userID } = client;
+  const lastRead = useMemo(() => channel.lastRead(), [channel]);
```
DanC5, line 178:
```suggestion
  const lastRead = useMemo(() => channel?.lastRead(), [channel]);
```

can't we do this to handle channel being a partial type?

DanC5 (approved):

src/components/MessageList/VirtualizedMessageList.tsx
petyosi: Question: Is rendering a list without a channel a valid scenario? If not, I would rather terminate this possible timeline as early as possible rather than optioning variables further down.

DanC5: It should never happen, but the channel is listed as optional because there are two different ways it can be set, either through props (not using ChannelList) or through context (using ChannelList). So both of these channel types are optional since it can happen either way, but one of them should exist. Looks like we just need to adjust the ChannelInner props to not expect a partial `channel` type, since we early return before this render anyways if the channel doesn't exist

PR #793, opened 2021-03-19
CRS-441 customizable message mentions

ath92:
Renders mentions as span with `str-chat__message-mention` class instead of just wrapping it in a `<strong>`. Also allows users to override the wrapping component for mentions by supplying a custom mentions component to the `renderText` function.

# Submit a pull request

## CLA

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -107,9 +107,30 @@ const emojiMarkdownPlugin = () => {
   return transform;
 };
 
+const mentionsMarkdownPlugin = (mentionedUsersRegex: RegExp) => () => {
+  function replace(match: RegExpMatchArray | null) {
+    return {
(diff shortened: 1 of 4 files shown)


src/utils.tsx
```diff
@@ -107,9 +107,51 @@ const emojiMarkdownPlugin = () => {
   return transform;
 };
 
-export const renderText = <Us extends DefaultUserType<Us> = DefaultUserType>(
+type MentionedUser<
+  Us extends DefaultUserType<Us> = DefaultUserType
+> = UserResponse<Us>;
```
DanC5, line 112: we shouldn't need this type, because in order to actually use it with the generic you'd have to call it like MentionedUser<Us>, so it would be identical to UserResponse<Us>

DanC5 (approved):

PR #801, opened 2021-03-23
Connect on user click

wmadden:
# Submit a pull request

## CLA

The `Avatar` component accepts `onClick` and `onMouseOver` props which are set in the `Message` components to the props `onUserClick` and `onUserHover`. It appears these are expected to come from the `ChannelContext` - or at least from the `MessageList` - but they don't.

This pull request connects the `Channel` and `MessageList` and passes `onUserClick` and `onUserHover` down so that it's possible to detect a click event on the `Avatar` in the `MessageList`.

Note that this does not affect other usages of the `Avatar`, where it's unlikely the click event would be useful, or that it would be interpreted as a "user click". E.g. in the channel header you'd probably want to invoke different behavior when the user clicks on the avatar.

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -19,6 +19,8 @@ export type AvatarProps = {
   size?: number;
 };
 
+const noop = () => undefined;
+
 /**
  * Avatar - A round avatar image with fallback to username's first letter.
  * @example ./Avatar.md
@@ -27,8 +29,8 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
   const {
     image,
     name,
-    onClick = () => undefined,
-    onMouseOver = () => undefined,
+    onClick = noop,
+    onMouseOver = noop,
     shape = 'circle',
     size = 32,
   } = props;
@@ -43,9 +45,11 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
 
   const initials = (name || '').charAt(0);
 
+  const clickableClass = onClick !== noop ? 'str-chat__avatar--clickable' : '';
+
   return (
     <div
-      className={`str-chat__avatar str-chat__avatar--${shape}`}
+      className={`str-chat__avatar str-chat__avatar--${shape} ${clickableClass}`}
       data-testid='avatar'
       onClick={onClick}
       onMouseOver={onMouseOver}
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -1,4 +1,5 @@
 import React, {
+  MouseEvent,
   PropsWithChildren,
   useCallback,
   useEffect,
@@ -25,6 +26,7 @@ import {
   Channel as StreamChannel,
   StreamChat,
   UpdatedMessage,
+  UserResponse,
 } from 'stream-chat';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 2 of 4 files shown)


nhannah (commented):
Adds values to context that are never used within the library. This seems to be a customization for a single use case that is not fully implemented to provide useful functionality out of the box.

wmadden:
@nhannah any response here? My changes here connect unreachable code in your library that we could really use

@DanC5 any response?

nhannah:
@wmadden Your PR adds values to context but never pulls them off of context to apply them to components. There is an ongoing refactor to reduce prop drilling and provide this sort of functionality efficiently (i.e. avoiding unnecessary re-renders due to context changes). But simply adding the values to a context so they can be pulled off again in a custom integration is a partial implementation. You could create a context yourself and get the same result.

wmadden:
@nhannah these props are already pulled off context in [Message.tsx](https://github.com/GetStream/stream-chat-react/blob/master/src/components/Message/Message.tsx#L105-L106) so I didn't need to pull them off in my PR

nhannah:
Only `channel` is coming off of channel context it appears: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/Message/Message.tsx#L220 .

wmadden:
@nhannah unless I'm mistaken, here is the path that the channel context values follow:

1. The complete channel context value is spread as props into `MessageList`: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/MessageList/MessageList.tsx#L451
2. `onUserClick` and `onUserHover` are then passed from `MessageListWithContext` to `MessageListInner` in its `internalMessageProps` prop: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/MessageList/MessageList.tsx#L325-L326
3. `UnmemoizedMessageListInner` spreads its `internalMessagProps` (including `onUserClick` and `onUserHover`) on to `Message`: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/MessageList/MessageListInner.tsx#L390-L398
4. `Message` blindly passes its props down to `MemoizedMessage`: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/Message/Message.tsx#L254-L256
5. `MemoizedMessage` passes `onUserClick` and `onUserHover` (after wrapping them with `useUserHandler()`) down to the final message rendering component: https://github.com/GetStream/stream-chat-react/blob/8b7fd4a7b584387dbd84849c8c148cffe8517cab/src/components/Message/Message.tsx#L176-L177

nhannah:
> There is an ongoing refactor to reduce prop drilling and provide this sort of functionality efficiently (i.e. avoiding unnecessary re-renders due to context changes).

@wmadden This prop spread and drill is already removed in a development branch. We can add `onUserClick` and `onUserHover` to appropriate contexts for both name and avatar during the refactor, @DanC5 is working on the refactor currently. But as it sits this functionality only works due to legacy code that is being removed.

wmadden:
@nhannah I agree all the code in this area is terrible and refactoring it is a good idea. I would like to be able to use the `onUserClick` prop that's present but inaccessible. It's clear you're not going to accept my changes so it'd be great if @DanC5 could complete this feature in his refactor.

In future I'd appreciate it if you took the time to understand my code before rejecting it, since I took the time to try to fix a problem for you.

PR #807, opened 2021-03-23
Add renderText options

DanC5:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/Message/types.ts
+++ b/src/components/Message/types.ts
@@ -28,6 +28,7 @@ import type {
   RetrySendMessage,
   StreamMessage,
 } from '../../context/ChannelContext';
+import type { MentionProps, RenderTextOptions } from '../../utils';
 
 import type {
   CustomTrigger,
@@ -281,6 +282,8 @@ export type MessageUIComponentProps<
   renderText?: (
     text?: string,
     mentioned_users?: UserResponse<Us>[],
+    MentionComponent?: React.ComponentType<MentionProps<Us>>,
+    options?: RenderTextOptions,
   ) => JSX.Element | null;
   /** Whether or not the current Message is in a Thread */
   threadList?: boolean;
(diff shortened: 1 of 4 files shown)


src/components/Message/types.ts
```diff
@@ -281,6 +282,8 @@ export type MessageUIComponentProps<
   renderText?: (
     text?: string,
     mentioned_users?: UserResponse<Us>[],
+    MentionComponent?: React.ComponentType<MentionProps<Us>>,
```
ath92, line 285: I think it might be cleaner to remove this param if we change the API to this, since users can now also override the Mention component by specifying it inside `options.customMarkdownRenderers`.

src/utils.tsx
```diff
@@ -188,7 +196,12 @@ export const renderText = <Us extends DefaultUserType<Us> = DefaultUserType>(
     if (noParsingNeeded.length > 0 || linkIsInBlock) return;
 
     const displayLink =
-      type === 'email' ? value : truncate(value.replace(detectHttp, ''), 20);
+      type === 'email'
```
ath92, line 199: Maybe we could move the truncation logic inside the custom renderer for `link` tags. That way we don't need to have a separate option for `truncationLimit`, but instead users can provide a custom renderer for `link` that truncates in some different way.

src/components/Message/types.ts
DanC5: ahh right, good call on that, will remove

src/utils.tsx
```diff
@@ -74,17 +74,26 @@ type MarkDownRenderers = {
...
+
+    if (!href || (!isEmail && !isUrl)) {
+      return children;
     }
 
     return (
-      <a href={props.href} rel='nofollow noreferrer noopener' target='_blank'>
-        {props.children}
-      </a>
+      <span>
```
ath92, line 87: Don't think we need this span right?

ath92 (approved):

PR #821, opened 2021-03-31
CRS-455 Emoji flag fonts

ath92:
# Submit a pull request

## CLA

Adds two fonts (EmojiOne, sourced from https://github.com/adobe-fonts/emojione-color and NotoColor, sourced from https://github.com/googlefonts/noto-emoji). These fonts are only applied when a user is on Windows, and are only used to override flag emojis. Modern versions of Windows do not display country flags out of the box, but instead show characters (see e.g. https://answers.microsoft.com/en-us/windows/forum/all/flag-emoji/85b163bc-786a-4918-9042-763ccf4b6c05). We use a color SVG font (EmojiOne) for users on Firefox, and a color PNG font (Noto) for users on Chrome and other browsers. 

This PR adds a prop to the `Chat` component called `useImageFlagEmojisOnWindows`, which if set to true will apply the necessary css classes to load the font in the user's browser. While the font is loading, it will fallback to the default system rendering. I'm happy to hear suggestions about the name of this prop, it's not the prettiest at the moment but I struggled to find a clear and concise name.


src/components/MessageList/__tests__/__snapshots__/VirtualizedMessageList.test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`VirtualizedMessageList should render the list without any message 1`] = `
 <div
-  className="str-chat str-chat-channel messaging light"
+  className="str-chat str-chat-channel messaging light "
```
DanC5, line 5: does this extra space need to be there?

src/context/ChatContext.tsx
```diff
@@ -36,6 +36,7 @@ export type ChatContextValue<
     watchers?: { limit?: number; offset?: number },
     event?: React.SyntheticEvent,
   ) => void;
+  useImageFlagEmojisOnWindows?: boolean;
```
DanC5, line 39: since we give this a default value in Chat looks like it doesn't need to be optional

DanC5: might be a good idea to look at all these types actually, some might always exist

DanC5 (approved):

src/context/ChatContext.tsx
ath92: Right, I figured we might be doing that because technically we don't really need to have a value for this to work. I thought maybe the other attributes were also optional for that reason, but I can change it if needed

src/components/MessageList/__tests__/__snapshots__/VirtualizedMessageList.test.js.snap
ath92: Was a little easier to just add the space to the snapshot, but I changed it now so that the space is only added if it's needed to add another className

DanC5 (approved):

PR #835, opened 2021-04-05
Merge MessageList and MessageListInner

petyosi:
Break down the major pieces to separate hooks.


DanC5 (approved):
@petyosi  made some small tweaks and fixed the super fun merge conflicts that arose from the other message list work we did this morning; app is working fine, will merge and then start looking at transferring some of the props to a context

PR #870, opened 2021-04-23
Slow Mode UX

DanC5:
# Submit a pull request

## CLA

after message send...

Screen Shot 2021-04-23 at 1 42 30 PM

after cooldown period...

Screen Shot 2021-04-23 at 1 43 12 PM

### DIFF
--- a/src/components/ChatAutoComplete/ChatAutoComplete.tsx
+++ b/src/components/ChatAutoComplete/ChatAutoComplete.tsx
@@ -150,6 +150,8 @@ export type ChatAutoCompleteProps = {
   placeholder?: string;
   /** The number of rows you want the textarea to have */
   rows?: number;
+  /** Disables the message input during the slow mode cooldown period */
+  slowModeDisabled?: boolean;
 };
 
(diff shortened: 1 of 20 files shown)


ath92 (commented):
In general, looks good. Couple of questions:

 - Would it make sense to apply the `useCooldownTimer` hook inside `useMessageInputState`, and return its values/callbacks from there? That way, its values and callbacks would get added to the `MessageInputContext`, and you'd have to do less prop-passing. It would mean that a few more components might get re-rendered whenever the number in the timer changes, but that doesn't seem like a big issue (they also re-render when the input value is changed while typing because that's also part of the context, and it's fast enough for that).
 - In the designs for the mobile SDK it looks like the countdown timer replaces the send button when it's active.  Would it make sense to move the timer to the right hand side of the input, and hide the send button during cool down?
 - Also, in the designs the timer looks a bit different, i.e. there's a background circle, and the contrast of the text is lower. I know we're only using the designs as a guideline for this sdk, but I think we can make it a little bit more similar out of the box.

ath92 (approved):

PR #900, opened 2021-05-05
Feature/quoted message

DanC5:
# Submit a pull request

## CLA

Screen Shot 2021-05-05 at 3 04 55 PM

Screen Shot 2021-05-05 at 3 05 52 PM


ath92 (commented):
I noticed some styling issues when quoting a message with a scraped URL in it:
image

ath92:
When I try to reply to a message that only has an image in it (a .svg of the React logo), it seems to not render that attachment:
image

ath92:
Not sure how easy it would be, but I'd personally like it if the message input textarea receives focus after I click "reply", so it's faster to start typing

ath92:
image
If I start a thread based on a message that has a quoted message with an attachment, things look a little weird

ath92:
Nit: I noticed that after pressing send, the value of the textarea is cleared, but the quoted message remains until the request has actually gone through (I think). Does it make sense to remove the quoted message a little earlier as well?

yetieaterxb1 (approved):
Grant quoted "LGTM"

PR #915, opened 2021-05-12
fix: Wrong "stream-chat" peer dependency

ardeois:
# Submit a pull request

## CLA

The peerDependency was declaring `stream-chat@3.1.x` which block us installing `stream-chat@3.7.0` when we use npm@7 in strict mode

Using `^3.1.0` should fix the issue as it would allow all `v3` major version that are above `3.1.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   "peerDependencies": {
     "react": "^17.0.0 || ^16.8.0",
     "react-dom": "^17.0.0 || ^16.8.0",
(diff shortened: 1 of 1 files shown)


DanC5:
@ardeois good catch here. I'm making a patch release today so will add this piece in, thanks for the contribution!

ardeois:
Perfect thank you for the fast merge @DanC5 !

PR #927, opened 2021-05-18
Css variables

yetieaterxb1:
# Submit a pull request

## CLA


src/styles/NewChannel.scss
```diff
@@ -1,47 +1,45 @@
 .str-chat__new-channel {
```
DanC5, line 1: doesn't look like the styles in this file are applied anywhere so we should delete

src/styles/Panel.scss
```diff
@@ -3,9 +3,9 @@
   bottom: calc(100%);
```
DanC5, line 3: doesn't look like the styles in this file are applied anywhere so we should delete

src/styles/ReadState.scss
```diff
@@ -5,5 +5,5 @@
   position: absolute;
```
DanC5, line 5: doesn't look like the styles in this file are applied anywhere so we should delete

src/styles/_variables.scss
```diff
@@ -1,26 +1,177 @@
...
+  --accent_blue: #005fff;
+  --accent_green: #20e070;
+  --accent_red: #ff3742;
+
+  /* base colors */
+
+  --bg-gradient-end: #f7f7f7;
+  --bg-gradient-start: #fcfcfc;
+  --black: #000000; //keep
+  --blue-alice: #e9f2ff; //keep
```
DanC5, line 137: why some of these say keep

src/styles/_variables.scss
```diff
@@ -1,26 +1,177 @@
...
+  --xxl-font: 26px;
+  --xxxl-font: 32px;
+
+  /* fonts weight */
+  --font-weight-regular: #{$base-font-weight}; //400
+  --font-weight-semi-bold: 600;
+  --font-weight-bold: #{$heavy-font-weight}; //700
+
+  /* colors */
+  --primary-color: #{$secondary-color};
```
ambaldwin, line 48: primary color is set to secondary-color?

ambaldwin: actually, disregard the above note, and just add a comment to the file here to say why it's being set this way!

ambaldwin (changes requested):

DanC5 (approved):

PR #929, opened 2021-05-19
Ensure new message exists in scroll manager

mikkel-pearl:
# Submit a pull request

## CLA

- [X] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).

I am getting the error `Unhandled Runtime Error TypeError: Cannot read property 'latest_reactions' of undefined` when an empty chat channel is active and I switch to another tab and then back. 

I believe it is because useMessageListScrollManager is being activated by measures and newMessages is empty. I've put in a check but am having trouble with the build so haven't been able to test.


DanC5:
@mikkel-pearl this is a good catch, and I can see where the issue lies as one level up the chain we give `messages` a default value of empty array, which is likely what's getting fed into the hook in your case since the Channel is empty.  Then we attempt to index the last value, which is now undefined, and then grab `latest_reactions`, which of course can't be pulled off undefined.

I was going to make a patch release today anyways, so I'm going to close this PR and roll your change into that release. Look for this to come out within the hour, thanks for the contribution!

PR #1015, opened 2021-06-26
Update index.ts

pyrytakala:
fix typo


DanC5 (approved):
@pyrytakala oh wow, good catch on this typo! Thank you very much for contributing!

PR #1045, opened 2021-07-12
Giphy preview refactor

DanC5:
This PR adds functionality to the virtualized list (via the `separateGiphyPreview` prop) that allows you to break the giphy preview logic out of the main list, so the preview doesn't scroll away in a large channel. Sending a /giphy command renders a component between the input and list. The main logical changes happen in VirtualizedMessageList, useGiphyPreview, and MessageList/utils.tsx.

The final UI (below) is quite simple, just a standard Message component wrapped in a div for styling, with the whole thing overridable via the ComponentContext. Let me know if you have question!

Screen Shot 2021-07-12 at 2 24 31 PM


ambaldwin (changes requested):
Looks really good - just need to add `useGiphyPreview` hook to the Custom Hooks/MessageList Hooks documentation section

petyosi (approved):

PR #1084, opened 2021-07-30
Mentions list

ambaldwin:
# Submit a pull request

Screen Shot 2021-07-29 at 6 43 51 PM

## CLA


examples/website-demos/social-messaging/src/components/ChannelPreview/SocialChannelPreview.tsx
```diff
@@ -28,12 +31,18 @@ export const SocialChannelPreview: React.FC<ChannelPreviewUIComponentProps> = (p
     const members = Object.values(channel.state.members).filter(
       ({ user }) => user?.id !== client.userID,
     );
+
+    const handleUnreadCounts = () => {
+      if (unread && unread > 0) setChatsUnreadCount(chatsUnreadCount - unread);
```
DanC5, line 36: can you just do this in both conditionals? since 0 is falsey
```suggestion
      if (unread) setChatsUnreadCount(chatsUnreadCount - unread);
```

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -58,7 +60,7 @@ export type UserType = { image?: string };
 export const ChatContainer: React.FC = () => {
     const [chatClient, setChatClient] = useState<StreamChat | null>(null);
 
-    const { isSideDrawerOpen } = useViewContext();
+    const { chatsUnreadCount, isListMentions, isNewChat, isSideDrawerOpen, mentionsUnreadCount, setChatsUnreadCount, setMentionsUnreadCount } = useViewContext();
```
DanC5, line 63: what is your print width set to? it should be 100, this seems longer

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -58,7 +60,7 @@ export type UserType = { image?: string };
 export const ChatContainer: React.FC = () => {
     const [chatClient, setChatClient] = useState<StreamChat | null>(null);
```
DanC5, line 61: i think you can just do this...
```suggestion
    const [chatClient, setChatClient] = useState<StreamChat>();
```

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -1,14 +1,16 @@
 import { useEffect, useState } from 'react';
 
-import { ChannelSort, LiteralStringForUnion, StreamChat } from 'stream-chat';
+import type { Channel as TypeChannel } from 'stream-chat';
```
DanC5, line 3: just for consistency, in the other app we've been using the name `StreamChannel` when we need to reassign the type

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -84,23 +86,63 @@ export const ChatContainer: React.FC = () => {
         };
     }, []); // eslint-disable-line
 
+    const customRenderFilter = (channels: TypeChannel[]) => {
+        const getTotalChatUnreadCount =
+            channels?.map(channel => channel.countUnread()).reduce((total, count) => total + count, 0);
```
DanC5, line 91: looks like the function param isn't optional so it will always be truthy and this map will run regardless, even if empty array

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -84,23 +86,63 @@ export const ChatContainer: React.FC = () => {
...
+    }
+
+    useEffect(() => {
+        const handlerNewMessageEvent = (event: Event) => {
+            if (event.user?.id !== chatClient?.userID) {
+                setChatsUnreadCount(chatsUnreadCount + 1)
+
+                const mentions = event.message?.mentioned_users?.filter(user => user.id === chatClient?.userID);
+
+                if (mentions && mentions.length) {
```
DanC5, line 115:
```suggestion
                if (mentions?.length) {
```

examples/website-demos/social-messaging/src/components/EmptyStateIndicator/SocialEmptyStateIndicator.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { EmptyStateIndicatorProps } from 'stream-chat-react';
+
+import './SocialEmptyStateIndicator.scss';
+
+type Props = EmptyStateIndicatorProps;
+
+export const SocialEmptyStateIndicator: React.FC<Props> = (props) => {
```
DanC5, line 9: you can just put EmptyStateIndicatorProps here unless you're planning to extend them

examples/website-demos/social-messaging/src/components/EmptyStateIndicator/SocialEmptyStateIndicator.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { EmptyStateIndicatorProps } from 'stream-chat-react';
```
DanC5, line 3: import type

examples/website-demos/social-messaging/src/components/ChannelPreview/SocialChannelPreview.tsx
ambaldwin: unread could be undefined though, so we need to check to see if it exists.

DanC5: yeah but if undefined it still won't enter into that conditional, it should perform the same either way, right?

ambaldwin: TypeScript doesn't like it

Screen Shot 2021-08-02 at 3 18 53 PM

DanC5: i mean just `if (unread)`, not > 0

ambaldwin: Oh, I misread what you suggested up there initially!
But, I still want to check if it's greater than 0. I was actually logging that number as zero and though the count would be correct (if not checking for greater then and then subtracting 0), I'd rather not run the that state setting calculation. Or, does that matter if you're trying to set state as the same number?

ambaldwin: Ohhhhh, disregard - I see. 0 is falsey. Ok, I will update to if (unread) once I'm back on my computer!

DanC5: haha yep that's what i was getting at, nothing like talking past each other in GH comments lol

examples/website-demos/social-messaging/src/components/ChatContainer/ChatContainer.tsx
```diff
@@ -94,18 +94,19 @@ export const ChatContainer: React.FC = () => {
     };
   }, []); // eslint-disable-line
 
-  const customRenderFilter = (channels: StreamChannel[]) => {
+  const CustomRenderFilter = (channels: StreamChannel[]) => {
```
DanC5, line 97: this should still be lowercase since it's not a component, but otherwise all looking good

DanC5 (approved):

PR #1112, opened 2021-08-11
Virtualized list improvements

petyosi:
- Expose `defaultItemHeight` prop to smooth out long uneven lists
- Fix quick "new messages" blink when loaded or switching channels


DanC5 (approved):
looks good, `defaultItemHeight` needs to be added to the docusaurus props list, but i'll do that right now

PR #1157, opened 2021-09-06
Add MessageInputState.setText

amccloud:
# Submit a pull request

## CLA

- [ ] Code changes are tested (I didn't see a place to add test for this case)

Hello team Stream!

I've spent the weekend trying to simply update the initial text in `MessageInput`. I know that with even just vanilla react this could be difficult as you're sorta in between a controlled and uncontrolled input. My first approach was to work with the textarea ref directly, but things get out of sync quickly

Once I stumbled on `MessageInputeState.insertText`, I was able to get most things working to insert my initial value. The problem is I'd like to re-run the same logic over again if the user chooses another option. As expected with `insertText`, this just accumulates initial values

```js
useEffect(() => {
  if (
    annotation &&
    annotation.start &&
  ) {
    messageInputContext.setText(""); // I'd really like to clear the textarea here, so I can re-run this if annotation.start changes
    messageInputContext.insertText(secondsLabel(annotation.start)); // seconds label e.g. 1:30
  }
}, [annotation]);
```

From what I can tell there is currently no way to clear or update MessageInput text? This PR introduces both capabilities by exposing `setText` from the reducer as `MessageInputState.setText`

- Add MessageInputState.setText


DanC5:
Hello @amccloud, sorry for the delayed response. I've been traveling this week so my schedule has been a bit weird. On first glance this looks like a good feature add. I'll dive into it more deeply early next week then if it makes sense we'll get something pushed in a release. Thanks for the contribution!

DanC5:
@amccloud Thanks for the contribution! I've actually reached for something similar in the past, so it definitely makes sense to add this. I've tweaked the typing slightly and added the new context value to our docs in this PR, https://github.com/GetStream/stream-chat-react/pull/1184.

I'm going to make a release later this morning, so look for v6.8.0, thanks!

PR #1170, opened 2021-09-14
fix mentions list to not include muted users

ambaldwin:
# Submit a pull request

### 🎯 Goal

Fix a bug that actually incorrectly only included muted users as mention options instead of the reverse.

### 🛠 Implementation details

Updated the conditional in the `filterMutes` method.

### 🎨 UI Changes

None

### DIFF
--- a/src/components/MessageInput/hooks/useUserTrigger.ts
+++ b/src/components/MessageInput/hooks/useUserTrigger.ts
@@ -126,7 +126,7 @@ export const useUserTrigger = <
 
       const filterMutes = (data: UserResponse<Us>[]) => {
         if (!mutes.length) return data;
-        return data.filter((suggestion) => mutes.some((mute) => mute.target.id === suggestion.id));
+        return data.filter((suggestion) => mutes.some((mute) => mute.target.id !== suggestion.id));
       };
 
       if (mentionAllAppUsers) {


DanC5 (approved):

src/components/MessageInput/hooks/useUserTrigger.ts
```diff
@@ -126,7 +126,7 @@ export const useUserTrigger = <
 
       const filterMutes = (data: UserResponse<Us>[]) => {
         if (!mutes.length) return data;
-        return data.filter((suggestion) => mutes.some((mute) => mute.target.id === suggestion.id));
+        return data.filter((suggestion) => mutes.some((mute) => mute.target.id !== suggestion.id));
```
JakeLokkesmoe, line 129: I don't think this is quite right either. Shouldn't the `some` here be `every` instead? Otherwise if I have two users with different `id`s muted, this filter will always return true.

```js
return data.filter((suggestion) => mutes.some((mute) => mute.target.id !== suggestion.id));
```

Thanks for your help on this!

ambaldwin: You're right - just added a new PR. Thanks!

PR #1176, opened 2021-09-15
Mentions autocomplete experience

ambaldwin:
# Submit a pull request

### 🎯 Goal

Add better support for mentions by ignoring diacritics, using transliteration (optionally via an optional dependency and by using a new prop, `useMentionsTransliteration` for the `MessageInput`), and Levenshtein distance to match similar names;

### 🛠 Implementation details

Add [Transliterate](https://github.com/sindresorhus/transliterate) library to use the Transliterator.
Add code for Levenstein distance. The minimum distance of 3.
Add Method to remove all diacritics.

### 🎨 UI Changes

None


docusaurus/docs/React/contexts/message-input-context.mdx
```diff
@@ -440,3 +440,11 @@ Function to upload an array of files to the `fileUploads` and `imageUploads` map
 | Type                                |
 | ----------------------------------- |
 | (files: FileList \| File[]) => void |
+
+### useMentionsTransliteration?: boolean;
```
DanC5, line 444: don't need the TS syntax in here

docusaurus/docs/React/message-input-components/message-input.mdx
```diff
@@ -190,3 +190,12 @@ If true, triggers typing events on text input keystroke.
 | Type    | Default |
 | ------- | ------- |
 | boolean | true    |
+
+
+### useMentionsTransliteration?: boolean;
```
DanC5, line 195: don't need the TS syntax in here

src/components/MessageInput/hooks/utils.ts
```diff
@@ -0,0 +1,93 @@
+import type { UserResponse } from 'stream-chat';
+// import transliterate from '@sindresorhus/transliterate';
```
DanC5, line 2: can delete

src/components/MessageInput/hooks/useUserTrigger.ts
```diff
@@ -153,16 +162,12 @@ export const useUserTrigger = <
...
-            return true;
-          }
-
-          return user.id.toLowerCase().includes(query.toLowerCase());
-        });
+        const matchingUsers = searchLocalUsers<Us>(
+          client.userID,
+          users,
+          query,
+          useMentionsTransliteration,
```
DanC5, line 169: instead of four different function args, let's do one object argument

src/components/MessageInput/hooks/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  return matrix[name.length][query.length];
+};
+
+export const searchLocalUsers = <Us extends DefaultUserType<Us> = DefaultUserType>(
+  ownUserId: string | undefined,
+  users: UserResponse<Us>[],
+  query: string,
+  useMentionsTransliteration?: boolean,
```
DanC5, line 62: after changing the function to an object argument, extract params into own type with something like
`(params: SearchLocalUsersParams)`

src/components/MessageInput/hooks/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (user.id === ownUserId) return false;
+    if (!query) return true;
+
+    let updatedId = removeDiacritics(user.id).toLowerCase();
+    let updatedName = removeDiacritics(user.name).toLowerCase();
+    let updatedQuery = removeDiacritics(query).toLowerCase();
+
+    if (useMentionsTransliteration) {
+      (async () => {
+        const { default: theDefault } = await import('@sindresorhus/transliterate');
```
DanC5, line 74: i don't understand what `theDefault` means, can we rename that to something more semantic?

DanC5: maybe even just `transliterate`

src/components/MessageInput/hooks/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+        updatedName = theDefault(user.name || '').toLowerCase();
+        updatedQuery = theDefault(query).toLowerCase();
+        updatedId = theDefault(user.id).toLowerCase();
+      })();
+    }
+
+    if (updatedName !== undefined && updatedName.length) {
```
DanC5, line 81: can this just be `if (updatedName?.length)`

src/components/MessageInput/hooks/utils.ts
ambaldwin: default is a reserved word, so it has to be renamed. I can see how theDefault doesn't work :)

src/components/MessageInput/hooks/utils.ts
ambaldwin: updatedName could also be an empty string, which is not undefined, yet we still want want to enter this block if that' the case.

ambaldwin: this would want to skip this block and use the Id for searching, which is below this code

DanC5: yeah an empty string is still falesy so it wouldn't enter this block if that was the case

src/components/MessageInput/hooks/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (useMentionsTransliteration) {
+      (async () => {
+        const { default: transliterate } = await import('@sindresorhus/transliterate');
+        updatedName = transliterate(user.name || '').toLowerCase();
+        updatedQuery = transliterate(query).toLowerCase();
+        updatedId = transliterate(user.id).toLowerCase();
+      })();
+    }
+
+    if (updatedName !== undefined) {
```
DanC5, line 87: actually can't this now just be `if (updatedName)` ?

ambaldwin: yes! I just changed this. An empty string is actually nullable, so this check didn't work with just updatedName !== undefined, which is why I added the length condition. But, if (updatedName) is the same.

DanC5 (approved):

PR #1193, opened 2021-09-21
Social message input

zacheryconverse:
# Submit a pull request

### 🎯 Goal

Attachment previews for files and images. Adjust emoji picker. Bolt onClick sets giphyState.

### 🛠 Implementation details

Attach button displays file picker via rfu FileUplaodButton and ImageDropzone. Add bar to emoji mart. stopPropogation onClick

### 🎨 UI Changes

Screen Shot 2021-09-21 at 10 20 58 AM

Screen Shot 2021-09-21 at 10 24 40 AM


examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.scss
```diff
@@ -27,19 +28,26 @@
...
     &-attach {
       display: flex;
-      padding: 8px 13px;
+      padding: 0 13px;
     }
 
     &-bolt {
       display: flex;
-      padding: 6px;
+      padding: 9px 6px;
```
DanC5, line 36: 9 seems like an odd choice, is that what the designs say?

examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.scss
```diff
@@ -27,19 +28,26 @@
 
     &-attach {
       display: flex;
-      padding: 8px 13px;
+      padding: 0 13px;
```
DanC5, line 31: 13? this what designs have?

zacheryconverse: 13 is what the design says - yes

examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.tsx
```diff
@@ -115,27 +115,29 @@ export const SocialMessageInput = (props: Props) => {
...
-                    className='input-ui-input-emoji-picker'
-                    ref={emojiPickerRef}
-                    onClick={openEmojiPicker}
-                  >
-                    <EmojiPickerIcon />
-                  </div>
-                )}
-              </>
-            }
+            <div className='input-ui-wrapper-emoji'>
```
DanC5, line 121: this classname doesn't make a ton of sense, in fact i think we can probably clean up the naming higher up the tree as well. for instance we have `input-ui-container` at the outermost level and then `input-ui-wrapper` nested on level within. outermost should probably be `input-ui` and then we append semantic naming related to their function after, so the first nested div would be something like `input-ui-icons`, which makes more sense than wrapper inside container

examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.tsx
```diff
@@ -87,22 +87,22 @@ export const SocialMessageInput = (props: Props) => {
...
+  const handleCommandsClick = (event: React.MouseEvent) => {
+    event.stopPropagation();
     openCommandsList();
     setGiphyState(false);
     setCommandsOpen(true);
   };
-
   return (
     <>
       <div className='input-ui-container'>
```
DanC5, line 98: let's get rid of container entirely as ideally we build upon the class name from parent to child, and in this case container isn't used/appended again

examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.tsx
```diff
@@ -115,27 +115,29 @@ export const SocialMessageInput = (props: Props) => {
...
-                    className='input-ui-input-emoji-picker'
-                    ref={emojiPickerRef}
-                    onClick={openEmojiPicker}
-                  >
-                    <EmojiPickerIcon />
-                  </div>
-                )}
-              </>
-            }
+            <div className='input-ui-icons-emoji'>
```
DanC5, line 121: this can also be changed, if we're building on the previous classname something like `input-ui-input-textarea` could work, do whatever you think makes the most sense

DanC5 (approved):
🥩

PR #1206, opened 2021-09-27
prevent user watch events from updating channel state

DanC5:
# Submit a pull request

### 🎯 Goal

Prevent the message list from re-rendering when users join/leave a channel

### 🛠 Implementation details

Do not initiate an action to update the state returned from the channel reducer when user watch events are received

### 🎨 UI Changes

None

### DIFF
--- a/examples/typescript/yarn.lock
+++ b/examples/typescript/yarn.lock
@@ -10063,15 +10063,6 @@ react-view-pager@^0.6.0:
     resize-observer-polyfill "1.5.0"
     tabbable "1.1.2"
 
-react-virtuoso@^1.10.8:
-  version "1.11.1"
-  resolved "https://registry.yarnpkg.com/react-virtuoso/-/react-virtuoso-1.11.1.tgz#350dc88866b3f42cb396bf8f40b90efc6daaadde"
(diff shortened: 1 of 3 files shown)


zacheryconverse (approved):

examples/typescript/yarn.lock
```diff
@@ -10082,6 +10073,14 @@ react-virtuoso@^1.5.8:
     react-app-polyfill "^1.0.6"
     resize-observer-polyfill "^1.5.1"
 
+react-virtuoso@^2.2.0:
```
petyosi, line 10076: having two react-virtuosos in yarn.lock does not seem right

DanC5: oh interesting, that second one might be coming from the MML library, let me check...

DanC5: yep it's coming from `mml-react` dependency

DanC5: should we make it a peer dep in that repo?

petyosi: I would rather upgrade it TBH

DanC5: ok i'll upgrade it in `mml-react`

PR #1222, opened 2021-10-04
Social slow mode

ambaldwin:
# Submit a pull request

### 🎯 Goal

Custom SlowMode UI component and functionality for Social App Demo.

### 🛠 Implementation details

Custom SlowMode UI component and CSS.

### 🎨 UI Changes

Screen Shot 2021-10-04 at 11 57 02 AM


examples/website-demos/social-messaging/src/components/CooldownTimer/SocialCooldownTimer.tsx
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from 'react';
```
DanC5, line 1: did you need to create your own component here vs using the one from the library? i ask because we used the library version on the virtual event app

examples/website-demos/social-messaging/src/components/MessageInput/SocialMessageInput.tsx
```diff
@@ -144,13 +163,7 @@ export const SocialMessageInput = (props: Props) => {
...
           onClick={handleSubmit}
         >
-          {giphyState && !numberOfUploads ? (
-            <GiphySearch />
-          ) : (
-            <>
-              <SendArrow />
-            </>
-          )}
+          {giphyState && !numberOfUploads ? <GiphySearch /> : <>{renderSendButton()}</>}
```
DanC5, line 166: instead of invoking the function in the jsx we can do the conditional outside the return so we create a constant reference to the component, so something like

```
const SendButton = cooldownRemaining ? (
      <SocialCooldownTimer
        cooldownInterval={cooldownInterval}
        setCooldownRemaining={setCooldownRemaining}
      />
    ) : (
      <SendArrow />
    );
  };
```

DanC5: and i guess you could even toss it in a useMemo with cooldownRemaining as the dependency so it doesn't recompute every other stuff in the component changes, like text for example

ambaldwin: good call - updated.

DanC5 (approved):

PR #1232, opened 2021-10-07
use channel capabilities for permissions

DanC5:
# Submit a pull request

### 🎯 Goal

use new `own_capabilities` field on channel data to determine user permissions

### 🛠 Implementation details

add `channelCapabilities` to `ChannelStateContext` to evaluate its values where appropriate

### 🎨 UI Changes

none


linhewinner:
I filed an issue that may be related to this PR:
https://github.com/GetStream/stream-chat-react/issues/1323

I wonder if there has been some Stream backend change recently.
"channel.updated" event no longer has "own_capabilities" field.

PR #1284, opened 2021-11-09
Also allow stream-chat 4.0.0 and up

vdhpieter:
# Submit a pull request

### 🎯 Goal

Allow stream-chat 4.0.0 to be used with this library.

### 🛠 Implementation details

I saw the newest api being used in examples so I assumed it should work. Not keeping peer dependencies up to date gives errors from npm v7 and up.


petyosi:
thank you @vdhpieter - we were about to do the same during our next sprint.

vdhpieter:
Fixed by https://github.com/GetStream/stream-chat-react/pull/1305

PR #1287, opened 2021-11-17
Adding custom group styles function

rcdvl:
# Submit a pull request

### 🎯 Goal

Adding a way to change the group styles makes it easier to customize its experience (like grouping only messages within a timeframe, etc).

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_

### DIFF
--- a/src/components/MessageList/MessageList.tsx
+++ b/src/components/MessageList/MessageList.tsx
@@ -25,6 +25,8 @@ import { LoadingIndicator as DefaultLoadingIndicator } from '../Loading';
 import { defaultPinPermissions, MESSAGE_ACTIONS } from '../Message/utils';
 import { TypingIndicator as DefaultTypingIndicator } from '../TypingIndicator';
 
+import type { GroupStyle } from './utils';
+
 import type { MessageProps } from '../Message/types';
 
 import type { StreamMessage } from '../../context/ChannelStateContext';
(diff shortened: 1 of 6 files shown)


src/components/MessageList/hooks/useEnrichedMessages.ts
```diff
@@ -72,10 +79,11 @@ export const useEnrichedMessages = <
     messagesWithDates = insertIntro(messagesWithDates, headerPosition);
   }
 
+  const groupStylesFn = groupStyles ? groupStyles : getGroupStyles;
```
ambaldwin, line 82: Change this to: `const groupStylesFn = groupStyles || getGroupStyles;` to match other overrides in the library.

src/components/MessageList/MessageList.tsx
```diff
@@ -38,6 +38,7 @@ import type {
   DefaultReactionType,
   DefaultUserType,
 } from '../../types/types';
+import type { GroupStyle } from './utils';
```
ambaldwin, line 41: Let's move this import to line 28, above the MessageProp type and with a line in between.

ambaldwin (commented):
Thanks for the addition! Please see very minor comments before we approve. We also need to export * from './utils' in MessageList/index.ts so folks can correctly type the return value of the callback function.

rcdvl (commented):
@ambaldwin All comments addressed (I think 😄), feel free to make any additional suggestion.

petyosi (approved):

PR #1300, opened 2021-12-03
docs: Fix broken links in the docusaurus docs

madsroskar:
# Submit a pull request

### 🎯 Goal

Docusaurus complained about some broken links, this should fix it.

### 🛠 Implementation details

For the relevant ones, link to the mdx file as is done (seemingly) everywhere else to not confuse the resolver in docusaurus. Linking to files also ensures that the CLI will give information about the link if it can't be resolved in the future.

A couple links seemed to have missed an update, or something.

Tested the changes out locally to make sure they worked fine after the changes.

### 🎨 UI Changes

No UI changes (except less 404 :D)

*NOTE*: I had to skip the hooks when making the commit as the python script checking for large files broke cause of "bogus GIT_CONFIG_PARAMETERS", but that should not matter in this case.

### DIFF
--- a/examples/website-demos/social-messaging/src/components/ChannelPreview/SocialPreviewActions.tsx
+++ b/examples/website-demos/social-messaging/src/components/ChannelPreview/SocialPreviewActions.tsx
@@ -25,7 +25,7 @@ export const SocialPreviewActions = (props: Props<SocialUserType>) => {
   const [isUserMuted, setIsUserMuted] = useState(false);
 
   const handleAction = (action: UserActions) => {
-    const user = members[0].user?.id;    
+    const user = members[0].user?.id;
     if (user) setActionId(user);
     setActionsModalOpenId(channelId!);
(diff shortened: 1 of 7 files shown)


examples/website-demos/social-messaging/src/components/ChannelPreview/SocialPreviewActions.tsx
```diff
@@ -25,7 +25,7 @@ export const SocialPreviewActions = (props: Props<SocialUserType>) => {
   const [isUserMuted, setIsUserMuted] = useState(false);
 
   const handleAction = (action: UserActions) => {
-    const user = members[0].user?.id;    
+    const user = members[0].user?.id;
```
madsroskar, line 28: This change was done by `yarn lint-fix`

zacheryconverse (approved):
Nice catch - thanks!

PR #1311, opened 2021-12-14
chore: add component context avatar to quoted message preview

ambaldwin:
# Submit a pull request

### 🎯 Goal

To keep avatars consistent across all components.

### 🛠 Implementation details

Use the avatar in the ComponentContext in the QuotedMessagePreview component if available.


petyosi:
I think this should not be a `chore:` it won't trigger a new release. More like `fix:`.

ambaldwin:
> I think this should not be a `chore:` it won't trigger a new release. More like `fix:`.

Ok, I can squash and merge with the fix prefix

petyosi (approved):

PR #1315, opened 2021-12-20
Next import updates

petyosi:
The PR updates to our scoped packages that work with NextJS node compilation. I can no longer reproduce an emoji-related error I had, which is suspicious. 

@ambaldwin can you test on your side?


ambaldwin:
@petyosi all seems to work on my side - emojis, attachments etc.

PR #1319, opened 2022-01-05
Keyboard

zacheryconverse:
# Submit a pull request

### 🎯 Goal

Allow non mouse users to navigate and use screen readers


petyosi (changes requested):
I did a quick test on my side, and the keyboard navigation does not work for the message list. For example, when I tab to the channel list, I can use `up/down` arrows to navigate the items. This does not seem to work for the Keyboard. 

Also, I think it will be very beneficial to have a documentation article that describes the keyboard behavior we support in detail.

src/components/Channel/Channel.tsx
```diff
@@ -475,6 +475,17 @@ const ChannelInner = <
     }
   }, [state.messages, state.thread]);
 
+  // Focuses last message instead of first
+  useEffect(() => {
+    if (state.messages?.length && !state.thread) {
+      if (document.activeElement?.previousElementSibling) {
+        const elements = document.getElementsByClassName('str-chat__message--regular');
```
yetieaterxb1, line 482: Should this class be dynamic for customized use cases?

ambaldwin (commented):
Looks good. Just needs focusedChannel and loadedChannels added to the ChannelPreview mdx documentation file and accessibility added to documentation too.

src/components/Channel/Channel.tsx
```diff
@@ -554,6 +609,7 @@ const ChannelInner = <
 
     const hasMoreMessages = queryResponse.messages.length === perPage;
     loadMoreFinished(hasMoreMessages, channel.state.messages);
+    setFocusedMessage(messageElements.length - 25);
```
petyosi, line 622: What's the hard-coded 25?

Jarenbeaver:
update modal close button and test

Jarenbeaver:
https://codecov.io/gh/GetStream/stream-chat-react/pull/1319?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=GetStream

Jarenbeaver (approved):

src/components/Channel/Channel.tsx
Jarenbeaver: What's the hard-coded 25?

examples/typescript/yarn.lock
```diff
@@ -9370,10 +9370,17 @@ react-fast-compare@^3.0.1, react-fast-compare@^3.2.0:
   resolved "https://registry.yarnpkg.com/react-fast-compare/-/react-fast-compare-3.2.0.tgz#641a9da81b6a6320f270e89724fb45a0b39e43bb"
   integrity sha512-rtGImPZ0YyLrscKI9xTpV8psd6I8VAtjKCzQDlzyDvqJA8XOW78TXYQwNRNd8g8JZnDu8q9Fu/1v4HPAVwVdHA==
 
+<<<<<<< HEAD
+react-file-utils@1.1.10:
+  version "1.1.10"
+  resolved "https://registry.yarnpkg.com/react-file-utils/-/react-file-utils-1.1.10.tgz#ce42839988b03ef8212901f094999cea8b518fb0"
+  integrity sha512-9zWHA7ttkU6issljXG0WvOTpdCW8qHuwOH2N6OB+lCx5j3slub/aIr2oJqanWOvA0VjBoyFnC7PJ+hRUJ39kRA==
+=======
```
petyosi, line 9378: @zacheryconverse you have a diff marker here

examples/typescript/yarn.lock
```diff
@@ -9370,10 +9370,17 @@ react-fast-compare@^3.0.1, react-fast-compare@^3.2.0:
...
 react-file-utils@1.1.9:
   version "1.1.9"
   resolved "https://registry.yarnpkg.com/react-file-utils/-/react-file-utils-1.1.9.tgz#28811b98323f56be3982ad5d4f83959295db8082"
   integrity sha512-w9zDaD6jWOWzp8au+Tc5yJabQ3FLj2VP0AE+7cB0V0ET1AafcP/CP7NbHpLYoOojM82uMGyzhPhuP4cCuOyVrA==
+>>>>>>> develop
```
petyosi, line 9383: diff marker

src/components/Channel/Channel.tsx
```diff
@@ -475,6 +476,192 @@ const ChannelInner = <
     }
   }, [state.messages, state.thread]);
 
+  /** Keyboard Navigation */
+
+  const regularMessages = channel.state.messages.filter((m) => m.type === 'regular');
+  const numberOfRegularMessages = regularMessages.length;
+  const messageElements = document.getElementsByClassName('str-chat__message--regular');
```
petyosi, line 483: Relying on `document.getElementsByClassName` to reference DOM elements is quite the brittle approach for two reasons. 

1. It searches the global DOM. If for some reason, the document contains more than one message list, it will get them.
2. It also couples the implementation logic to styling classes. If those classes are changed for some reason, this code will stop working. 

The best way to avoid that is to use `refs`. If you need to iterate through multiple elements, then you can still query the DOM, but make that through `qerySelectorAll` of a **known parent** (which you reference with a `ref`) and use a data attribute for that (e.g. `data-role`).

src/components/Channel/Channel.tsx
petyosi: @zacheryconverse

src/components/Thread/Thread.tsx
```diff
@@ -221,12 +221,86 @@ const ThreadInner = <
     }
   }, [threadMessages?.length]);
 
+  /** Keyboard Navigation */
```
petyosi, line 224: This code block seems quite similar to the change in `Channel`. Is there a way for this to be extracted in a separate piece (hook?)

PR #1330, opened 2022-01-14
Set ImagePreviewer to disabled if multiple upload is not allowed

davidavz:
# Submit a pull request

### 🎯 Goal

The `ThumbnailPlaceholder` component should not be visible if multiple upload is disabled

### 🛠 Implementation details

We have to set the `disabled` prop of `ImagePreviewer` component to false if multiple upload is disabled


petyosi (approved):

petyosi:
Thanks, this makes sense.

Issue #1337, opened 2022-01-17
firstOfGroup not working on first message


ivanros:
The `VirtualizedMessageList` prop `firstOfGroup` is not working with the first message in a chat as it doesn't have any message previous to itself.

https://github.com/GetStream/stream-chat-react/blob/9f7be762a36b765c6afcef1fe749cbcf4f620bf5/src/components/MessageList/VirtualizedMessageList.tsx#L239-L242

Maybe if there is not previous message it should be `true`.

zacheryconverse:
Thanks for reaching out about this issue. We have a fix for this and you can expect to see it in our next release. I'll go ahead and close this issue, but feel free to contact us anytime. Have a beautiful day.

Cheers

PR #1338, opened 2022-01-18
firstOfGroup prop is truthy if first message

zacheryconverse:
# Submit a pull request

### 🎯 Goal

Fix bug - prop was returning false if first message of channel


zacheryconverse:
Link to issue https://github.com/GetStream/stream-chat-react/issues/1337

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -238,7 +238,7 @@ const VirtualizedMessageListWithContext = <
 
       const firstOfGroup =
         shouldGroupByUser &&
-        streamMessageIndex > 0 &&
+        message.user?.id === messageList[streamMessageIndex + 1]?.user?.id &&
```
yetieaterxb1, line 241: This will cause a couple of css bugs just so you are aware

ivanros: Hi guys! Thank you for reviewing this bug I've reported, you are doing a great job about it.

Just to clarify the current issue, I'm having this problem:
image

As you can see the first message in a chat is not showing as expected (No avatar, no user title,...). And that's because the `firstOfGroup` there is returning false, because it doesn't have a previous message.

I don't know if this solution will work, as you are checking that the next message exists too. So if a user puts a single message then `firstOfGroup` will be `false` as it doesn't have any messages next to it. For example, in the screenshot the last message will be not showing either the Avatar or the user title.

zacheryconverse: Hi Ivan,

I've updated the PR to remove the index check to fix the first message in a channel. We don't need to check for a previous message because the [endOfGroup](https://github.com/GetStream/stream-chat-css/blob/bde73322aa0a8023be82b4e730b8d128161b4451/src/styles/VirtualMessage.scss#L62) will add a second class the messages like the last message above, which will include attributes like the avatar.

yetieaterxb1 (approved):

PR #1347, opened 2022-01-25
Add wordReplace prop to ChatAutoComplete to override default behavior

zencephalon:
### 🎯 Goal

Make it possible to override the default wordReplace functionality. Our product includes a number of custom emotes that don't have `.native` values that we'd like to handle in a different way using `ChatAutoComplete`, but currently we don't have access to change the behavior of the autocomplete selection.

### 🛠 Implementation details

Adds an optional prop that takes a function to apply custom autocomplete behavior.

### 🎨 UI Changes

No UI changes, change is backwards compatible and passes the existing tests.


petyosi:
@zencephalon thank you for your contribution! We are in the middle of preparing a major release, but we will try to release this sooner rather than later.

PR #1352, opened 2022-01-26
fixes #1349 Issue - display thread send button if upload preview

### LINKED ISSUE #1349: UploadsPreview for MessageInputSmall displays the send button of MessageInputFlat
`MessageInputFlat` and `MessageInputSmall` use the same `UploadsPreview` component. `UploadsPreview`, when deciding if it should show the `str-chat__send-button`, does a `getElementsByClassName` and uses the first element returned. Which is always the send button of the `MessageInputFlat` component.

So if in a thread you try to upload a file the send button will appear in the main chat's input and not in the thread's input

zacheryconverse:
# Submit a pull request

### 🎯 Goal

Show send button in thread input if there is an upload preview

### 🛠 Implementation details

Check for second str-chat__send-button


petyosi (changes requested):
I would rather move away from using `getElementsByClassName`. This approach queries the global scope and relies on presentational classes for logic. This needs to be moved to `ref`s.

ambaldwin:
> I would rather move away from using `getElementsByClassName`. This approach queries the global scope and relies on presentational classes for logic. This needs to be moved to `ref`s.

Updated.

petyosi (changes requested):
The approach taken is better but can be improved. It introduces an additional context dependency (and a re-render) of the button. This can be avoided with a parent CSS selector like I originally suggested. The class can be added to the two message input component element wrappers: ([Flat](https://github.com/GetStream/stream-chat-react/blob/664faf96789fcfa1c4b4438a90a889c753e1e4f3/src/components/MessageInput/MessageInputFlat.tsx#L74-L78) and [Small](https://github.com/GetStream/stream-chat-react/blob/664faf96789fcfa1c4b4438a90a889c753e1e4f3/src/components/MessageInput/MessageInputSmall.tsx#L85-L89)). Those components already use the context containing `numberOfUploads`. 

The CSS class name should be something semantically correct. I can suggest something like `upload previews visible`. It should also follow the BEM conventions which our CSS follows so far.

src/components/MessageInput/__tests__/MessageInput.test.js
```diff
@@ -885,3 +885,104 @@ const ActiveChannelSetter = ({ activeChannel }) => {
...
+              getAsFile: () => image,
+              kind: 'file',
+            },
+          ],
+        };
+        const formElement = await findByPlaceholderText(inputPlaceholder);
+        formElement.dispatchEvent(clipboardEvent);
+        const sendButton = await findByTestId(sendButtonTestId);
+        await waitFor(() => {
+          expect(sendButton).toBeInTheDocument();
```
petyosi, line 968: I am fairly certain that `toBeInTheDocument` won't detect if the button is hidden/visible due to CSS rules. 

In other words, those tests will pass with or without the code change introduced in this PR. Let me know if I am missing something.

zacheryconverse: Instead of testing for css rules, I've changed the tests to confirm if `str-chat__input-flat-has-attachments` is present or not.

petyosi (approved):

PR #1354, opened 2022-01-28
Remove UNSAFE warning from textarea

petyosi:
The regexp calculated value was not used anywhere. 

Please test if mentions, emojis, and slash commands work as expected.


ambaldwin (approved):
I pulled locally and did a lot of testing - looks good to me. it's nice to have have those warnings, too!

PR #1356, opened 2022-01-31
feat: Quote reply to thread messages

szuperaz:
# Submit a pull request

### 🎯 Goal

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_

Stream-chat-css changes: https://github.com/GetStream/stream-chat-css/pull/37

### DIFF
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -582,7 +582,7 @@ const ChannelInner = <
       id,
       mentioned_users: mentions,
       parent_id,
-      quoted_message_id: quotedMessage?.id,
+      quoted_message_id: parent_id === quotedMessage?.parent_id ? quotedMessage?.id : undefined,
       text,
       ...customMessageData,
     } as Message<At, Me, Us>;
@@ -604,7 +604,7 @@ const ChannelInner = <
         });
       }
 
-      if (quotedMessage) setQuotedMessage(undefined);
+      if (quotedMessage && parent_id === quotedMessage?.parent_id) setQuotedMessage(undefined);
     } catch (error) {
       // error response isn't usable so needs to be stringified then parsed
       const stringError = JSON.stringify(error);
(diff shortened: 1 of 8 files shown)


src/components/Channel/channelState.ts
```diff
@@ -178,6 +178,7 @@ export const channelReducer = <
       const { channel, message } = action;
       return {
         ...state,
+        quotedMessage: undefined,
```
szuperaz, line 181: I'd like to deselect quoted message if a thread is opened, but that didn't work with this code

szuperaz: also tried `setQuotedMessage` in `Channel.tsx` but no luck with that

ambaldwin: Hi Zita - we shouldn't need to reset this value within the reducer because the handle function to 'Reply' only sets that local state variable using [setQuotedMessage](https://github.com/GetStream/stream-chat-react/blob/develop/src/components/MessageActions/MessageActionsBox.tsx#L157). I was able to get everything to correctly update by adding `setQuotedMessage(undefined)` within the `openThread` method in `Channel.tsx`. This correctly updates the `ChannelStateContextValue`. Let me know if you're seeing differently or if I misunderstood your intent.

szuperaz: Thanks, I moved the logic to the `Channel.tsx`, but that prompted another question (see my other comment)

src/components/Channel/Channel.tsx
```diff
@@ -687,6 +687,13 @@ const ChannelInner = <
     event: React.BaseSyntheticEvent,
   ) => {
     event.preventDefault();
+    // This is the desired behavior I want to achieve
```
szuperaz, line 690: I don't understand why quoted message is undefined here, despite having a message selected for quote reply, can you also help with this?

ambaldwin: Hmm, I see this - I did some poking around, and I'm not sure either. Maybe @petyosi can assist?

szuperaz:
@petyosi was able to fix the issue, @ambaldwin I also updated @stream-io/stream-chat-css version, let me know if I should also update the version somewhere else in the code

PR #1362, opened 2022-02-07
fix: replace font loading from google by the default system stack

MartinCupela:
### 🎯 Goal

Avoid performing HTTP requests to Google API to load fonts and use the system defaults instead. Developers should have possibility to choose their own fonts without performing redundant HTTP calls.


petyosi (approved):

MartinCupela:
Changes targeted the styles/ folder that has been moved to stream-chat-css.

PR #1374, opened 2022-02-15
Support choosing a giphy version

petyosi:
# Submit a pull request

### 🎯 Goal

The PR exposes a new property that lets the developer choose [which giphy size is rendered](https://developers.giphy.com/docs/api/schema#image-object). It also sets the image dimensions explicitly, preventing the message list from jumping on image load.

### 🛠 Implementation details

The giphy attachment payload includes several versions. By default, we use the `original` one. The `Channel` prop can change the version name.

### 🎨 UI Changes
Here's an image of a smaller giphy:

image


ambaldwin:
Looks good. docusaurus/core-components/channel.mdx needs to be updated with the new `giphyVersion` prop and docusaurus/context/channel-state-context.mdx file needs updated with the new context value of the same name.

petyosi:
@ambaldwin please check the docs I created.

ambaldwin (approved):

PR #1380, opened 2022-02-18
Fixes issue 1376

zacheryconverse:
# Submit a pull request

### 🎯 Goal

Bug fix: Do not render message actions option if possible thread actions are disabled

### 🛠 Implementation details

If all permissions except `pin` were disabled, `showMessageActionsBox` would return a false positive because MessageActionsBox does not allow pins in 

### 🎨 UI Changes

Screen Shot 2022-02-18 at 9 58 13 AM
thread.


petyosi (changes requested):
If I understand the three `ifs` correctly, the behavior can be handled by code that looks like this:

```tsx
const ACTIONS_NOT_WORKING_IN_THREAD = ['pin', 'react', 'reply']; // why reply?

export const showMessageActionsBox = (actions: MessageActionsArray, inThread?: boolean) => {
  if (actions.length === 0) {
    return false;
  }

  if (
    inThread &&
    actions.filter((action) => !ACTIONS_NOT_WORKING_IN_THREAD.includes(action)).length === 0
  ) {
    return false;
  }
//...
```

zacheryconverse:
Much better, yes this is correct. The reason for reply is because it refers to 'reply in thread' (not quote) - so of course you don't want to start a thread from in a thread.

petyosi (approved):

PR #1390, opened 2022-02-23
update i18Next

zacheryconverse:
Bump i18Next from v19 to v21. Allows the use of JSON format v4 plural handling


zacheryconverse:
@petyosi The way we handle replies is now updated.

src/components/Message/MessageRepliesCountButton.tsx
```diff
@@ -25,17 +25,15 @@ const UnMemoizedMessageRepliesCountButton: React.FC<MessageRepliesCountButtonPro
...
   }
 
   if (reply_count && reply_count > 1) {
     if (labelPlural) {
       pluralReplyText = `${reply_count} ${labelPlural}`;
     } else {
-      pluralReplyText = t('{{ replyCount }} replies', {
-        replyCount: reply_count,
-      });
+      pluralReplyText = t('replyCount_other', { replyCount: reply_count });
```
petyosi, line 36: Based on what [I am seeing here](https://www.i18next.com/translation-function/plurals#singular-plural), you're not supposed to use `key_one` or `key_other` as a key. Instead, you should use just `key` and let the framework pull the right key based on the json provided and the count available.

petyosi (changes requested):
If I understand [the samples from the code](https://www.i18next.com/translation-function/plurals#singular-plural), We should not use `key_one` and `key_other` in the source code.

examples/capacitor/yarn.lock
```diff
@@ -9799,6 +9814,14 @@ react-virtuoso@^2.2.1, react-virtuoso@^2.8.2:
     "@virtuoso.dev/react-urx" "^0.2.12"
     "@virtuoso.dev/urx" "^0.2.12"
 
+react-virtuoso@^2.3.1:
```
petyosi, line 9817: Can we remove the example yarn bumps? Not like any harm, but it pollutes the PR?

src/components/Message/MessageRepliesCountButton.tsx
```diff
@@ -18,41 +18,26 @@ const UnMemoizedMessageRepliesCountButton: React.FC<MessageRepliesCountButtonPro
...
-    return (
-      <button
-        className='str-chat__message-replies-count-button'
-        data-testid='replies-count-button'
-        onClick={onClick}
-      >
-        <ReplyIcon />
-        {reply_count === 1 ? singleReplyText : pluralReplyText}
-      </button>
-    );
```
petyosi, line 52: Lovely change.

petyosi (approved):

arnautov-anton (approved):

examples/capacitor/yarn.lock
MartinCupela: Removed via git checkout

PR #1402, opened 2022-02-28
chore: update to channel search and message ui component docs

ambaldwin:
# Submit a pull request

### 🎯 Goal

Add the missing CSS that is referenced in the Custom Message UI example. Add information for the `searchFunction` prop for the `ChannelSearch` customization example as this comes up in support often.

Also, bumped the typescript example app's stream-chat-css version to latest.


petyosi (approved):
@ambaldwin this looks good. Small note: conventional commit format has the `docs:` label which is more suitable for such type of work.

Issue #1403, opened 2022-03-01
Make it easier to append additional autocomplete triggers while keeping the default ones


calvinwyoung:
In order to customize `autocompleteTriggers`, the Stream docs advise us to implement a custom `TriggerProvider`:
https://getstream.io/chat/docs/sdk/react/custom-code-examples/suggestion_list/

Implementing a custom `TriggerProvider` requires, at a minimum, supporting the `/`, `:`, and `@` triggers, whose implementations are provided by `useCommandTrigger`, `useEmojiTrigger`, and `useUserTrigger`.

Unfortunately those hooks are *not* currently exported publicly:
https://github.com/GetStream/stream-chat-react/blob/develop/src/components/MessageInput/hooks/index.ts

As a result, we're forced to copy/paste the implementation of those triggers into our own codebase. We'd prefer to simply import the hooks directly from `stream-chat-react`.

Is there a reason to keep those hooks private to this package? If not, can you please export them? I'd happily open a PR if this is okay with your team.

Thank you!

calvinwyoung:
I went ahead and opened #1406 to address this issue. Thank you for your consideration!

ambaldwin:
Hi @calvinwyoung I believe you're thinking that after creating a custom `TriggerProvider`, [you'd then pass it in as a prop into `Channel`](https://getstream.io/chat/docs/sdk/react/customization/input_ui/#custom-triggers)? Just want to make sure you've seen that documentation page, too. Also, if you're implementing a custom trigger provider, you have the option to remove or add triggers as you wish. At a minimum, you don't necessarily have to use the 3 default triggers.

And, thanks for submitting that PR. I'm not sure of the history of not exposing those hooks, though I know that has come up before in the context of a custom trigger provider, I'm just not sure of the outcome. We will take a look and will update this issue accordingly.

calvinwyoung:
@ambaldwin Yes, I'm thinking about creating a custom `TriggerProvider` that I'd then pass into a `Channel`.

The implementation of `DefaultTriggerProvider` involves initializing a `TriggerSettings` object that maps a string character to one of the trigger hooks:
https://github.com/GetStream/stream-chat-react/blob/0651fe693ef8752bf5453b05d7fb9e033cf17d01/src/components/MessageInput/DefaultTriggerProvider.tsx#L77-L87

`TriggerSettings` is implemented as a union type that requires the `/`, `:`, and `@` triggers to be defined:
[...]

PR #1406, opened 2022-03-01
Export more MessageInput hooks

calvinwyoung:
# Submit a pull request

### 🎯 Goal

This PR exports more hooks from `components/MessageInput/hooks` to simplify the implementation of custom components. More information is provided in this issue: https://github.com/GetStream/stream-chat-react/issues/1403

### 🛠 Implementation details

The original `index.ts` only exported two hooks from the `src/components/MessageInput/hooks` directory. This PR simply exports all of the hooks in that directory.

### 🎨 UI Changes

N/A


petyosi:
The underlying motivation is a valid one, but we need to find a way to make this more robust. Exporting those hooks can prevent us from refactoring the area later.

calvinwyoung:
@petyosi Thanks for the quick response. Yes this makes sense — I certainly wouldn't to increase your public API surface by too much.

But as a compromise, what if we *only* exported the hooks that are strictly required for implementing custom trigger providers? This includes:

- `useCommandTrigger`
- `useEmojiTrigger`
- `useUserTrigger`

Does that keep the public API surface sufficiently manageable? We're stuck in a difficult situation because we *need* a custom `TriggerProvider` (as is recommended by your docs), but by copying the implementation into our codebase, we risk something breaking if your team updates some of the underlying trigger behavior.

Thanks again for your consideration!

PR #1407, opened 2022-03-02
fix: Slow mode role check and limitations

szuperaz:
# Submit a pull request

### 🎯 Goal

Previously slow mode used built-in role check, now the `slow-mode` capability is used.
Previously the cooldown timer was reset after channel change, and wasn't synced between the thread and main channel (if the user sent a message in the channel, and then opened a thread, the thread didn't have the cooldown timer).

### 🛠 Implementation details

We track when the current chat user sent the latest message to each channel (this is updated when a `message.new` event is received in `Channel.tsx`). When the user changes channel, opens a thread or sends a message we check whether or not a cooldown should be activated (done in `useCooldownTimer.tsx`). 

### 🎨 UI Changes

_Add relevant screenshots_


src/components/MessageInput/hooks/useCooldownTimer.tsx
```diff
@@ -53,29 +41,32 @@ export type CooldownTimerState = {
...
+  const { latestMessageDatesByChannels } = useChatContext<StreamChatGenerics>('useCooldownTimer');
   const { channel } = useChannelStateContext<StreamChatGenerics>('useCooldownTimer');
 
-  const { cooldown: cooldownInterval } = (channel.data ||
+  const { cooldown: cooldownInterval, own_capabilities } = (channel.data ||
     {}) as ChannelResponse<StreamChatGenerics>;
 
   const [cooldownRemaining, setCooldownRemaining] = useState<number>();
 
-  const globalRole = client.user?.role || '';
```
ambaldwin, line 64: I just checked documentation for own_capabilities and looks like these are at the channel level. Do we still need to account for global roles like moderator and admin?

ambaldwin (changes requested):
I left one comment, and the only other thing is to add `latestMessageDatesByChannel` to the `ChatContext` documentation - docusaurus/docs/React/contexts/chat-context.mdx

src/components/MessageInput/hooks/useCooldownTimer.tsx
szuperaz: I think we don't because own_capabilities also takes [permissions](https://getstream.io/chat/docs/javascript/chat_permission_policies/?language=javascript) into account, which in my understanding contain global role

szuperaz:
> I left one comment, and the only other thing is to add `latestMessageDatesByChannel` to the `ChatContext` documentation - docusaurus/docs/React/contexts/chat-context.mdx

Thanks, will add that

szuperaz:
> > I left one comment, and the only other thing is to add `latestMessageDatesByChannel` to the `ChatContext` documentation - docusaurus/docs/React/contexts/chat-context.mdx
> 
> Thanks, will add that

Updated

src/components/MessageInput/hooks/useCooldownTimer.tsx
ambaldwin: ok! I will approve.

ambaldwin (approved):

src/components/MessageInput/hooks/useCooldownTimer.tsx
```diff
@@ -53,29 +41,32 @@ export type CooldownTimerState = {
...
+    const remainingCooldown = Math.round(
+      cooldownInterval - (new Date().getTime() - latestMessageDate.getTime()) / 1000,
+    );
+    if (remainingCooldown > 0 && !skipCooldown) {
+      setCooldownRemaining(remainingCooldown);
+    }
+  }, [channel.id, cooldownInterval, latestMessageDatesByChannels[channel.cid]]);
+
+  return {
+    cooldownInterval: cooldownRemaining || cooldownInterval || 0,
```
MartinCupela, line 68:
```suggestion
    cooldownInterval: cooldownRemaining || 0,
```

MartinCupela, line 68: I was looking into the use of cooldownInterval and it seems that you always want cooldownRemaining. But maybe I am missing something?

MartinCupela (changes requested):

src/components/MessageInput/hooks/useCooldownTimer.tsx
szuperaz: Well, this bit is a bit of hack, previously the `cooldownInterval` was the fixed value of the cooldown period, and `cooldownRemaining` was the value that showed how much time is remaining in the current timer.

The problem was that `cooldownInterval` was used as a prop to the CooldownTimer component. But this was not the right behavior because if you switch channels while during the cooldown, the cooldown should not start from `cooldownInterval` but rather from `cooldownRemaining`.

But since `cooldownInterval` originally means the fixed value of the timeout, it seemed wrong to always return cooldownRemaining since that could be 0.

The proper solution would've been to use `cooldownRemaining` as the prop for the CooldownTimer component, but I was too chicken to update everywhere, not being sure if that would cause issues elsewhere.

I'm not sure if this was clear or not.

szuperaz: Hm, after seeing how much explanation this code snippet needed, I think it's best if I choose the right solution, not the hacky one and update the props of the `CooldownTimer` components.

szuperaz: @MartinCupela done

PR #1438, opened 2022-03-07
Dont hide message actions dialog on hover away

### LINKED ISSUE #1401: Message actions dialog disappears once not hovered
It can be annoying for a user if the dialog disappears after a smallest move of the cursor in a wrong direction. From a UX point of view the menu should be visible until the user either:
- clicks outside
- presses Escape key
- presses on an item in the menu

MartinCupela:
### 🎯 Goal

The original behavior of actions box to disappear on hover away keeps being opt in by passing the `messageWrapperRef`. However the `MessageSimple` component does not pass neither create this ref anymore to `MessageOption` (and thus down to `MessagActions`). Moreover event listener for `keyup` event is now registered and hides message actions box on pressing `Escape` key only.

These changes then lead to the message actions box being hidden when user:
- clicks outside
- presses Escape key
- presses on an item in the menu

fixes #1401

### DIFF
--- a/src/components/MessageActions/MessageActions.tsx
+++ b/src/components/MessageActions/MessageActions.tsx
@@ -71,7 +71,12 @@ export const MessageActions = <
 
   const isMuted = useCallback(() => isUserMuted(message, mutes), [message, mutes]);
 
-  const hideOptions = useCallback(() => setActionsBoxOpen(false), []);
+  const hideOptions = useCallback((event: MouseEvent | KeyboardEvent) => {
+    if (event instanceof KeyboardEvent && event.key !== 'Escape') {
+      return;
+    }
+    setActionsBoxOpen(false);
+  }, []);
   const messageActions = getMessageActions();
   const messageDeletedAt = !!message?.deleted_at;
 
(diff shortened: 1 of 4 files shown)


petyosi (approved):
I like the new default behavior - it's much more user-friendly. 

I am, however, confused by the documentation aspect, and I feel like it is more of an internal implementation detail - maybe I am not seeing the overall picture. Here's what confuses me. 

- Why should I, as an integrator, know about `messageWrapperRef`? 
- Why its presence (or lack of such) affects the `mouseleave` behavior?

I will mark that as an approval, but feel free to introduce an additional iteration if you feel like it.

src/components/MessageActions/MessageActions.tsx
```diff
@@ -90,8 +95,10 @@ export const MessageActions = <
   useEffect(() => {
     if (actionsBoxOpen) {
```
arnautov-anton, line 96: image

```jsx
useEffect(() => {
	if (!actionsBoxOpen) return;

	document.addEventListener("click", hideOptions);
	document.addEventListener("keyup", hideOptions);

	return () => {
		document.removeEventListener("click", hideOptions);
		document.removeEventListener("keyup", hideOptions);
	};
}, [actionsBoxOpen, hideOptions]);
```
Just a suggestion.

arnautov-anton: This adds event listeners only if `actionBoxOpen` is set to `true` and removes them once it changes back to `false`.

arnautov-anton (approved):
LGTM :taco:

petyosi (dismissed):

PR #1439, opened 2022-03-07
fix: make scrollToLatestMessageOnFocus smarter

### LINKED ISSUE #1415: scrollToLatestMessageOnFocus should scroll only if new messages have been received
As an end-user, I don't want to have my message list scroll location reset when I re-activate the document.

**Acceptance Criteria**
* When the property is turned on, 
* An the end user scrolls upwards
* And moves the page to background (going to another tab or window)
* And then focuses back
* If no new messages are received in the meantime, the scroll location should remain unchanged.

petyosi:
### 🎯 Goal
This change prevents the unnecessary reset of scroll position if no new messages have been received.

Fixes #1415

### DIFF
--- a/src/components/MessageList/VirtualizedMessageList.tsx
+++ b/src/components/MessageList/VirtualizedMessageList.tsx
@@ -185,7 +185,7 @@ const VirtualizedMessageListWithContext = <
 
   const resetNewMessagesReceivedInBackground = useCallback(() => {
     setNewMessagesReceivedInBackground(false);
-  }, [setNewMessagesReceivedInBackground]);
+  }, []);
 
   useEffect(() => {
     setNewMessagesReceivedInBackground(true);
--- a/src/components/MessageList/VirtualizedMessageList.tsx
+++ b/src/components/MessageList/VirtualizedMessageList.tsx
@@ -179,21 +179,39 @@ const VirtualizedMessageListWithContext = <
     setNewMessagesNotification(false);
   }, [virtuoso, processedMessages, setNewMessagesNotification, processedMessages.length]);
 
+  const [newMessagesReceivedInBackground, setNewMessagesReceivedInBackground] = React.useState(
+    false,
+  );
+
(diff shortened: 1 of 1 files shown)


src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -179,21 +179,39 @@ const VirtualizedMessageListWithContext = <
     setNewMessagesNotification(false);
   }, [virtuoso, processedMessages, setNewMessagesNotification, processedMessages.length]);
 
+  const [newMessagesReceivedInBackground, setNewMessagesReceivedInBackground] = React.useState(
+    false,
+  );
+
+  const resetNewMessagesReceivedInBackground = useCallback(() => {
+    setNewMessagesReceivedInBackground(false);
+  }, [setNewMessagesReceivedInBackground]);
```
arnautov-anton, line 188:
```suggestion
  }, []);
```
You can omit this one, `setState` functions are "stable" and won't change.

MartinCupela (approved):

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -179,21 +179,39 @@ const VirtualizedMessageListWithContext = <
     setNewMessagesNotification(false);
   }, [virtuoso, processedMessages, setNewMessagesNotification, processedMessages.length]);
 
+  const [newMessagesReceivedInBackground, setNewMessagesReceivedInBackground] = React.useState(
+    false,
+  );
+
+  const resetNewMessagesReceivedInBackground = useCallback(() => {
```
MartinCupela, line 186: @petyosi shouldn't these changes be tested?

petyosi: They should, but we don't have the infrastructure for e2e testing set up. Testing that in a simulated rendering is pointless.

arnautov-anton (approved):
LGTM :taco:

PR #1458, opened 2022-03-18
fix: Update channel preview on connection recovery

### LINKED ISSUE #1351: Preview for channel with unread message isn't rendered as bold  .... after connection recovered.
Say I have Channel A and B. I am on Channel A. If I receive a message from Channel B, the Preview in ChannelList would display the new message snippet and become bolded to indicate that there is a new unread message. This works perfectly realtime when the websocket is well and alive.

Now I turn off Wifi. Someone sends some messages in Channel B. I turn Wifi back on ....
The Preview for Channel B would actually pick up the new message and render the snippet, BUT .... the Preview isn't bolded somehow (as if there is no unread message). If I switch to other unrelated Channel (or reload the page), Channel B's Preview would be rendered as bold correctly.

MartinCupela:
### 🎯 Goal

Update unread counts in `ChannelPreview` once `connection.recovered` is dispatched and thus update visual cues of updates in the `ChannelPreview`.

### 🛠 Implementation details

`ChannelPreviewProps` exposes  `channelUpdateCount`. It is an update counter passed down from `ChannelList`. Its incremented on the following events being emitted:

- `'channel.truncated'`
- `'channel.updated'`
- `'connection.recovered'`

From now the effect that updates local unread count state in `ChannelPreview` takes  `channelUpdateCount` as a dependency.

fixes #1351


petyosi (approved):

src/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -82,23 +82,20 @@ export const ChannelPreview = <
...
 
-  useEffect(() => {
+  const refreshUnreadCount = useCallback(() => {
     if (isActive || muted) {
       setUnread(0);
     } else {
       setUnread(channel.countUnread());
     }
-  }, [channel, isActive, muted]);
+  }, [channel, isActive, muted, channelUpdateCount]);
```
arnautov-anton, line 91:
```suggestion
  }, [channel, isActive, muted]);
```
Is this dependency needed? It's not being used within the scope of that function.

src/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -110,7 +107,7 @@ export const ChannelPreview = <
       channel.off('message.updated', handleEvent);
       channel.off('message.deleted', handleEvent);
     };
-  }, [channel, isActive, muted]);
+  }, [refreshUnreadCount]);
```
arnautov-anton, line 110:
```suggestion
  }, [channelUpdateCount]);
```
I think this might be the place you'd want to use it instead?

src/components/ChannelPreview/ChannelPreview.tsx
MartinCupela: @arnautov-anton Yes, it is needed - it is a value incremented to force rerender of the `ChannelPreview` component. It is used to notify `ChannelPreview` component that connection has been recovered. It is the actual fix that solves the problem that the component have not reacted to the `'connection.recovered'` event. By registering the the listener one level above you create only one listener for connection recovery instead of registering it in every `ChannelPreview` component.

arnautov-anton (approved):
LGTM :taco:

PR #1471, opened 2022-03-22
chore(tests): End-to-end testing infra + first test

jeevcat:
# Submit a pull request

### 🎯 Goal

* Add end-to-end testing of components against live backend

### 🛠 Implementation details

* Playwright as e2e framework and test runner
* Ladle as component tester (== Storybook)
* Tests following behaviors:
    * Channel List should update for current user
    * Channel List should update for channel members and show unread
    * Message list should update for different users on the channel

### 🎨 UI Changes

None


jeevcat:
@petyosi should e2e run in CI?

petyosi (changes requested):
Pushed a few tweaks to the part with env variables, but the setup looks right, and works for me. Here's what I see as necessary:
- add `.env.example` so that the developers have a reference point for their local setup;
- add the necessary env variables in the repo (I gave you permissions);
- modify the GitHub workflow so that the e2e tests are run on PR, etc. 

I personally prefer to have separate commands for unit and e2e tests for greater flexibility.

petyosi:
> @petyosi should e2e run in CI?

Yes! it should, but through a separate step - otherwise, devs won't have the option to run unit tests only.

petyosi:
We need this in the workflow: https://playwright.dev/docs/cli#install-browsers

jeevcat:
@petyosi I've addressed all requested changes. I also "improved" the workflow a little https://github.com/GetStream/stream-chat-react/actions/runs/2040487332.

petyosi (dismissed):
LGTM. Let's have the rest of the team review it, so that they can start using the setup as well.

MartinCupela:
@jeevcat I apologize for the delay in the review. 

I have proposed some changes in the [e2e-proposals branch](https://github.com/GetStream/stream-chat-react#:~:text=See%20Dependabot%20alerts-,e2e-proposals,-had%20recent%20pushes) - tests were not passing if the channel list was long - there was overflow and the channel preview could not be clicked.

I was also thinking about the `ConnectedUser` component as it actually represents a whole app instead of only few necessary pieces stitched together. The question is, whether we will actually be building stories that will be composed of only few interacting components.

MartinCupela (changes requested):
@jeevcat I apologize for the delay in the review. 

I have proposed some changes in the [e2e-proposals branch](https://github.com/GetStream/stream-chat-react#:~:text=See%20Dependabot%20alerts-,e2e-proposals,-had%20recent%20pushes) - tests were not passing if the channel list was long - there was overflow and the channel preview could not be clicked.

I was also thinking about the `ConnectedUser` component as it actually represents a whole app instead of only few necessary pieces stitched together. The question is, whether we will actually be building stories that will be composed of only few interacting components.

jeevcat:
> I have proposed some changes in the [e2e-proposals branch](https://github.com/GetStream/stream-chat-react#:~:text=See%20Dependabot%20alerts-,e2e-proposals,-had%20recent%20pushes) - tests were not passing if the channel list was long - there was overflow and the channel preview could not be clicked.

Looks great. I'll merge these into this branch.

jeevcat:
> I was also thinking about the `ConnectedUser` component as it actually represents a whole app instead of only few necessary pieces stitched together. The question is, whether we will actually be building stories that will be composed of only few interacting components.

The ConnectsedUser component ended up growing to be most of a whole app, so I could test all outcomes I wanted. I would say, as a rule, stories should be the bare minimum required.

src/stories/connected-user.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+    </div>
+  );
+};
+
+// Sort in reverse order to avoid auto-selecting unread channel
+const sort: ChannelSort = { last_updated: 1 };
+
+const chatClient = StreamChat.getInstance<StreamChatGenerics>(apiKey);
+
+const ConnectedUser = ({ token, userId }: { token: string; userId: string }) => {
```
arnautov-anton, line 48: Took a look at this one and fixed it up a bit in Martin's branch [here](https://github.com/GetStream/stream-chat-react/compare/e2e...e2e-proposals) - other than that I think it's good to go. :taco:

e2e/fixtures.mjs
```diff
@@ -0,0 +1,55 @@
...
+    ADD_MESSAGE_CHANNEL,
+    APP_KEY,
+    APP_SECRET,
+    JUMP_TO_MESSAGE_CHANNEL,
+    TEST_USER_1,
+    TEST_USER_2,
+  } = Object.entries(process.env).reduce((acc, [key, value]) => {
+    acc[key.replace('E2E_', '')] = value;
+    return acc;
+  }, {});
```
arnautov-anton, line 18: Also not sure about this one but consider it a nitpick - if it's going to be used only once I think it's fine for the variables to have `E2E_` at the beginning. :)

src/stories/connected-user.stories.tsx
jeevcat: Thanks for cleaning this up. I was confused :D

e2e/fixtures.mjs
jeevcat: I'd agree. I've replaced this with a direct destructure now.

.ladle/styles.css
```diff
@@ -0,0 +1,10 @@
+.ladle-main {
+  padding: 0 1rem 3rem;
+}
+.str-chat {
+  height: 700px!important;
```
arnautov-anton, line 5:
```suggestion
  height: 700px !important;
```

arnautov-anton (dismissed):

MartinCupela (approved):

PR #1473, opened 2022-03-22
Fix Portuguese timestamp translation

hdiogenes:
### 🎯 Goal

Fix Portuguese timestamp translations.

### 🛠 Implementation details

The previous translation didn't make much sense to Portuguese speakers, so this patch corrects it.

Note: Portuguese is a gendered language, so while this fixes the translation for the feminine week days (Monday to Friday), Saturday and Sunday will be slightly wrong. They're masculine and should be written "Sábado passado" and "Domingo passado", but I don't know how to fix this right away.


petyosi:
Interesting. I don't know if i18n next (the lib we use) can't solve the gendered days of the week.

MartinCupela (approved):

PR #1494, opened 2022-04-04
fix: autocomplete items clicking works on webkit

petyosi:
# Submit a pull request

### 🎯 Goal
A user discovered that tapping the autocomplete suggestions does not work on the iPhone. 

### 🛠 Implementation details
The problem is due to [relatedTarget being null in webkit](https://stackoverflow.com/questions/42764494/blur-event-relatedtarget-returns-null). 
Adding a `tabindex` wrapper solves the issue.
The PR also adds testing in Webkit and Firefox.


arnautov-anton (commented):
Added few small changes for you to consider [here](https://github.com/GetStream/stream-chat-react/compare/fix-autocomplete-webkit...e2e-autocomplete).

szuperaz (dismissed):

arnautov-anton (dismissed):

PR #1480, opened 2022-03-29
test: fix warnings in message input

petyosi:
# Submit a pull request

### 🎯 Goal

The changeset fixes the warnings caused by the test execution of `src/components/MessageInput/__tests__/MessageInput.test.js`

### 🛠 Impementation details

- initial rendering is wrapped in an async `act`, to capture the translations promise
- button click is wrapped in an async `act`
- Finally, one long timeout was extracted and mocked

PR #1503, opened 2022-04-05
fix: Update quoted message preview on message update

### LINKED ISSUE #1405: QuotedMessagePreview keeps being displayed even after the parent message deletion
Somehow unintuitively, if a (parent) message to which the user is about to reply, is deleted before the reply submission, the reply textarea with parent message preview does not disappear (QuotedMessagePreview). Thus it is possible to reply to a message that does not exist.

Slack approaches this scenario by deleting the preview upon the parent message deletion. 

Steps to reproduce:
1. post a new message 

image

2. select Reply option in the actions box 

image

3. select the Delete option from the actions box - the placeholder message appears
 
image

4. post the text in the reply form - the failure message is displayed

image

MartinCupela:
### 🎯 Goal

The quoted message preview should reflect changes every time the original message being quoted is updated (incl. deleted). If the original message content is changed, then the quoted message preview has to reflect that change. If the message is deleted, then the quoted message preview should be canceled.

fixes #1405

### 🎨 UI Changes

[Demo video](https://user-images.githubusercontent.com/32706194/161866360-3fa25f66-08a5-4873-a517-79a0891a578e.mp4)


petyosi (approved):

MartinCupela:
> Is this PR in any way related to #1480? I noticed similar changes to `addNotification` function in both and so if they're related we could merge #1480 into this one.
> 
> Also, we should really get rid of the MessageInputSmall (in a separate PR), both this one and MessageInputFlat share like 95% of the code. We could add atributes to MessageInputFlat and export it as "small" as to not introduce breaking changes sort of like this:
> 
> ```tsx
> /**
>  * @deprecated - will be removed in the next major release
>  */
> export const MessageInputSmall = (props) => <MessageInputFlat size="small" {...props} />
> ```
> 

@arnautov-anton no please, do not merge them. That PR was just showing a way which this PR could take.

Regarding the deprecation of `MessageInputSmall` - I would suggest to do it in a separate PR in order this PR is not blocked. Also the aim of this PR is not to deprecate `MessageInputSmall`.

src/components/MessageInput/MessageInputFlat.tsx
```diff
@@ -52,6 +56,24 @@ export const MessageInputFlat = <
     SendButton = DefaultSendButton,
   } = useComponentContext<StreamChatGenerics>('MessageInputFlat');
 
+  useEffect(() => {
+    const handleQuotedMessageUpdate = (e: Event<StreamChatGenerics>) => {
+      if (!(quotedMessage && e.message?.id === quotedMessage.id)) return;
```
arnautov-anton, line 61:
```suggestion
      if (e.message?.id !== quotedMessage?.id) return;
```

arnautov-anton (approved):

PR #1514, opened 2022-04-13
fix: update `stream-chat-js` to version 6.5.0 containing the `markRead` fix, add E2E tests

arnautov-anton:
# Submit a pull request

### 🎯 Goal

Implementation of custom channel switching with the tools provided by the SDK would cause `channel.markRead` method to not run at all in certain situations. You can reproduce this issue with the simple setup below, two different users, three separate channels and these reproduction steps:

1. make sure you have two separate users running side by side (either in tabs or in one application)
2. send a random message to each of the channels with `user-1`
3. with `user-2`, you should have three channels loaded in the list (`channel-1`, `channel-2` and `channel-3`), `channel-1` is active by default thus unread count is 0
4. click on `channel-2` which should have unread count 1, `channel.markRead` method does not get triggered yet the unread count gets set to 0
5. click back on `channel-1`, visual state of `channel-2` gets set back to 1
6. click back on `channel-2`, `channel.markRead` gets triggered, visual state updates to 0 
7. repeat steps 1-3, and try switching between channels 2 and 3 with similar results though after switching to `channel-3` for the second time (considering you switched to `channel-2` the first time), visual state does not update and neither does the server state (`channel.markRead` does not get triggered)
8. reload the page, `channel-3` has still unread count of 1

image

### 🛠 Implementation details
- add E2E tests
- ~provide `channel.lastRead` attribute to `channel.countUnread` method~
- refactor of the thread useEffect hook
- fix provided in `stream-chat-js` with [this PR](https://github.com/GetStream/stream-chat-js/pull/955)


MartinCupela:
@arnautov-anton could you add some e2e tests as a check for a potential future regression?

src/components/Channel/Channel.tsx
```diff
@@ -418,14 +418,11 @@ const ChannelInner = <
   }, [channel.cid, doMarkReadRequest]);
 
   useEffect(() => {
-    if (state.thread && state.messages?.length) {
-      for (let i = state.messages.length - 1; i >= 0; i -= 1) {
```
petyosi, line 422: This seems like a micro-optimization to search the messages backwards. I would rather swap that for readability.

petyosi (dismissed):

MartinCupela (approved):

PR #1516, opened 2022-04-14
fix: prevent duplicate message id generation in message list

### LINKED ISSUE #1510: MessageList component throws 'duplicated key' error when provided an array of messages
I'm using the `MessageList` component with the prop `messages` to which I pass an array of filtered messages but it always throws the 'duplicated key' warning in the console. I made sure that there are no repeated messages in the array that is passed but it always causes the error. If the `messages` prop is not passed everything works as expected. Any clue on what I might be missing?

Here's how I am using the component:

```
 <MessageList
        onlySenderCanEdit
        closeReactionSelectorOnClick
        formatDate={formatDate}
        hideDeletedMessages
        hideNewMessageSeparator
        Message={MyMessage}
        messageActions={[]}
        {...messageListProps}
      />
```

In which `messageListProps` is just an object which has the `messages` array or is otherwise empty.

EDIT:

I was digging deeper and it seems that when I pass the array `[{id: 'some-message-id', text: 'Some message'}]` as the `messages` prop when I check the prop `enrichedMessages` inside the library's `useMessageListElements` hook it always has an added element like `{
    "customType": "message.date",
    "date": "2022-04-11T16:07:04.444474Z",
    "id": "some-message-id"
}`. 

For some reason this custom type message is added with the same id as the first element of the passed `messages` array.

EDIT2: 

I believe the problem is [here](https://github.com/GetStream/stream-chat-react/blob/0a812b4e494ceae5505465bd856e66f5d3155918/src/components/MessageList/utils.ts#L95) because although this message should have been filtered [here](https://github.com/GetStream/stream-chat-react/blob/0a812b4e494ceae5505465bd856e66f5d3155918/src/components/MessageList/hooks/useMessageListElements.tsx#L79) it's ending up the in the final array.

MartinCupela:
### 🎯 Goal

Fix the bug that generated custom messages of type `'message.date'` with duplicate id as the message they preceded.

fixes #1510 

### 🛠 Implementation details

This PR adds a refactor to the use of custom message type values `'message.date'` and `'channel.intro'`. These values were scattered all around the code base. If they were to be changed, the maintainer would have to make sure those values are changed correctly in all the places. Now the values are referenced by a `CUSTOM_MESSAGE` constant dictionary in a new folder `src/constants`.

### DIFF
--- a/src/components/Message/MessageCommerce.tsx
+++ b/src/components/Message/MessageCommerce.tsx
@@ -8,7 +8,7 @@ import { MessageTimestamp as DefaultMessageTimestamp } from './MessageTimestamp'
 import { areMessageUIPropsEqual, messageHasAttachments, messageHasReactions } from './utils';
 
 import { Avatar as DefaultAvatar } from '../Avatar';
-import { CUSTOM_MESSAGE } from '../../constants/messageTypes';
+import { CUSTOM_MESSAGE_TYPE } from '../../constants/messageTypes';
 import { MML } from '../MML';
 import {
   ReactionSelector as DefaultReactionSelector,
@@ -74,7 +74,7 @@ const MessageCommerceWithContext = <
     return <MessageDeleted message={message} />;
   }
 
-  if (message.customType === CUSTOM_MESSAGE.date) {
+  if (message.customType === CUSTOM_MESSAGE_TYPE.date) {
     return null;
   }
 
--- a/src/components/Message/MessageCommerce.tsx
+++ b/src/components/Message/MessageCommerce.tsx
@@ -8,6 +8,7 @@ import { MessageTimestamp as DefaultMessageTimestamp } from './MessageTimestamp'
 import { areMessageUIPropsEqual, messageHasAttachments, messageHasReactions } from './utils';
 
 import { Avatar as DefaultAvatar } from '../Avatar';
+import { CUSTOM_MESSAGE } from '../../constants/messageTypes';
 import { MML } from '../MML';
 import {
   ReactionSelector as DefaultReactionSelector,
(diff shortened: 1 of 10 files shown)


src/components/Message/MessageCommerce.tsx
```diff
@@ -73,7 +74,7 @@ const MessageCommerceWithContext = <
     return <MessageDeleted message={message} />;
   }
 
-  if (message.customType === 'message.date') {
+  if (message.customType === CUSTOM_MESSAGE.date) {
```
petyosi, line 77: This does not read very well - you compare `type` to `date`. Should not the constant be `CUSTOM_MESSAGE_TYPE`?

petyosi (approved):
Looks good, feel free to apply or ignore my suggestion.

src/components/Message/MessageCommerce.tsx
MartinCupela: I was considering, whether to use `_TYPE ` or not and originally decided to go with the shorter version. But after your feedback I will change it to `CUSTOM_MESSAGE_TYPE`.

petyosi (approved):

src/constants/messageTypes.ts
```diff
@@ -0,0 +1,6 @@
+import type { CustomMessageType } from '../types/types';
+
+export const CUSTOM_MESSAGE_TYPE: Record<'date' | 'intro', CustomMessageType> = {
+  date: 'message.date',
+  intro: 'channel.intro',
+};
```
arnautov-anton, line 6:
```suggestion
export const CUSTOM_MESSAGE_TYPE = {
  date: 'message.date',
  intro: 'channel.intro',
} as const;
```

arnautov-anton: That `Record` type casting would mean that `CUSTOM_MESSAGE_TYPE.date` could be either `message.date` or `channel.intro`:
image
Let's make it type constant so the types look like this:
image

MartinCupela: Added const assertion

arnautov-anton (approved):

petyosi (approved):

PR #1544, opened 2022-04-25
fix: remove `stream-chat-react-<version>` from client User-Agent

arnautov-anton:
### 🎯 Goal

Remove everything revolving around `version.ts` file which seems like it's not being properly used since PR #1250.

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,11 +183,10 @@
   },
   "scripts": {
     "analyze": "yarn build -- --stats && webpack-bundle-analyzer build/bundle-stats.json",
-    "build": "rm -rf dist && mkdir -p dist/assets assets && yarn --silent copy-version && yarn build-translations && yarn types && yarn bundle",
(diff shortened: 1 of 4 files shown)


petyosi:
@arnautov-anton why do you think that it's not properly used? as far as I can see, the correct version is set here: https://unpkg.com/browse/stream-chat-react@8.1.1/dist/

arnautov-anton:
> @arnautov-anton why do you think that it's not properly used? as far as I can see, the correct version is set here: https://unpkg.com/browse/stream-chat-react@8.1.1/dist/

image

Didn't think that version is being updated and it always stays on `0.0.0-development`. Nevermind, closing, thanks for the heads up. :+1:

arnautov-anton:
> @arnautov-anton why do you think that it's not properly used? as far as I can see, the correct version is set here: https://unpkg.com/browse/stream-chat-react@8.1.1/dist/

image

Didn't think that version is being updated and it always stays on `0.0.0-development`. Nevermind, closing, thanks for the heads up. :+1:

PR #1545, opened 2022-04-25
docs: `connect/disconnect(User)`/instantiation code snippets update

arnautov-anton:
### 🎯 Goal

Update documentation to use basic instantiation rather than using `StreamChat.getInstance` which causes connection issues for the users who tend to switch between user profiles within one application.


MartinCupela:
@arnautov-anton are these the only files where we recommend to use `Stream.getInstance`? Do we have some other tutorials using it somewhere else?

MartinCupela (approved):

PR #1553, opened 2022-04-27
test: fix randomly failing E2E tests

arnautov-anton:
### 🎯 Goal

Get rid of excessive `setState` action in `ConnectedUser` component and adjust Playwright configuration.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -3,8 +3,8 @@ import type { PlaywrightTestConfig } from '@playwright/test';
 // https://playwright.dev/docs/test-configuration
 const config: PlaywrightTestConfig = {
   testDir: './e2e',
+  timeout: 15 * 1000,
   use: {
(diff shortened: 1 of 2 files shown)


playwright.config.ts
```diff
@@ -3,8 +3,8 @@ import type { PlaywrightTestConfig } from '@playwright/test';
 // https://playwright.dev/docs/test-configuration
 const config: PlaywrightTestConfig = {
   testDir: './e2e',
+  timeout: 15 * 1000,
   use: {
-    headless: true,
```
petyosi, line 7: Should not this be explicitly set? Or it's the default value.

MartinCupela: The default value is `false`. If removed, the browser windows do not appear.

MartinCupela (approved):

PR #1559, opened 2022-04-29
test: fix race conditions

arnautov-anton:
### 🎯 Goal

Fix race conditions which make E2E tests fail from time to time.


arnautov-anton:
Closing in favor of #1569.

PR #1566, opened 2022-05-05
fix: propagate user update data to UI

### LINKED ISSUE #1565: bug: neither `ChannelPreview` nor `MessageSimple` reflect changes signaled by `user.updated` event
**Describe the bug**

If a user's data get updated, the changes are not reflected in UI (`ChannelPreview`, `MessageSimple`) until the page is refreshed.

**To Reproduce**

1. Set up a direct messaging channel (2 members)
2. Open two browser windows each for one member of the channel
3. Send HTTP request to update data of the first user (`await client.partialUpdateUser({ id: client.user.id, set: {name:'new-name'} })`)
4. Observe in the browser's Network tab that `user.updated` event is received in the both windows
5. Observe that the message author,  `Avatar` title neither `ChannelPreview` title get updated

**Expected behavior**

Message author,  `Avatar` title neither `ChannelPreview` title get updated

**Screenshots**

**Package version**
- stream-chat-react: 8.1.2
- stream-chat-css: 2.8.0
- stream-chat-js: 6.5.0

MartinCupela:
### 🎯 Goal

Reflect updates to user's data in:

* 2-members channel preview
* Avatar (in MessageList, Thread, ChannelPreview)

fixes #1565

#### `MessageList`
The change's are performed upon reception of WS event `user.updated`. The implementation takes into account that a user can have custom fields and therefore the memoization in `MessageList` takes this into account. Until now, the changes were propagated to the most bottom components only if user's `image` or `name` changed. Those are default attributes. Therefore if a custom `Avatar` component displays custom user data, their update would not be reflected until page is refreshed.

#### `ChannelPreview`
The `ChannelList` component neither its children were listening to `user.updated` event and therefore the event listener has been added to the `ChannelPreview`'s effect. 

#### Out of scope
`ChannelHeader` does not display image if the channel is DM channel (`channel.data.name === undefined`). This could be considered a bug or insufficiency. The `ChannelHeader` should probably behave the same way as `ChannelPreview` in terms of its title and displayed image to support consistency. This is out of scope of this PR, because it would introduced breaking changes in form of changed component behavior.

### DIFF
--- a/src/components/ChannelPreview/ChannelPreview.tsx
+++ b/src/components/ChannelPreview/ChannelPreview.tsx
@@ -112,15 +112,16 @@ export const ChannelPreview = <
 
   useEffect(() => {
     const handleEvent = () => {
-      const newDisplayTitle = getDisplayTitle(channel, client.user);
-      const newDisplayImage = getDisplayImage(channel, client.user);
-      if (newDisplayTitle !== displayTitle) {
-        setDisplayTitle(newDisplayTitle);
-      }
-      if (newDisplayImage !== displayImage) {
-        setDisplayImage(newDisplayImage);
-      }
+      setDisplayTitle((displayTitle) => {
+        const newDisplayTitle = getDisplayTitle(channel, client.user);
+        return displayTitle !== newDisplayTitle ? newDisplayTitle : displayTitle;
+      });
+      setDisplayImage((displayImage) => {
+        const newDisplayImage = getDisplayImage(channel, client.user);
+        return displayImage !== newDisplayImage ? newDisplayImage : displayImage;
+      });
     };
+
     client.on('user.updated', handleEvent);
     return () => {
       client.off('user.updated', handleEvent);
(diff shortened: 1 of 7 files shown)


src/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -109,10 +110,25 @@ export const ChannelPreview = <
...
+  useEffect(() => {
+    const handleEvent = () => {
+      const newDisplayTitle = getDisplayTitle(channel, client.user);
+      const newDisplayImage = getDisplayImage(channel, client.user);
+      if (newDisplayTitle !== displayTitle) {
+        setDisplayTitle(newDisplayTitle);
+      }
+      if (newDisplayImage !== displayImage) {
+        setDisplayImage(newDisplayImage);
+      }
```
arnautov-anton, line 122: Let's not rely on "outside" values and rather use previous value inside setState action:
```tsx
setDisplayTitle((previousValue) => previousValue !== newValue ? newValue : previousValue);
```
Returning the same value won't trigger the rerender as [stated here](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update).

MartinCupela: @arnautov-anton thank you for pointing this out. I have updated the snippet.

arnautov-anton (approved):

PR #1569, opened 2022-05-08
refactor: write e2e testcases as user stories

MartinCupela:
### 🎯 Goal

Make E2E testing more attractive by providing boilerplate:

1.  to generate more readable code similar to human readable text
2. to generate code faster by avoiding to repeatedly rewrite selectors and other hard to remember things

### 🛠 Implementation details

1. Create custom fixtures `user` and `controller`
- `user` - interacts with the UI
- `controller` performs actions that a user would not normally do (send requests without interacting with the chat UI etc. - `controller.activateChannel()`, `await controller.openStory('add-message--user1', selectors.channelPreviewButton)`)
2. Provide custom fixtures via custom [ extension of `test`](https://playwright.dev/docs/test-fixtures#overriding-fixtures)

The aim is to hide the (still) low level interaction with Playwright API behind Jasmine style expressions for user interactions:

- `user.sees.MessageList.not.empty()`
- `user.sees.ChannelPreview.read()`


petyosi (dismissed):

MartinCupela:
@petyosi I decided not to go with extending the `expect` function. My goal is to provide curated high-level API for interacting with our components combined with reusable assertion + improved readability. Extended `expect` would not be able to meet both goals. It would also probably lead to clutter of assertion functions specific to many Stream components.

At the moment I feel there is a pattern in how to access methods of that API (`user.(sees | clicks | get | typesTo).<Component>.<component-specific-routine>`).

arnautov-anton (approved):

PR #1571, opened 2022-05-10
fix: pre-filter readBy array and rid it of nullish values

arnautov-anton:
### 🎯 Goal

Customer states that their application crashes when `lastReadUser` object tries to access the `image` property meaning that object `lastReadUser` must be either `null` or `undefined`. This PR adds pre-filtering of nullish values of the `readBy` array.

### DIFF
--- a/src/components/Message/MessageStatus.tsx
+++ b/src/components/Message/MessageStatus.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
 import { DeliveredCheckIcon } from './icons';
 import { getReadByTooltipText } from './utils';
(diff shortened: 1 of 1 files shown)


arnautov-anton:
I'd like to know why is that even the case - why does the `readBy` array even contain such values?

MartinCupela (approved):

src/components/Message/MessageStatus.tsx
```diff
@@ -57,15 +59,15 @@ const UnMemoizedMessageStatus = <
...
     return (
       <span
         className={`str-chat__message-${messageType}-status`}
         data-testid='message-status-read-by'
       >
-        <Tooltip>{getReadByTooltipText(readBy, t, client)}</Tooltip>
+        <Tooltip>{getReadByTooltipText(filteredReadBy, t, client)}</Tooltip>
         <Avatar
           image={lastReadUser.image}
           name={lastReadUser.name || lastReadUser.id}
```
vishalnarkhede, line 73: Can we add the guards here just in case? `lastReadUser?.image` and `lastReadUser?.name`

I know its not required anymore ... but to avoid the hassle regardless in case the issue persists

arnautov-anton:
Closing as the issue was resolved by upgrading their package to the latest version.

PR #1598, opened 2022-05-17
fix: filter out OG scraped images from attachments

arnautov-anton:
### 🎯 Goal

Every time user updated message with URL with OG attachments, additional attachment (image) has been added to the array of attachments.

### 🛠 Implementation details

One of the solutions is to filter out any attachments which contain `og_scrape_url` property before submission so the BE can re-fetch them. The solution is fairly simple but does not work for when only attachments are updated thus not triggering BE re-fetch leaving the message without OG attachments.
Second solution is to keep all of the needed properties of the OG attachments (`title`, `title_link`, `author_name`, `text` and `og_scrape_url`) when updating the message but then we'd have to filter the OG attachments based on whether the `og_scrape_url` still exists within the message text.

Due to [BE implementation](https://www.notion.so/stream-wiki/OG-OpenGraph-URL-Enrichment-35bb3b6fdacf41e8b4dab59232a528aa) I opted for third solution which checks whether the text has been edited or not - if it was, the OG attachments get filtered out thus triggering re-enrichment.

### 🎨 UI Changes

(message after edit with two excessive attachments)
image


e2e/edit-message.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  }) => {
+    await user.clicks.MessageActionsBox.editMessage('', 0);
+
+    await user.typesTo.MessageInput.text('');
+
+    // remove all attachments
+    const locator = page.locator(selectors.buttonCancelUpload);
+    const count = await locator.count();
+    for (let i = 0; i < count; ++i) {
+      await user.clicks.Modal.removeAttachment(0);
```
MartinCupela, line 88:
```suggestion
      await user.clicks.EditMessageForm.removeAttachment(0);
```

src/components/MessageInput/hooks/useSubmitHandler.ts
```diff
@@ -42,37 +43,53 @@ export const useSubmitHandler = <
...
       ) => self.every((upload) => upload.id === id || upload.url !== url))
-      .map(
-        (upload) =>
-          ({
-            fallback: upload.file.name,
-            image_url: upload.url,
-            type: 'image',
-          } as Attachment<StreamChatGenerics>),
-      );
+      .filter((upload) => {
```
MartinCupela, line 66: @arnautov-anton probably needs unit tests to check that the logic of `hasChanged` is covered?

MartinCupela (changes requested):
Nice work :)

e2e/edit-message.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    await user.typesTo.MessageInput.text('jest: https://jestjs.io/docs/cli');
+
+    await Promise.all([
+      page.waitForResponse((r) => r.url().includes('/message') && r.ok()),
+      page.waitForSelector(`${selectors.attachmentCard} >> text=jestjs.io`),
+      user.clicks.Modal.send(),
+    ]);
+
+    expect(page.locator(selectors.attachmentCard)).toHaveCount(1);
```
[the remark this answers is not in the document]
arnautov-anton: Good point, still confused about the semantics. Will change this. :+1:

MartinCupela (dismissed):
Very nice refactor :+1:

MartinCupela (approved):

PR #1608, opened 2022-06-01
feat: use ResizeObserver to keep Channel scrolled to bottom on page load

MartinCupela:
### 🎯 Goal

The logic, that assures the channel to be scrolled to the bottom on page load uses `scrollToBottom` function call in multiple places and multiple hook effects. The current logic does not handle the scroll behavior on message list `scrollHeight` change robustly. It just sets timeout 200ms to scroll to the bottom after the last invocation of `scrollToBottom` function. On slower network the may images load slower than that leaving the channel not scrolled to the bottom.

### 🛠 Implementation details

Use `ResizeObserver` to keep the channel scrolled to the bottom until the user initiates own scroll behavior. Once user starts scrolling, the resize observer is cancelled.
The heuristic used to infer that the scroll has been user-initiated compares the total number of calls to `scrollToBottom` function and the number of invocations of `scroll` event handler on the message list. The resize observer is cancelled once:

`the number of `scrollToBottom` calls < the number of invocations of scroll event handler on the message list `.

Package `stream-chat-react` supports browsers with > 0.2% market share. Among those browsers, [there are some](https://caniuse.com/resizeobserver), that do not support the `ResizeObserver` implementation. For that reason a resize observer [ponyfill](https://github.com/sindresorhus/ponyfill) is used (https://github.com/juggle/resize-observer). The `ResizeObserver` ponyfill is also needed for unit tests.

### DIFF
--- a/src/components/MessageList/hooks/useScrollLocationLogic.tsx
+++ b/src/components/MessageList/hooks/useScrollLocationLogic.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useLayoutEffect, useRef, useState } from 'react';
+import React, { RefObject, useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
 
 import { useMessageListScrollManager } from './useMessageListScrollManager';
 
@@ -11,6 +11,7 @@ export type UseScrollLocationLogicParams<
 > = {
   hasMoreNewer: boolean;
   suppressAutoscroll: boolean;
+  ulRef: RefObject<HTMLUListElement>;
   currentUserId?: string;
   messages?: StreamMessage<StreamChatGenerics>[];
   scrolledUpThreshold?: number;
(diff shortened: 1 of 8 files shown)


oliverlaz (approved):

src/components/MessageList/hooks/useScrollLocationLogic.tsx
```diff
@@ -21,49 +26,78 @@ export const useScrollLocationLogic = <
...
   const scrollToBottom = useCallback(() => {
-    if (!listRef.current?.scrollTo || hasMoreNewer || suppressAutoscroll) {
+    if (!listElement?.scrollTo || hasMoreNewer || suppressAutoscroll) {
       return;
     }
 
-    listRef.current?.scrollTo({
-      top: listRef.current.scrollHeight,
+    scrollCounter.current.autoScroll += 1;
+    listElement?.scrollTo({
```
arnautov-anton, line 51:
```suggestion
    listElement.scrollTo({
```

arnautov-anton (dismissed):

arnautov-anton (approved):

oliverlaz (approved):

PR #1615, opened 2022-06-08
test(e2e): fix failing "navigate-long-message-lists" test

arnautov-anton:
### 🎯 Goal

Attempt number 34908 to fix E2E tests.

### 🛠 Implementation details

image

- added `type="button"` fix to `react-file-utils` which caused `Modal` to close when "close buttons" on attachments were clicked triggering form submit (`edit-message.test.ts` in Firefox)
- refactor to `EditMessageForm`
- refactor to `Modal` which now removes children (and itself) from the DOM when not opened


arnautov-anton:
`jump-to-message.test.ts` keeps failing on me, @petyosi (specifically - `regular jumps to message 29 and then back to bottom`) - Firefox and Chromium seem to be having the most trouble with it though this is not my final statement - browsers might have nothing to do with it.

image
image
This is second pass for the WebKit:
image

Do you think there might be an interference of this feature with the #1608, @MartinCupela?

MartinCupela:
> `jump-to-message.test.ts` keeps failing on me, @petyosi (specifically - `regular jumps to message 29 and then back to bottom`) - Firefox and Chromium seem to be having the most trouble with it though this is not my final statement - browsers might have nothing to do with it.
> 
> image image This is second pass for the WebKit: image
> 
> Do you think there might be an interference of this feature with the #1608, @MartinCupela?

@arnautov-anton looking into this

MartinCupela (approved):

PR #1618, opened 2022-06-08
fix: rely on MessageList scroll management in Thread component

MartinCupela:
### 🎯 Goal
1. When thread is open no additional pagination requests are executed for the query message list
2. When thread is open Thread's message list is scrolled to the bottom

Depends on https://github.com/GetStream/stream-chat-css/pull/123

##### User scenarios
*Prerequisite: the message list scroll height inside the thread has to be greater than the height of its parent forcing the scrolling behavior*

 - [X] User does not scroll and opens a thread
 - [X] User scrolls up to a message that has to be additionally loaded by pagination -> opens its thread

Out of scope (target of #1631):
 - User click quoted message that does not have to be loaded by additional HTTP request  -> opens its thread 
 - User click quoted message that has to be loaded by additional HTTP request  -> opens its thread

###### Expected behavior:
1. The thread is scrolled to the bottom on open
2. The pagination requests do not take place

###### E2E tests: 
All the above scenarios have to be covered
- [X] non-virtualized message list

###### Unit tests
- [X] All the have to be adjusted to the refactor + added new where needed

##### Documentation
- [X] Document the possibility to pass custom MessageListHead component to context
- [X] Document the component API changes

### 🛠 Implementation details

New component `MessageListHead` head was introduced. It contains the `Message` component rendering the parent message and `ThreadStart` component. These were originally excluded from the message list what broke the scroll logic. This PR changes this and includes the parent message with start thread indicator within the list of message elements generated by `useMessageListElements`.


oliverlaz (approved):

oliverlaz (approved):

MartinCupela:
Changing the target branch to theming-v2 as the breaking change in markup in thread message list (removal of div container around it is removed).

Issue #1623, opened 2022-06-11
bug: window not found during NextJS compilation. happens in index.cjs.js


kdsbatra:
**Describe the bug**

When using latest version with NextJS. it complains window not found during compilation. Happens on 9.1 and after

**To Reproduce**

1. `npm install stream-chat-react@latest`
2. Start NextJS server
3. Go to any page that loads/compiles stream-chat-react in your app
4. See error in browser and nextjs console (happens in dist/index.cjs.js around line 8770)

**Expected behavior**

There shouldn't be a error

**Screenshots**

**Package version**
- stream-chat-react: 9.1.0+
- stream-chat-css: ~
- stream-chat-js: 6.6.0

**Desktop (please complete the following information):**

- OS: [e.g. iOS]  MacOS
- Browser [e.g. chrome, safari]    Chrome
- Version [e.g. 22]     102.0.5005.61

**Additional context**

MartinCupela:
@kdsbatra please upgrade to v9.1.2 which contains the fix. Thank you

PR #1624, opened 2022-06-11
fix: add check if window is defined before using ResizeObserver

cabutler10:
# Submit a pull request

### 🎯 Goal

Solves open Issue #1623. Build will fail with Gatsby and Next.js if window is undefined.

### 🛠 Implementation details

Add check if window is defined


MartinCupela:
@cabutler10 thank you for the fix. Could you please amend the commit message in order it meets the conventional commit requirements? In this case `fix: add check if window is defined before using ResizeObserver`?

cabutler10:
@MartinCupela Is that what you were looking for?

MartinCupela (approved):

MartinCupela:
@cabutler10 thank you :)

Issue #1627, opened 2022-06-13
bug: messageActionsHandler in Message should include messageActions as part of the dependency array


jpsierens88:
**Describe the bug**

Currently, you have a [useCallback](https://github.com/GetStream/stream-chat-react/blob/HEAD/src/components/Message/Message.tsx#L95-L109) `messageActionsHandler` that handles the actions a user can make in a message, like `['delete', 'react']`. This useCallback has several dependencies, but its missing 1: `messageActions`. `messageActions` is the prop us users can pass through the `MessageList` to tell you what actions we want a user to have.

Currently, our `messageActions` is the result of a condition:
```
<MessageList messageActions={isMemberOfChannel && [MESSAGE_ACTIONS.edit, MESSAGE_ACTIONS.delete, MESSAGE_ACTIONS.react]} />
```

if the current user is a member of the channel, we allow some actions, if he's not, he cant do anything. 

What happens now is that when one of our users leaves the channel, he's still able to do the actions above, because your `messageActionsHandler` isnt being updated by a change in `messageActions`. Same thing the other way, when a user enters a chat, he cant take any defined action.

As a work around for now, I have this set up in our own custom message ui component:
```
const messageActionsHandler = useCallback(
    () =>
      getMessageActions(allowedActions, {
        canDelete,
        canEdit,
        canFlag,
        canMute,
        canPin: false,
        canQuote,
        canReact,
        canReply,
      }),
    [allowedActions, canDelete, canEdit, canFlag, canMute, canQuote, canReact, canReply]
  )
```
note that `allowedActions` is in the dependency array

**To Reproduce**

Add a `MessageList` component with a `messageActions` prop. Try conditionally passing actions to `messageActions`, the `messageActionsHandler` wont update itself because `messageActions` is not part of the dependencies

**Expected behavior**
`messageActionsHandler` should udpdate on changes of `messageActions` so that the UI can be properly updated

**Package version**
- stream-chat-react: 8.0.0
- stream-chat-css: n/a
- stream-chat-js: 6.5.1

petyosi:
Thanks for the report - we will look into that.

MartinCupela:
@jpsierens88 Could you please provide some more details regarding the following?

1. Do we understand well that you are trying to handle user permissions via `messageActions` enum?
2. Why have you decided to manage permissions via `messageActions` instead of interaction with the API?
3. How do you grant / remove user permissions on the API side?

Thank you

[...]

PR #1634, opened 2022-06-16
fix: respect updates to messageActions prop, support messageActions overrides in Thread

oliverlaz:
### 🎯 Goal
This PR fixes the issue described in #1627. While working on it, we discovered that `Thread` messages always have all messageActions enabled and afterwards filtered out based on user's permissions.

### 🛠 Implementation details
- `messageActionHandler` now respects the `messageActions` prop
- `Thread` component now accepts additional `messageActions` property which will later be passed down to the selected `MessageList` implementation


petyosi (approved):

MartinCupela:
@oliverlaz I was thinking about the usage implications of the prop `messageActions`. Here are the scenarios:

Issue #1641, opened 2022-06-20
bug: align UI breakpoints for Message component


oliverlaz:
**Describe the bug**

There is an inconsistency between UI breakpoints in the React SDK and its companion stylesheets in `stream-chat-css`.

In this SDK, one hook deals with UI breakpoints: [useBreakpoint#getDeviceWidth](https://github.com/GetStream/stream-chat-react/blob/da4be3ae23f501017397c32c5a5e143c9f89e48b/src/components/Message/hooks/useBreakpoint.ts#L9-L13).
Here, the breakpoints are set to:
- 768px for mobile
- 1024px for tablet
- \>= 1024px for desktop

Inside `stream-chat-css` the [`Message.scss`](https://github.com/GetStream/stream-chat-css/blob/71a565d5e050cda647663d8cc6584257d0ae0f0f/src/styles/Message.scss) defines other breakpoints (search for `@media screen`):
- 375px
- 414px
- 420px
- 960px

**Expected behavior**
Breakpoints should be aligned. At best, they shall be defined in our customer's projects.

**Package version**
- stream-chat-react: v9.1.3
- stream-chat-css: v2.10.0

oliverlaz:
Fixed and will be rolled out together with our new theming. See:
- https://github.com/GetStream/stream-chat-css/pull/138
- https://github.com/GetStream/stream-chat-react/pull/1648
- https://github.com/GetStream/stream-chat-angular/pull/322

PR #1648, opened 2022-06-23
Remove useMobilePress and useBreakpoint hooks

oliverlaz:
### 🎯 Goal

Remove useMobilePress and useBreakpoint hooks. Closes: #1641

### 🛠 Implementation details

**useMobilePress:**
Historically, this hook programmatically handled the user interaction with Message components by toggling `mobile-press` class upon user interaction. The goal of this operation was to have the message actions displayed on the screen.
Internally, we found a better solution by offloading this behavior to the browser and utilizing `:focus` and `:focus-within` CSS pseudo-selectors.

**useBreakpoint:**
This hook did hold the "programmatic" responsive UI breakpoints. We realized they aren't always in line with our stylesheet breakpoints and possibly with our customer's breakpoints. This misalignment was causing some inconsistencies and issues.
We are removing this hook because we believe defining UI breakpoints should be the responsibility of our customers.

**SearchResults:**
During the refactoring, we stumbled upon one side-effect where `popupResults` prop wasn't always respected. 
The fix of it could be a breaking change for a small percentage of our customers, but we believe this fix is the right thing to do.

### Related PRs:
- https://github.com/GetStream/stream-chat-css/pull/138
- https://github.com/GetStream/stream-chat-angular/pull/322


szuperaz:
I think if it's a breaking change we should have a note in the commit message body https://github.com/semantic-release/semantic-release#commit-message-format

szuperaz (dismissed):

oliverlaz:
> > I think if it's a breaking change we should have a note in the commit message body https://github.com/semantic-release/semantic-release#commit-message-format
> 
> I agree, I can see that you have pretty sizable message in the first commit so you can extend it to contain `BREAKING CHANGE:` and you should be set. 👍

Sure thing! I'll update the commit message accordingly 👍

MartinCupela (approved):

petyosi (approved):

PR #1661, opened 2022-07-05
ci(docusaurus): add staging enviroment

arnautov-anton:
### 🎯 Goal

Add ability to push documentation changes to `stream-chat-docusaurus` staging branch when new changes are pushed to the `develop` branch within this repository. 

### 🛠 Implementation details

Implementation [tested here](https://github.com/GetStream/stream-chat-react/runs/7200293672?check_suite_focus=true#step:2:1).

### DIFF
--- a/.github/workflows/docusaurus.yml
+++ b/.github/workflows/docusaurus.yml
@@ -2,17 +2,21 @@ name: docusaurus
 on:
   push:
     branches:
-      - master # default branch name for each repo that docs should be push from
+      - master
+      - develop
     paths:
       - docusaurus/**
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/docusaurus.yml
```diff
@@ -5,14 +5,17 @@ on:
       - master # default branch name for each repo that docs should be push from
```
szuperaz, line 5: @arnautov-anton I think you have to add `develop` here as well to trigger the action

arnautov-anton: Yes! Missed that one, thank you, Zita! :)

szuperaz (approved):

PR #1662, opened 2022-07-05
fix: allow quoting quoted messages

arnautov-anton:
### 🎯 Goal

Allow quoting quoted messages by removing part of the condition within the `MessageActionsBox` component.

image


szuperaz:
Will enable in Angular as well

petyosi:
I'm ok with this. Why still in draft?

petyosi (approved):

oliverlaz (approved):

PR #1666, opened 2022-07-08
fix: align user role checks with channel membership, deprecate isModerator, isAdmin, isOwner

rvanlaarhoven:
# Submit a pull request

### 🎯 Goal

I'm not entirely sure how this was/is supposed to work. But in my case the api/library does not provide a `role` inside the channel state membership. Instead, it receives an `is_moderator: bool` and a `channel_role: string`.

Screen Shot 2022-07-08 at 11 53 33

### 🛠 Implementation details

This pull requests extends the `isModerator` check with these two props to correctly ensure a user is a moderator of the given channel. I didn't want to break previous behavior, so I left the `role` check in there.

### DIFF
--- a/src/components/Message/hooks/useUserRole.ts
+++ b/src/components/Message/hooks/useUserRole.ts
@@ -22,7 +22,11 @@ export const useUserRole = <
    * `isAdmin` will be removed in future release. See `channelCapabilities`.
    */
   const isAdmin = client.user?.role === 'admin' || channel.state.membership.role === 'admin';
-  const isMyMessage = client.userID === message.user?.id;
+
+  /**
+   * @deprecated as it relies on `membership.role` check which is already deprecated and shouldn't be used anymore.
+   * `isOwner` will be removed in future release. See `channelCapabilities`.
+   */
   const isOwner = channel.state.membership.role === 'owner';
 
   /**
@@ -36,6 +40,7 @@ export const useUserRole = <
     channel.state.membership.is_moderator === true ||
     channel.state.membership.channel_role === 'channel_moderator';
 
+  const isMyMessage = client.userID === message.user?.id;
   const canEdit =
     (!onlySenderCanEdit && channelCapabilities['update-any-message']) ||
     (isMyMessage && channelCapabilities['update-own-message']);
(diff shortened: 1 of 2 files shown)


src/components/Message/hooks/useUserRole.ts
```diff
@@ -24,7 +24,9 @@ export const useUserRole = <
   const isModerator =
     client.user?.role === 'channel_moderator' ||
     channel.state.membership.role === 'channel_moderator' ||
-    channel.state.membership.role === 'moderator';
+    channel.state.membership.role === 'moderator' ||
+    channel.state.membership.is_moderator === true ||
+    channel.state.membership.channel_role === 'channel_moderator';
```
AnatolyRugalev, line 29: While this is true, there are multiple things to keep in  mind:

1. `role` is deprecated and we should opt out from using it
2. When v2 of permissions is used, custom roles may be assigned to channel role, making these role checks fragile
3. When v2 of permissions is used, any extra moderator-level role can be added, so it's not binary

I'm OK with this change as long as we put a deprecation notice for this const. We should opt out from any role comparisons in the SDK and rely on provided capabilities instead.

@oliverlaz @petyosi , how is `isModerator` used in core SDK?

@rvanlaarhoven how do you use `isModerator` check? Do you use v2 of permissions?

Thanks

rvanlaarhoven: Yes, I'm using the v2 permissions.

I'd be happy to update the pull request. I left the `role` in there for backwards compatibility. Just let me know how to address the issue properly to match all the requirements to indicate a user as moderator. 

But since you say these role checks are fragile because of the ability to assign custom roles, it might make sense to deprecate `isModerator` completely to avoid confusion. I had to dig into the source to see what's going on and probably other people reading about the availability of `isModerator` will too.

oliverlaz: hi @rvanlaarhoven, In order to speed up the process, I added some deprecation comments to this PR as discussed for both `isModerator` and `isAdmin` properties (both of them rely on the already deprecated `membership.role` property). With this, we are ready to merge this PR.

However, we would like to hear more about your use case to use `isModerator` property. Is there something that can't be achieved with `channelCapabilities`?

oliverlaz: hi @rvanlaarhoven, thanks for sharing the info. `channelCapabilities` is the way to go for sure. I'm happy you can achieve the desired behavior with it.

oliverlaz (approved):

PR #1670, opened 2022-07-13
feat(message-status): extendable user information in read-by tooltip

jfacoustic:
# Override read-by tooltip format

### 🎯 Goal

Allow the user to customize read-by tooltips based on user information.

I needed this in my project so that I could add information about users from another database to the readby tooltip.  Due to our architecture, user.name was email, and we wanted it to be first/last name from our own users table.

### 🛠 Implementation details

Adds an optional callback to MessageStatus that takes in a UserResponse and and returns a string of the custom tooltip text.  This is passed into getReadByTooltipText.  If it's null, it uses the original format.

I also added some testing around getReadByTooltipText, including a minimal test helper function to enable testing with i18n. 

### 🎨 UI Changes

Screen Shot 2022-07-12 at 4 37 18 PM

### DIFF
--- a/src/components/Message/MessageStatus.tsx
+++ b/src/components/Message/MessageStatus.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { DeliveredCheckIcon } from './icons';
-import { getReadByTooltipText } from './utils';
+import { getReadByTooltipText, ReadByToolTipFormat } from './utils';
 
 import { AvatarProps, Avatar as DefaultAvatar } from '../Avatar';
 import { LoadingIndicator } from '../Loading';
@@ -17,6 +17,7 @@ import type { DefaultStreamChatGenerics } from '../../types/types';
 export type MessageStatusProps = {
   Avatar?: React.ComponentType<AvatarProps>;
   messageType?: string;
+  readByToolTipFormat?: ReadByToolTipFormat;
 };
 
 const UnMemoizedMessageStatus = <
@@ -24,7 +25,7 @@ const UnMemoizedMessageStatus = <
 >(
   props: MessageStatusProps,
 ) => {
-  const { Avatar: propAvatar, messageType = 'simple' } = props;
+  const { Avatar: propAvatar, messageType = 'simple', readByToolTipFormat } = props;
 
   const { client } = useChatContext<StreamChatGenerics>('MessageStatus');
   const { Avatar: contextAvatar } = useComponentContext<StreamChatGenerics>('MessageStatus');
@@ -65,7 +66,7 @@ const UnMemoizedMessageStatus = <
         className={`str-chat__message-${messageType}-status`}
         data-testid='message-status-read-by'
       >
-        <Tooltip>{getReadByTooltipText(readBy, t, client)}</Tooltip>
+        <Tooltip>{getReadByTooltipText(readBy, t, client, readByToolTipFormat)}</Tooltip>
         <Avatar
           image={lastReadUser.image}
           name={lastReadUser.name || lastReadUser.id}
--- a/src/components/Message/MessageStatus.tsx
+++ b/src/components/Message/MessageStatus.tsx
@@ -13,11 +13,12 @@ import { useMessageContext } from '../../context/MessageContext';
 import { useTranslationContext } from '../../context/TranslationContext';
 
 import type { DefaultStreamChatGenerics } from '../../types/types';
+import type { UserResponse } from 'stream-chat';
 
 export type MessageStatusProps = {
   Avatar?: React.ComponentType<AvatarProps>;
   messageType?: string;
-  readByToolTipFormat?: TooltipUsernameMapper;
+  tooltipUserNameMapper?: TooltipUsernameMapper;
 };
 
 const UnMemoizedMessageStatus = <
@@ -25,7 +26,11 @@ const UnMemoizedMessageStatus = <
 >(
   props: MessageStatusProps,
 ) => {
-  const { Avatar: propAvatar, messageType = 'simple', readByToolTipFormat } = props;
+  const {
+    Avatar: propAvatar,
+    messageType = 'simple',
+    tooltipUserNameMapper = (user: UserResponse<StreamChatGenerics>) => user.name || user.id,
+  } = props;
 
   const { client } = useChatContext<StreamChatGenerics>('MessageStatus');
   const { Avatar: contextAvatar } = useComponentContext<StreamChatGenerics>('MessageStatus');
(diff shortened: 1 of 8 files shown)


jfacoustic:
@oliverlaz On it!  Should have it sometime today.

src/components/Message/utils.tsx
```diff
@@ -334,23 +334,32 @@ export const getNonImageAttachments = <
...
     throw new Error('`getReadByTooltipText was called, but translation function is not available`');
   }
 
+  const defaultFormat = (user: UserResponse<StreamChatGenerics>) => user.name || user.id;
+
   // first filter out client user, so restLength won't count it
   const otherUsers = users
     .filter((item) => item && client?.user && item.id !== client.user.id)
-    .map((item) => item.name || item.id);
+    .map(readByToolTipFormat || defaultFormat);
```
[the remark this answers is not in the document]
jfacoustic: How about tooltipUsernameOverride?

src/components/Message/MessageStatus.tsx
```diff
@@ -17,14 +17,15 @@ import type { DefaultStreamChatGenerics } from '../../types/types';
 export type MessageStatusProps = {
   Avatar?: React.ComponentType<AvatarProps>;
   messageType?: string;
+  readByToolTipFormat?: ReadByToolTipFormat;
```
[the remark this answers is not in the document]
jfacoustic: Sounds good, didn't see that there!  Thanks

src/components/Message/utils.tsx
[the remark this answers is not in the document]
oliverlaz: I think `tooltipUsernameOverride` can lead to wrong initial assumption that the actual tooltip value comes, as a value, from outside. This line here actually extracts the name of the user hence, the name proposal. On second thought, maybe we can go with `<MessageStatus tooltipUserNameMapper={(user) => myMapper(user)} />`

What do you think about that?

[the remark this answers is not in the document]
jfacoustic: Works for me

jfacoustic:
@oliverlaz Got it, let me know if you need anything else!

src/components/Message/__tests__/utils.test.js
```diff
@@ -324,4 +326,54 @@ describe('Message utils', () => {
...
+      ];
+      const result = getReadByTooltipText(users, mockTranslatorFunction, client);
+      expect(result).toStrictEqual(`1, 2, and 3`);
+    });
+    it('returns "1, 2, 3, 4, 5, and 5 more if read by ten users', () => {
+      const users = [...Array(10).keys()].map((n) => generateUser({ name: (n + 1).toString() }));
+      const result = getReadByTooltipText(users, mockTranslatorFunction, client);
+      expect(result).toStrictEqual(`1, 2, 3, 4, 5, and 5 more`);
+    });
+    it('overrides user format with readByToolTipFormat', () => {
```
oliverlaz, line 363:
```suggestion
    it('overrides user format with tooltipUserNameMapper', () => {
```

oliverlaz: With my suggestions, the mapper function isn't optional anymore and it has to be provided as an argument to `getReadByTooltiptText` in these tests as well. Can you please take care of that?

oliverlaz (changes requested):
Hi @jfacoustic, I've added a couple of suggestions as part of my last review. We want to make the behavior a bit more obvious. Can you please have a look at them?

Also, it would be very good if you could update the documentation with the information about this new prop. Something along the lines of:
```
Type:
tooltipUserNameMapper: (user: UserResponse<StreamChatGenerics>) => string

Description:
When provided, it allows customization of the user information in the read-by status tooltip. By default, it will use the name of the user.
```

src/components/Message/MessageStatus.tsx
```diff
@@ -17,14 +17,15 @@ import type { DefaultStreamChatGenerics } from '../../types/types';
...
+  readByToolTipFormat?: TooltipUsernameMapper;
 };
 
 const UnMemoizedMessageStatus = <
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
 >(
   props: MessageStatusProps,
 ) => {
-  const { Avatar: propAvatar, messageType = 'simple' } = props;
+  const { Avatar: propAvatar, messageType = 'simple', readByToolTipFormat } = props;
```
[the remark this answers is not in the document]
jfacoustic: Why would you prefer that to be a required parameter rather than optional?  Sorry for the late reply, last week was hectic for me.

src/components/Message/__tests__/utils.test.js
jfacoustic: Why would you prefer that to be a required parameter rather than optional? Sorry for the late reply, last week was hectic for me.

jfacoustic: Also, where should I add the docs for the MessageStatus?  Should I create a new mdx file?   What's the difference between `docusaurus/docs/React/customization/message-ui.mdx` vs `docusaurus/docs/React/message-components/message-ui.mdx`?

oliverlaz: Hey @jfacoustic, no problem at all.

> Why would you prefer that to be a required parameter rather than optional?


`getReadByTooltipText` is internal API. If we have a stricter contract on it, we reduce the probability of future bugs. On the other hand, the MessageStatus API is public and there we are totally ok to have an optional parameter, as long as we are providing a default value for it.
For the tests, we can also use the default implementation. It can be exported from `utils.ts`.

oliverlaz: > Also, where should I add the docs for the MessageStatus? Should I create a new mdx file? What's the difference between `docusaurus/docs/React/customization/message-ui.mdx` vs `docusaurus/docs/React/message-components/message-ui.mdx`?

Please add the docs [here](https://github.com/GetStream/stream-chat-react/blob/develop/docusaurus/docs/React/message-components/ui-components.mdx#messagestatus-props).

`customization/message-ui.mdx` serves as an example of how a custom message can be implemented. The API docs live in the other file.

jfacoustic: @oliverlaz Added the docs and fixed the code comments!  Let me know if it needs anything else.

oliverlaz (approved):

jfacoustic:
@oliverlaz thanks so much!

oliverlaz (approved):

MartinCupela (approved):

oliverlaz:
Hi @jfacoustic, this PR is now merged and awaiting to be officially released. I hope you don't mind the small adjustments that I and @MartinCupela did. 
Many thanks for your contribution!

jfacoustic:
I'm onboard with that, that is much clearer.

PR #1697, opened 2022-08-01
ThemingV2: user testing and adjustments

arnautov-anton:
### 🎯 Goal

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_

### DIFF
--- a/src/components/Attachment/AttachmentContainer.tsx
+++ b/src/components/Attachment/AttachmentContainer.tsx
@@ -36,13 +36,12 @@ export const AttachmentWithinContainer = <
         : '';
   }
 
-  const classNames = clsx(
-    'str-chat__message-attachment',
-    componentType && `str-chat__message-attachment--${componentType}`,
-    attachment?.type && `str-chat__message-attachment--${attachment?.type}`,
-    componentType && extra && `str-chat__message-attachment--${componentType}--${extra}`,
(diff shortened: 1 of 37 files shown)


src/components/Attachment/AttachmentContainer.tsx
```diff
@@ -43,16 +42,7 @@ export const AttachmentWithinContainer = <
...
-    <div
-      className={classNames}
-      key={`${isGAT ? '' : attachment?.id || attachment?.fallback || nanoid()}-${
-        attachment?.type || 'none'
-      }`}
-    >
-      {children}
-    </div>
-  );
+  return <div className={classNames}>{children}</div>;
```
petyosi, line 45: I am fairly certain that those keys are not necessary here, as there's no loop.

src/components/MessageList/hooks/useScrollLocationLogic.tsx
```diff
@@ -57,7 +58,14 @@ export const useScrollLocationLogic = <
 
   useEffect(() => {
     if (!listElement) return;
-    const observer = new ResizeObserver(scrollToBottom);
```
petyosi, line 60: @MartinCupela the resizeobserver was very trigger happy, throwing events even when the height was not changed (for example, when gallery was open).

PR #1698, opened 2022-08-02
fix: detect mentions of users who have email as their name

oliverlaz:
### 🎯 Goal
With this PR, we want to fix one edge case with user mentions feaure in which a user could have their name set to their e-mail address. Before, the SDK would wrongly recognize the mention as an e-mail address and would wrap it into `<a href="mailto:..." />` tag.

### 🛠 Implementation details
The link detection process kicks in before user mentions are detected.
With this change, while detecting links we also consider whether the detected link matches an email identical to the mentioned user's name.

As part of this change, `mdast-util-find-and-replace` is updated to its latest version as the one we were previously using had a bug, and sometimes it could skip matching nodes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "lodash.isequal": "^4.5.0",
     "lodash.throttle": "^4.1.1",
     "lodash.uniqby": "^4.7.0",
-    "mdast-util-find-and-replace": "1.1.1",
+    "mdast-util-find-and-replace": "^2.2.1",
     "nanoid": "^3.3.4",
     "pretty-bytes": "^5.4.1",
     "prop-types": "^15.7.2",
@@ -103,6 +103,7 @@
     "@types/lodash.isequal": "^4.5.5",
     "@types/lodash.throttle": "^4.1.6",
     "@types/lodash.uniqby": "^4.7.6",
+    "@types/mdast": "^3.0.10",
     "@types/moment": "^2.13.0",
     "@types/react": "^18.0.8",
     "@types/react-dom": "^18.0.3",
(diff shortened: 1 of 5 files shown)


MartinCupela:
@oliverlaz the original request complained about the `options.customMarkDownRenderers` not having an impact on the render result. I am thinking, whether the solution should address that in order we avoid implementing code handling edge cases and allow our users to handle those edge cases by themselves. How do you see that?

petyosi (approved):

MartinCupela (approved):

petyosi (approved):

alexlopezit:
@oliverlaz @MartinCupela @petyosi Thanks for the update and new release version.

As the latest version of `mdast-util-find-and-replace` uses ESM I got a quick question; our tests are failing and I was just wondering if you know a way around this?

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/.../node_modules/mdast-util-find-and-replace/index.js from /Users/.../node_modules/stream-chat-react/dist/index.cjs.js not supported.
Instead change the require of index.js in /Users/.../node_modules/stream-chat-react/dist/index.cjs.js to a dynamic import() which is available in all CommonJS modules.
```

oliverlaz:
> @oliverlaz @MartinCupela @petyosi Thanks for the update and new release version.
> 
> As the latest version of `mdast-util-find-and-replace` uses ESM I got a quick question; our tests are failing and I was just wondering if you know a way around this?
> 
> ```
> Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/.../node_modules/mdast-util-find-and-replace/index.js from /Users/.../node_modules/stream-chat-react/dist/index.cjs.js not supported.
> Instead change the require of index.js in /Users/.../node_modules/stream-chat-react/dist/index.cjs.js to a dynamic import() which is available in all CommonJS modules.
> ```

hi @alexlopezit, thanks for reporting the issue. I'll have a look and I'll get back to you.

oliverlaz:
hi @alexlopezit, we just released v9.4.1 which fixes the problem you faced. Can you please upgrade and let us know whether everything runs well on your end?

alexlopezit:
Thanks @oliverlaz, works perfect now 👍

PR #1713, opened 2022-08-16
Theming v2 user testing fix scroll to bottom in non-virtualized message list

MartinCupela:
### 🎯 Goal

Prevent non-virtualized message list from scrolling to the bottom while scrolling up and loading more messages.

### 🛠 Implementation details

The reason, why the list kept scrolling to the bottom when loading more messages was that the `listElement.scrollBy(scrollHeightDelta)` was called even when the container was automatically scrolled by the browser. The `scrollBy` call had sense only when the container is scrolled to the top (`scrollTop === 0`). Otherwise the browser handles the increase of the `scrollHeight`.

Once the condition was put in place, the ResizeObserver was no more needed.

I have also removed the requirement that the container is scrolled to the bottom on asset load only if it is near the bottom but not near the top. The latter condition (not near the top) was breaking the scroll behavior when few messages and an image were present.


src/components/MessageList/hooks/useScrollLocationLogic.tsx
```diff
@@ -127,7 +85,7 @@ export const useScrollLocationLogic = <
       const prevCloseToBottom = closeToBottom.current;
       closeToBottom.current = scrollHeight - (scrollTop + offsetHeight) < scrolledUpThreshold;
       closeToTop.current = scrollTop < scrolledUpThreshold;
-
+      console.log('scrollTop, scrolledUpThreshold', scrollTop, scrolledUpThreshold);
```
petyosi, line 88: This should be cleaned up?

petyosi (approved):

src/components/MessageList/hooks/useScrollLocationLogic.tsx
MartinCupela: @petyosi thank you, removed

PR #1714, opened 2022-08-17
feat(ThemingV2): PopperTooltip component

arnautov-anton:
### 🎯 Goal

Replace current tooltip components with `PopperTooltip` component which utilises package [`PopperJS`](https://popper.js.org/react-popper/v2/). This PR relies on [CSS updates](https://github.com/GetStream/stream-chat-css/pull/165).

### 🎨 UI Changes
image
image
(viewport-aware, adjusts position by itself for better visibility)
image

### DIFF
--- a/src/components/Reactions/hooks/useProcessReactions.tsx
+++ b/src/components/Reactions/hooks/useProcessReactions.tsx
@@ -86,8 +86,21 @@ export const useProcessReactions = <
     [reactionCounts, supportedReactionsArePresent],
   );
 
+  const aggregatedNamesByType = useMemo(
+    () =>
+      latestReactions.reduce<Record<string, Array<string>>>((typeMap, { type, user }) => {
+        typeMap[type] ??= [];
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        typeMap[type].push(user?.name || user!.id);
+        return typeMap;
+      }, {}),
+    [latestReactions],
+  );
+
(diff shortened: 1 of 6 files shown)


oliverlaz (approved):

src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -86,8 +86,21 @@ export const useProcessReactions = <
     [reactionCounts, supportedReactionsArePresent],
   );
 
+  const aggregatedNamesByType = useMemo(
+    () =>
+      latestReactions.reduce<Record<string, Array<string>>>((pv, { type, user }) => {
+        pv[type] ??= [];
```
MartinCupela, line 92: @arnautov-anton just nitpicking but what `pv` stands for?

MartinCupela (approved):

src/components/Reactions/hooks/useProcessReactions.tsx
arnautov-anton: `pv` as in "previous value". It's something I've started with and it stuck with me. I like it more than `acc` which in this case is more suitable but never sat right with me. Might start using `av` as in "accumulated value". :smile:

petyosi: I was also tripped by the variable naming here.

petyosi (approved):

PR #1716, opened 2022-08-18
feat: render MessageActionsBox on modal

MartinCupela:
### 🎯 Goal

Render the message actions menu above an overlay that would prevent the erratic behavior of menu disappearing when hovered away. Make sure that only one menu is rendered in a given container.

### 🛠 Implementation details

This implementation uses React portal to append the modal to a selected container. That avoids the message hover state to influence the display of the message actions menu.

Only one message actions menu can be rendered per message list.

Each menu is positioned using `react-popper`. This assures that the menu position will be adjusted so that it is always entirely visible.

Message options are kept visible meanwhile the message actions menu is open.

During the implementation the `Modal` component received changes that allows us to pass the properties to the inner modal container `div` as well as to set the reference to this `div` for the parent components. This is used in connection with the `react-popper`. 

### 🎨 UI Changes

[demo_message_actions_on_modal.webm](https://user-images.githubusercontent.com/32706194/185423321-57caea77-fc39-4e35-9606-7e25a9ad4a1a.webm)


DannyBiezen:
I've been trying to find a way to not have the message actions menu disappear when hovering away and found this PR which seems to do exactly what I want!

Are there any plans on getting this merged? It would be great having this provided by this SDK instead of having to customize a lot of components in my codebase.

MartinCupela:
Hello @DannyBiezen , this is currently on hold due the focus on other priorities. I plan to get back to this, but I do not have an exact roadmap to provide you.

MartinCupela:
fixed by #2489

PR #1721, opened 2022-08-22
fix(ThemingV2): MessageInputFlat missing lodash/zipObject

arnautov-anton:
### 🎯 Goal

Node allows you to import packages that are not specified in the `package.json` but essentially are in the closest `node_modules` folder. Due to this "feature" I was able to import and use `lodash/zipObject` which is not available in the final bundle thus making it "broken".  

### 🛠 Implementation details

Based on our conversation with @oliverlaz I've added a [`import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md) `eslint` rule to prevent such mistakes from happening in the future.


oliverlaz (approved):

src/components/MessageInput/MessageInputFlat.tsx
```diff
@@ -214,10 +213,10 @@ const MessageInputV2 = <
 
   const accept = useMemo(
     () =>
-      zipObject(
-        acceptedFiles,
-        Array.from({ length: acceptedFiles.length }, () => []),
-      ),
+      acceptedFiles.reduce<Record<string, Array<string>>>((av, cv) => {
```
petyosi, line 216: TIL about `??=` ;). Apart from that, I feel like the variable naming for `av` and `cv` can be improved to something that requires less mental overhead.

petyosi (approved):

petyosi (approved):

PR #1727, opened 2022-08-29
ThemingV2.3 beta release 

arnautov-anton:


src/components/MessageInput/__tests__/AttachmentPreviewList.test.js
```diff
@@ -17,6 +16,13 @@ const uploadsReducer = (pv, cv) => {
   return pv;
 };
 
+/**
+ * @param {string} word
+ * @returns string
+ */
+const capitalize = ([firstLetter, ...restOfTheWord]) =>
+  `${firstLetter.toUpperCase()}${restOfTheWord.join('')}`;
```
MartinCupela, line 24: @arnautov-anton this implementation does not follow the lodash definition:

"Converts the first character of string to upper case and the remaining to lower case."

[lodash/capitalize](https://lodash.com/docs/4.17.15#capitalize)

If the `restOfTheWOrd` contains uppercase letters, then those would be kept.

src/components/Reactions/ReactionsList.tsx
```diff
@@ -32,6 +36,40 @@ export type ReactionsListProps<
   reverse?: boolean;
 };
 
+const ButtonWithTooltip = ({
+  children,
+  onMouseEnter,
```
MartinCupela, line 41: How will users be able to pass a custom callback, if it is not forwarded inside `UnMemoizedReactionsList`?

src/components/MessageInput/__tests__/AttachmentPreviewList.test.js
arnautov-anton: This is just a utility function that has to capitalize the first letter of the word just the way it's designed, it doesn't need (nor is meant) to follow full `lodash/capitalize` implementation.

src/components/Reactions/ReactionsList.tsx
arnautov-anton: Fair point but I'd rather have this button fully replaceable (passing down a custom component) instead of just passing down these two handle functions.

src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -86,8 +86,21 @@ export const useProcessReactions = <
     [reactionCounts, supportedReactionsArePresent],
   );
 
+  const aggregatedNamesByType = useMemo(
+    () =>
+      latestReactions.reduce<Record<string, Array<string>>>((typeMap, { type, user }) => {
+        typeMap[type] ??= [];
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        typeMap[type].push(user?.name || user!.id);
```
MartinCupela, line 95: Can it happen that the user will be undefined?

src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -86,8 +86,21 @@ export const useProcessReactions = <
     [reactionCounts, supportedReactionsArePresent],
   );
 
+  const aggregatedNamesByType = useMemo(
```
MartinCupela, line 89:
```suggestion
  const aggregatedUserNamesByType = useMemo(
```

src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -86,8 +86,21 @@ export const useProcessReactions = <
...
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        typeMap[type].push(user?.name || user!.id);
+        return typeMap;
+      }, {}),
+    [latestReactions],
+  );
+
   return {
     additionalEmojiProps: reactionsAreCustom ? additionalEmojiProps : emojiSetDef,
+    aggregatedNamesByType,
```
MartinCupela, line 103:
```suggestion
    aggregatedUserNamesByType,
```

src/components/Reactions/ReactionsList.tsx
```diff
@@ -44,6 +82,7 @@ const UnMemoizedReactionsList = <
 
   const {
     additionalEmojiProps,
+    aggregatedNamesByType,
```
MartinCupela, line 85:
```suggestion
    aggregatedUserNamesByType,
```

src/components/Reactions/ReactionsList.tsx
```diff
@@ -80,7 +119,11 @@ const UnMemoizedReactionsList = <
               })}
               key={emojiObject.id}
             >
-              <button aria-label={`Reactions: ${reactionType}`}>
+              <ButtonWithTooltip
+                aria-label={`Reactions: ${reactionType}`}
+                title={aggregatedNamesByType[reactionType].join(', ')}
```
MartinCupela, line 124:
```suggestion
                title={aggregatedUserNamesByType[reactionType].join(', ')}
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,9 +1,46 @@
...
+  });
+
+  if (!visible) return null;
 
-export const Tooltip = (props: ComponentProps<'div'>) => {
-  const { children, ...rest } = props;
   return (
-    <div className='str-chat__tooltip' {...rest}>
+    <div
+      className='str-chat__tooltip str-chat__button-tooltip'
```
MartinCupela, line 39: Will this tooltip always wrap buttons? Asking because of the class `str-chat__button-tooltip`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,9 +1,46 @@
...
+  </div>
+);
+
+export const PopperTooltip = <T extends HTMLElement>({
+  children,
+  offset = [0, 10],
+  referenceElement,
+  placement = 'top',
+  visible: visible = false,
+}: React.PropsWithChildren<{
```
MartinCupela, line 16: It would be great to document the props purpose with comments. Wouldn't it be more robust to keep them in a separate type?

src/components/Tooltip/hooks/useEnterLeaveHandlers.ts
```diff
@@ -0,0 +1,26 @@
+import React, { useCallback, useState } from 'react';
+
+export const useEnterLeaveHandlers = <T extends HTMLElement>({
+  onMouseEnter,
+  onMouseLeave,
+}: Partial<Record<'onMouseEnter' | 'onMouseLeave', React.MouseEventHandler<T>>> = {}) => {
+  const [popperVisible, setPopperVisible] = useState(false);
```
MartinCupela, line 7: I would not give names according to a library that can change over time. Could we instead use name like `tooltipVisible`?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,9 +1,46 @@
...
+        name: 'offset',
+        options: {
+          offset,
+        },
+      },
+    ],
+    placement,
+  });
+
+  if (!visible) return null;
```
MartinCupela, line 35: This may need a test.

src/components/Tooltip/Tooltip.tsx
arnautov-anton: This class isn't even defined within our styling so I'll remove it completely, must've missed it.

PR #1738, opened 2022-09-08
fix: avoid race condition crash in jumping

petyosi:
### 🎯 Goal

This change fixes a production-only issue where rapid jumping between messages created a race condition + exception. 
@szuperaz - more of an FYI if you copied this mistake from me :(.

### DIFF
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -524,6 +524,8 @@ const ChannelInner = <
     return queryResponse.messages.length;
   };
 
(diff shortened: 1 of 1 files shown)


oliverlaz (approved):

szuperaz:
I did! But how was this a production-only issue?

szuperaz (approved):

arnautov-anton (approved):

petyosi:
> I did! But how was this a production-only issue?

@szuperaz I could not fully confirm that, but my guess is that this was somehow mitigated by React running effects twice in development :(.

Issue #1743, opened 2022-09-12
bug: Newest release does not disclose usage of React 18 features (useId hook)


tejashah88:
**Describe the bug**

I tried upgrading to the newest release (v10.0.0) and I got the following bug shown below. I'm using React v17 and after a bit of research, I found that `useId` is a hook [introduced in React 18](https://reactjs.org/blog/2022/03/29/react-v18.html#useid). Since the React version upgrade wasn't explicit in the changelog, this was surprising to say the least.

**To Reproduce**

1. `npm install --save stream-chat-react@10.0.0`
2. `npm start` or `npm run build`
3. See error above

**Expected behavior**

Either the changelog should reflect that React 18 is a minimum requirement or `useId` should be replaced in favor of an alternative solution while keeping React 17 compatibility.

**Screenshots**

<img width="632" alt="image" src="https://user-images.githubusercontent.com/11187343/189576112-552076ff-5b33-415e-91fe-0e97e14e214e.png">

**Package version**
- stream-chat-react: `10.0.0`
- stream-chat-css: `3.0.0`
- stream-chat-js: `7.1.0`

petyosi:
Thanks for reporting this, we will look into removing the usage as soon as possible.

PR #1744, opened 2022-09-12
fix: remove `useId` from `LoadingIndicatorIcon`

arnautov-anton:
### 🎯 Goal

Remove React's `useId` which breaks compatibility for older versions of React and replace it with memoized `nanoid` "equivalent".


petyosi (approved):

oliverlaz:
Fixes: #1743

PR #1761, opened 2022-09-19
fix(VirtualizedMessageList): use memoized values as hook dependencies directly

MartinCupela:
### 🎯 Goal

Avoid unnecessary re-renders caused by regenerating a new array of message group style keys (`Object.keys()` was called directly in the hook dependency array).

### DIFF
--- a/src/components/MessageList/VirtualizedMessageList.tsx
+++ b/src/components/MessageList/VirtualizedMessageList.tsx
@@ -206,6 +206,7 @@ const VirtualizedMessageListWithContext = <
         if (style) acc[message.id] = style;
         return acc;
       }, {} as Record<string, GroupStyle>),
+    // processedMessages were incorrectly rebuilt with a new object identity at some point, hence the .length usage
     [processedMessages.length, shouldGroupByUser],
   );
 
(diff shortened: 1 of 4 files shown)


oliverlaz (approved):

petyosi (changes requested):
I believe that we had a bug where `processedMessages` were incorrectly rebuilt with a new object identity at some point, hence the `.length` usage (which, of course, is not correct but we are not aware of any issues because of it). 
Given that, I would say that this change should not touch on that.

MartinCupela:
> I believe that we had a bug where `processedMessages` were incorrectly rebuilt with a new object identity at some point, hence the `.length` usage (which, of course, is not correct but we are not aware of any issues because of it). Given that, I would say that this change should not touch on that.

@petyosi I have reverted the changes and added the explanatory comment for future reference.

petyosi (approved):

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -206,6 +206,7 @@ const VirtualizedMessageListWithContext = <
         if (style) acc[message.id] = style;
         return acc;
       }, {} as Record<string, GroupStyle>),
```
arnautov-anton, line 208:
```suggestion
      }, {}),
```
Not part of the PR so feel free to ignore this comment but noticed this casting which could be refactored to this:
```ts
array.reduce<Record<string, GroupStyle>>(...)
```

arnautov-anton (approved):

src/components/MessageList/VirtualizedMessageList.tsx
MartinCupela: @arnautov-anton added generic type def to all reduce calls

arnautov-anton (approved):

PR #1786, opened 2022-09-27
fix: add image attachment height from CSS

szuperaz:
# Submit a pull request

### 🎯 Goal

Thet attachment sizing didn't work with the virtualized message list because the virtualized message list uses a resize observer and that can be called after the initial attachment rendering but before the `useLayoutEffect` that set the height for the attachment -> this caused scrolling issues. @petyosi created a fix where the height calculation is done in CSS (https://github.com/GetStream/stream-chat-css/pull/191) ensuring that we set the height "soon enough" for the virtualized message list.

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_


petyosi (approved):

src/components/Attachment/attachment-sizing.tsx
```diff
@@ -65,29 +56,11 @@ const getSizingRestrictions = (url: URL, htmlElement: HTMLElement) => {
...
-              )
-          : cssSizeRestriction.maxHeight;
-      height = `${heightNum}px`;
-    }
   } else {
     resizeDimensions = undefined;
   }
 
-  return { height, resizeDimensions };
+  return { resizeDimensions };
```
MartinCupela, line 63:
```suggestion
  return resizeDimensions;
```

src/components/Attachment/attachment-sizing.tsx
```diff
@@ -1,10 +1,8 @@
 import type { Attachment } from 'stream-chat';
 
 export const getImageAttachmentConfiguration = (attachment: Attachment, element: HTMLElement) => {
-  const url = new URL(
-    (attachment.img_url || attachment.image_url || attachment.thumb_url || '') as string,
-  );
-  const { height, resizeDimensions } = getSizingRestrictions(url, element);
+  const url = new URL((attachment.image_url || attachment.thumb_url || '') as string);
+  const { resizeDimensions } = getSizingRestrictions(url, element);
```
MartinCupela, line 5:
```suggestion
  const resizeDimensions = getSizingRestrictions(url, element);
```

PR #1790, opened 2022-09-28
docs: add guide to integrate 100ms video calls

DaemonLoki:
# Submit a pull request

### 🎯 Goal

Add a guide how to integrate the 100ms SDK to implement video calls.

### 🛠 Implementation details

Docs only. The repository with the code can be found [here](https://github.com/GetStream/react-video-integration-100ms) (it's private, ask @DaemonLoki to get access)

### 🎨 UI Changes

No UI changes.


docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+}: { children: ReactNode } => {
+  const [connectionState, setConnectionState] = useState(ConnectionState.NoCall);
+
+  const store = {
+    connectionState: connectionState,
+    createCall: () => {},
+    joinCall: () => {},
+    leaveCall: () => {},
+    endCall: () => {},
+  };
```
arnautov-anton, line 327:
```suggestion
  const store = useMemo(() => ({
    connectionState,
    createCall: () => {},
    joinCall: () => {},
    leaveCall: () => {},
    endCall: () => {},
  }), [connectionState]);
```
The reference to the `store` object and to each of the call functions would change each time this provider function runs essentially triggering updates in places which rely on it. I'd go as far as pulling each of the <create/join/leave/end>call functions to their own definitions (if they rely on some "contextual value" of the provider function then define them within provider function wrapped in useCallback but if they don't then I'd pull them out of the provider completely to keep them stable) and then passing them down to the store object adding them also to the dependency array.

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+const [connectionState, setConnectionState] = useState(
+  channel.data?.data?.callActive ? ConnectionState.CallAvailable : ConnectionState.NoCall,
+);
+```
+
+With that, we can add a `useEffect` hook to register for `channel.updated` events. Add the code below the `connectionState` creation:
+
+```tsx
+useEffect(() => {
+  const handleChannelUpdate = (event: any) => {
```
arnautov-anton, line 403:
```suggestion
  const handleChannelUpdate = (event: Event) => {
```

We have a type that comes from the `stream-chat` packate that defines the event object but the `any` might have been intentional so feel free to ignore this comment. :)

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+    if (event.channel.data?.callActive) {
+      if (connectionState == ConnectionState.NoCall) {
+        setConnectionState(ConnectionState.CallAvailable);
+      } else if (connectionState == ConnectionState.Connecting) {
```
arnautov-anton, line 407:
```suggestion
      if (connectionState === ConnectionState.NoCall) {
        setConnectionState(ConnectionState.CallAvailable);
      } else if (connectionState === ConnectionState.Connecting) {
```

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+
+```tsx
+joinCall: async () => {
+  setConnectionState(ConnectionState.Connecting);
+
+  const response = await client.getCallToken(channel.data?.data?.callId);
+
+  await hmsActions.join({
+    authToken: `${response.token}`,
+    userName: client.user?.name ?? 'Unkown',
```
arnautov-anton, line 501:
```suggestion
    userName: client.user?.name || 'Unkown',
```
This is a nitpick - is there a possibility that the username could be an empty string?

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+```
+
+With that we have all the logic available, there is just one more thing we’ll add. In order for us to not be stuck in a call when the window is closed, we’ll also leave the call when the `window.onunload` event is happening:
+
+```tsx
+useEffect(() => {
+  window.onunload = () => {
+    hmsActions.leave();
+  };
+}, []);
```
arnautov-anton, line 547:
```suggestion
useEffect(() => () => {  
    hmsActions.leave();
}, []);
```
We can utilise React's cleaning function which gets invoked whenever either dependency changes or the component itself unmounts.

arnautov-anton: Now also thinking about `hmsActions` - is that object stable? So that when the cleaning function gets invoked it doesn't call `leave` on a "stale" `hmsActions` reference.

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+          <button className='call-button leave-call-button' onClick={onVideoButtonClick}>
+            <p>{buttonText()}</p>
+          </button>
+          <button className='call-button end-call-button' onClick={endCall}>
+            <p>End call</p>
+          </button>
+        </>
+      );
+  }
+};
```
arnautov-anton, line 658: I know this works and looks fine but by React standards it's an antipattern - meaning it's a function that returns React nodes so it as well could be a proper component.

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+import Peer from './Peer';
+
+import './VideoGrid.css';
+
+const VideoGrid = () => {
+  const { connectionState } = useVideoContext();
+  const peers = useHMSStore(selectPeers);
+  return (
+    <>
+      {connectionState == ConnectionState.InCall && (
```
arnautov-anton, line 814:
```suggestion
      {connectionState === ConnectionState.InCall && (
```

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
```diff
@@ -0,0 +1,881 @@
...
+    case ConnectionState.CallAvailable:
+      return 'Join call';
+    case ConnectionState.Connecting:
+      return 'Connecting';
+    case ConnectionState.Disconnecting:
+      return 'Disconnecting';
+    case ConnectionState.InCall:
+      return 'Leave Call';
+  }
+};
```
arnautov-anton, line 673:
```suggestion
const buttonText = (connectionState: ConnectionState) => {
  switch (connectionState) {
    case ConnectionState.NoCall:
      return 'Create call';
    case ConnectionState.CallAvailable:
      return 'Join call';
    case ConnectionState.Connecting:
      return 'Connecting';
    case ConnectionState.Disconnecting:
      return 'Disconnecting';
    case ConnectionState.InCall:
      return 'Leave Call';
  }
};
```
This function then can live outside the component scope to keep its reference stable.

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
DaemonLoki: What would you consider the best way to do this then?

I feel like it's the best way to switch between the different states the enum can have and show UI accordingly. I know there's a way to do it [inside the DOM](https://stackoverflow.com/a/51432223) but that looks/feels even more ugly.

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
DaemonLoki: It could probably, potentially be an empty string and in that case it would use the `'Unknown'` one which I feel like is a tad nicer, than showing an empty string here. :) 

So, will change it. Good catch!

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
DaemonLoki: Object should be stable. That is a weird syntax returning from useEffect with only a function but makes sense. :)

docusaurus/docs/React/custom-code-examples/video-integration-100ms.mdx
DaemonLoki: Good catch! Refactored it!

arnautov-anton (approved):

Issue #1813, opened 2022-10-13
stream-chat-react 10.3.0 incorrect dependency resolution with stream-chat 8.0.0


chinesehemp:
**Describe the bug**

I've upgraded  `stream-chat-react` to version 10.3.0 and also upgraded `stream-chat` to version 8.0.0, but when I try to deploy my app, I get npm dependency tree error.

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: my-app@0.1.0
npm ERR! Found: stream-chat@8.0.0
npm ERR! node_modules/stream-chat
npm ERR!   stream-chat@"^8.0.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer stream-chat@"^7.1.0" from stream-chat-react@10.3.0
npm ERR! node_modules/stream-chat-react
npm ERR!   stream-chat-react@"^10.3.0" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

**To Reproduce**

Steps to reproduce the behaviour:
1. Install `stream-chat` version 8.0.0
2. Install `stream-chat-react` version 10.3.0
3. Run npm install
**Expected behavior**

No dependency tree error.

**Screenshots**

<img width="670" alt="Screenshot 2022-10-13 at 11 11 43" src="https://user-images.githubusercontent.com/58605231/195569662-66c63d91-8ce4-456e-8b84-a253e2da4b01.png">

**Package version**
- stream-chat-react: 10.3.0
- stream-chat: 8.0.0

oliverlaz:
Hi @chinesehemp, `stream-chat-react@10.3.0` requires `stream-chat@^7.1.0`. From semver perspective, `stream-chat@8.0.0` doesn't fulfill the `stream-chat-react` requirements hence the error you see.

We'll soon release a new React SDK version compatible with the `stream-chat@^8.0.0`.
Until then, we suggest you downgrade to `stream-chat@7.2.0`.

MartinCupela:
@chinesehemp new version 10.3.1 of stream-chat-react is available. You should be now able to install stream-chat@8.0.0 with it.

chinesehemp:
Thank you.

PR #1814, opened 2022-10-13
fix: add stream-chat@8.x.x to peerDependencies

MartinCupela:


package.json
```diff
@@ -64,7 +64,7 @@
   "peerDependencies": {
     "react": "^18.0.0 || ^17.0.0 || ^16.8.0",
     "react-dom": "^18.0.0 || ^17.0.0 || ^16.8.0",
-    "stream-chat": "^7.1.0"
+    "stream-chat": "^8.0.0 || ^7.1.0"
```
arnautov-anton, line 67:
```suggestion
    "stream-chat": ">= 7.1.0"
```

arnautov-anton: Or whatever oldest version we decide to support.

oliverlaz: I think we should stay in the range that @MartinCupela proposed for the time being.
I'm not sure how safe would it be if we have, for example, `chat@10.4` and `client@12`.

oliverlaz (approved):

oliverlaz:
Fixes: #1813

PR #1874, opened 2022-12-13
customSearchFunction example in docs is not correct

GuidovdRiet:
### 🎯 Goal

Improve the customSearchFunction example in [docs](https://getstream.io/chat/docs/sdk/react/guides/customization/channel_search/#the-searchfunction-prop) is not correct.

- Use of hook (`useChatContext`) within regular async function (no hook or component).
- `chatClient` is not defined.
- setResultsOpen doesn't exist.

### 🛠 Implementation details

Errors fixed:

- React Hook "useChatContext" is called in function "customSearchFunction" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"

- `chatClient` doesn’t exist ➜ should be client.

- Provided `client` as prop instead of the use of a hook within a regular function. 

- Re-used `e.target.value` assigned to var ➜ improves re-usability.

### DIFF
--- a/docusaurus/docs/React/guides/customization/channel-search.mdx
+++ b/docusaurus/docs/React/guides/customization/channel-search.mdx
@@ -317,26 +317,37 @@ To override the search method, completely use the `searchFunction` prop. This pr
 and for only channels that the current logged in user is a member of. See the example below for this.
 
 ```jsx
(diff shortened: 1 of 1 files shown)


MartinCupela (approved):
@GuidovdRiet thank you for spotting this and creating the fix.

PR #1916, opened 2023-01-31
feat: introduce useCreateChatClient hook

arnautov-anton:
### 🎯 Goal

Introduce `useCreateChatClient` hook which takes care of creation and initiation of `StreamChat` client. 

closes: GetStream/stream-chat-react-issues#19

### DIFF
--- a/src/components/Chat/hooks/useCreateChatClient.ts
+++ b/src/components/Chat/hooks/useCreateChatClient.ts
@@ -1,0 +1,123 @@
+import { useEffect, useState } from 'react';
+
+import {
+  DefaultGenerics,
+  ExtendableGenerics,
+  OwnUserResponse,
+  StreamChat,
+  TokenOrProvider,
+  UserResponse,
+} from 'stream-chat';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const makeCancellable = <T extends (...functionArguments: any[]) => any>(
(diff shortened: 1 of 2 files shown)


src/components/Chat/hooks/useCreateChatClient.ts
```diff
@@ -0,0 +1,47 @@
...
+      didUserConnectInterrupt = true;
+      setChatClient(null);
+      connectionPromise
+        .then(() => client.disconnectUser())
+        .then(() => {
+          console.log(`Connection for user "${userData.id}" has been closed`);
+        });
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [apiKey, userData.id, tokenOrProvider]);
```
myandrienko, line 44: In case you want to handle this without ignoring `exhaustive-deps`:

```tsx
const [cachedUserData, setCachedUserData] = useState(userData);

if (userData.id !== cachedUserData.id) {
  setCachedUserData(userData);
}
```

And then you can safely add `cachedUserData` to the deps list.

arnautov-anton: This adds extra overhead but it makes sense so I'll add it, thank you! 👍

myandrienko (approved):

oliverlaz (approved):

PR #1952, opened 2023-02-24
feat(Attachments): introduce `UnsupportedAttachment` component

arnautov-anton:
### 🎯 Goal

Introduces `UnsupportedAttachment` component to handle unsupported attachment types.

### 🛠 Implementation details

This component is disabled by default, enable by passing `UnsupportedAttachment` component to `Attachment`'s `UnsupportedAttachment` property:

```tsx
import { Attachment, UnsupportedAttachment } from 'stream-chat-react';

const CustomAttachment = (props) => <Attachment {...props} UnsupportedAttachment={UnsupportedAttachment} />

/* ... */

<Channel Attachment={CustomAttachment} >
    /* ... */
</Channel>

```

### 🎨 UI Changes

image

### DIFF
--- a/src/components/Attachment/Attachment.tsx
+++ b/src/components/Attachment/Attachment.tsx
@@ -19,6 +19,7 @@ import {
   GalleryContainer,
   ImageContainer,
   MediaContainer,
+  UnsupportedAttachmentContainer,
 } from './AttachmentContainer';
 
 import type { AttachmentActionsProps } from './AttachmentActions';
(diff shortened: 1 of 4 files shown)


src/components/Attachment/Unsupported.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type { Attachment } from 'stream-chat';
+import type { DefaultStreamChatGenerics } from '../../types/types';
+
+export type UnsupportedProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = {
+  attachment: Attachment<StreamChatGenerics>;
+};
+
+export const Unsupported = <
```
oliverlaz, line 11: `import { Unsupported } from 'stream-chat-react'` doesn't say much.
Maybe we should include `Attachment` in the name of this component -> `UnsupportedAttachment` or `UnknownAttachment`

src/components/Attachment/Attachment.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable sort-keys */
```
oliverlaz, line 1: I guess this is dev-only leftover. Shall we remove it?

arnautov-anton: I moved it to affected lines, I'd love to disable the rule altogether though.

oliverlaz (approved):

PR #2004, opened 2023-05-04
fix: prevent duplicate simultaneous query channel quests

MartinCupela:
### 🎯 Goal

Prevent duplicate simultaneous requests querying channel data, if multiple events that trigger querying data for a channel are delivered to the client. Example of such events are `message.new` and `notification.added_to_channel`. They are received when user sends a message to a newly created channel (for example a direct message). The queries are triggered by the hooks inside the `ChannelList` component.

### 🛠 Implementation details

Keep a map of promises representing the query channel HTTP requests being in flight and prevent retriggering that request for a given channel and only await till the query finishes..

### DIFF
--- a/src/components/ChannelList/utils.ts
+++ b/src/components/ChannelList/utils.ts
@@ -1,8 +1,23 @@
-import type { Channel, StreamChat } from 'stream-chat';
+import type { Channel, QueryChannelAPIResponse, StreamChat } from 'stream-chat';
 import uniqBy from 'lodash.uniqby';
 
 import type { DefaultStreamChatGenerics } from '../../types/types';
 
+/**
+ * prevent from duplicate invocation of channel.watch()
+ * when events 'notification.message_new' and 'notification.added_to_channel' arrive at the same time
+ */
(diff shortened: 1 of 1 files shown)


oliverlaz (approved):

src/components/ChannelList/utils.ts
```diff
@@ -1,8 +1,23 @@
...
 import type { DefaultStreamChatGenerics } from '../../types/types';
 
+/**
+ * prevent from duplicate invocation of channel.watch()
+ * when events 'notification.message_new' and 'notification.added_to_channel' arrive at the same time
+ */
+const WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL: Record<
+  string,
+  Promise<QueryChannelAPIResponse> | undefined
+> = {};
```
arnautov-anton, line 13: This is a nitpick so you don't have to act on it (this is good as it is) but I believe using `Map` could make a code a bit nicer.

MartinCupela: I will leave this as is to save time retesting everything :)

PR #2007, opened 2023-05-05
feat: allow to retrieve the default message input value dynamically

MartinCupela:
### 🎯 Goal

Allow users to generate the default value for `MessageInput` components instead of providing static `defaultValue` through `additionalTextareaProps`. This circumvents the limitation posed by passing `defaultValue` to `MessageInput` in `Thread` component. If the default value has to be generated / retrieved, it cannot be set  with the first render of the `Thread` component as the thread is not yet opened / known and the original `defaultValue` passed to `Thread` would not get overwritten once the `thread` value is set in `ChatContext`.

### DIFF
--- a/docusaurus/docs/React/guides/customization/persist-input-text-in-local-storage.mdx
+++ b/docusaurus/docs/React/guides/customization/persist-input-text-in-local-storage.mdx
@@ -1,0 +1,181 @@
+---
+id: persist_input_text_in_localstorage
+sidebar_position: 6
+slug: /guides/persist-input-text-in-localstorage/
+title: Storing message drafts
+---
+
+In this recipe, we would like to demonstrate how you can start storing unsent user's messages as drafts. The whole implementation turns around the use of `MessageInput`'s prop `getDefaultValue` and custom change event handler. We will store the messages in localStorage.
+
+
+## Building the draft storage logic
+Below, we have a simple logic to store all the message text drafts in a localStorage object under the key `@chat/drafts`.
+
+```ts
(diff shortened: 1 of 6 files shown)


docusaurus/docs/React/guides/customization/persist-input-text-in-local-storage.mdx
```diff
@@ -0,0 +1,181 @@
...
+    drafts[key] = value
+  }
+  localStorage.setItem(STORAGE_KEY, JSON.stringify(drafts))
+}
+```
+
+On top of this logic we build a hook that exposes the change handler functions for both thread and main `MessageInput` components as well as functions for `MessageInput`'s `getDefaultValue` prop. We also have to override the `MessageInput`'s default submit handler, because we want to remove the draft from storage when a message is sent.
+
+```ts
+import {ChangeEvent, useCallback } from 'react';
```
arnautov-anton, line 41: Apart from code formatting in the docs I think this is a really nice guide. 👍

docusaurus/docs/React/guides/customization/persist-input-text-in-local-storage.mdx
```diff
@@ -0,0 +1,181 @@
...
+import {
+    MessageToSend,
+    useChannelActionContext,
+    useChannelStateContext,
+} from 'stream-chat-react';
+import {
```
arnautov-anton, line 47:
```suggestion
import type {
```

oliverlaz (approved):

arnautov-anton (approved):

PR #2018, opened 2023-05-25
fix: cooldown timer

MartinCupela:
### 🎯 Goal

Adapt the cooldown timer enablement to the latest bug fix on the backend:

The `'slow-mode'` own capability was replaced by `'skip-slow-mode'`. This capability should only be used to determine if the user can skip cooldown mode and not to determine if the channel has cooldown enabled. The `channel.own_capabilities` array will contain capability `'skip-slow-mode'` if a user is assigned a role that has `skip-channel-cooldown` permission.

### 🛠 Implementation details
Cooldown mode enablement is signaled by the presence of `channel.cooldown` value (num of seconds).

When cooldown mode is enabled a user can send a message only if:
1.  in the cooldown period if `channel.own_capabilities` includes `'skip-slow-mode'`
2. the time elapsed since the last message sent is > `channel.cooldown`
3. no previous message was sent by the given user in the given channel 

Otherwise the user has to wait until the cooldown period has elapsed before sending next message.

#### Updating cooldown state
✅  If the cooldown is enabled during active user connection (`channel.updated` event contains `channel.cooldown` value), then the remaining cooldown time has to be recalculated based on the above criteria.
✅  If the cooldown period is updated (`channel.updated` event contains new `channel.cooldown` value), then the remaining cooldown time has to be recalculated based on the above criteria.
✅  If the cooldown is disabled (`channel.updated` event does not contain `channel.cooldown` value), then the remaining cooldown time has to be set to 0 immediately, thus enabling sending messages.

#### Known limitations
❌  Cooldown is not updated over WS if `skip-channel-cooldown`  perm is assigned to a role in dashboard. Channel data has to be queried anew to reflect that change.
❌  Cooldown is not updated over WS if `skip-channel-cooldown`  perm is removed from a role in dashboard. Channel data has to be queried anew to reflect that change.

### DIFF
--- a/src/components/MessageInput/hooks/useCooldownTimer.tsx
+++ b/src/components/MessageInput/hooks/useCooldownTimer.tsx
@@ -1,9 +1,7 @@
 import React, { useEffect, useMemo, useState } from 'react';
-
-import { useChatContext } from '../../../context/ChatContext';
-import { useChannelStateContext } from '../../../context/ChannelStateContext';
-
 import type { ChannelResponse } from 'stream-chat';
+
+import { useChannelStateContext, useChatContext } from '../../../context';
 
 import type { DefaultStreamChatGenerics } from '../../../types/types';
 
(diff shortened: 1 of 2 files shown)


src/components/MessageInput/hooks/useCooldownTimer.tsx
```diff
@@ -36,17 +35,19 @@ export const useCooldownTimer = <
...
-    const remainingCooldown = Math.round(
-      cooldownInterval - (new Date().getTime() - ownLatestMessageDate.getTime()) / 1000,
+    const timeSinceOwnLastMessage = ownLatestMessageDate
+      ? (new Date().getTime() - ownLatestMessageDate.getTime()) / 1000
+      : undefined;
+
+    setCooldownRemaining(
+      !skipCooldown && timeSinceOwnLastMessage && cooldownInterval > timeSinceOwnLastMessage
+        ? Math.round(cooldownInterval - timeSinceOwnLastMessage)
+        : 0,
```
arnautov-anton, line 45: Why this change? It looks like it's doing the same thing just a bit harder to read on a first glance.

MartinCupela: You need to set also 0 cooldown remaining, what was not happening before. Now the condition states the 3 cases, when remaining cooldown is non-zero. Does not seem to me harder to read.

oliverlaz (approved):

src/components/MessageInput/hooks/useCooldownTimer.tsx
arnautov-anton: What does 0 seconds cooldown remaining look like in the UI? I'd imagine 0 seconds remaining would also mean skip - or?

MartinCupela: yes, it means no cooldown

PR #2020, opened 2023-05-29
feat: add polish translations to i18n

Soib:
# Submit a pull request

### 🎯 Goal

The objective of this pull request is to integrate the stream-chat-react library into our application to provide support for Polish, Spanish, and English-speaking users. Currently, the UI does not support the Polish language, and the purpose of this pull request is to extend stream-chat-react to include Polish language support.

### 🛠 Implementation details

To achieve this objective, I have undertaken the following implementation steps:

1. Translated the necessary content and created a new `pl.json` file in the i18n folder.
2. Added imports and configs in `Streami18n.ts` and `TranslationContext.tsx`

### 🎨 UI Changes

none


MartinCupela:
@Soib thank you for submitting the PR. Before merging this PR, could you please try to follow [our guide](https://getstream.io/chat/docs/sdk/react/guides/theming/translations/#add-a-language) on adding different than supported language to your app? We would prefer not to extend the support for more language mutations in our SDK at the moment.

Soib:
@MartinCupela yeah, I've created hook to load custom translations in our app
image

it works well, but I didn't find a way to provide translations for texts related to dates:

image

as you can see, the dates are in English. Is it possible to work around it somehow?

MartinCupela:
Hey @Soib, it is possible to add translations for weekdays, calendar ([see dayjs docs here](https://day.js.org/docs/en/plugin/calendar)) etc when initiating the i18n instance:
image

As you can see, I am casting the calendar config object to circumvent typescript. This is because the `dayjsLocaleConfigForLanguage` param is not correctly typed (missing `calendar` key).

I will prepare a fix for this + add missing documentation, but you do not need to wait for it as the rest of the config keys work with intellisense. Please let me know, whether you were able to add accomplish the task.

Soib:
@MartinCupela yeah, that solves the problem. I can live with @ts-ignore for now. Thanks a bunch for help!🥇 

I guess we can close the PR 
image

PR #2041, opened 2023-07-09
fix: enable search results scrolling

MartinCupela:
### 🎯 Goal

Fix the bug in channel search results list to make it scrollable. If the channel search result list overflew the parent, it was not possible to scroll the list.

This fix depends on https://github.com/GetStream/stream-chat-css/pull/232

### DIFF
--- a/src/components/ChannelSearch/ChannelSearch.tsx
+++ b/src/components/ChannelSearch/ChannelSearch.tsx
@@ -1,6 +1,7 @@
+import clsx from 'clsx';
 import React from 'react';
 
-import { useChatContext } from '../../context/ChatContext';
+import { useChatContext } from '../../context';
 
 import { ChannelSearchControllerParams, useChannelSearch } from './hooks/useChannelSearch';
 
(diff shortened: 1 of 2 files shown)


arnautov-anton (approved):

ayushidianapps:
Team, your fix has created issues in our project.
The inline class has distorted the UI, I think this is not a good solution, you might need to add more conditions before giving classes.

arnautov-anton:
Hey, @ayushidianapps, thank you for pointing this out - could you also share screenshots of the broken UI, please? @MartinCupela we should probably revisit this one.

MartinCupela:
@ayushidianapps if I understood well, you have CSS rules for `inline` class and those rules now affect the element, that received this class in search component. Is that right? We should probably namespace this class to `str-chat__channel-search--inline`. Would that solve your problem?

ayushidianapps:
Hi,
I am not sure that would solve our problem because that class would be
picked from the css automatically, right?

Is there any condition for you to have the inline class?

MartinCupela:
@ayushidianapps CSS rules are applied based on classes associated with the elements. Could you please answer the following for us to better understand the problem:

> if I understood well, you have CSS rules for inline class and those rules now affect the element, that received this class in search component.

ayushidianapps:
Yes the inline class had affected our CSS

MartinCupela:
@ayushidianapps we have [released fix for this](https://github.com/GetStream/stream-chat-react/releases/tag/v10.8.7). The added class is now namespaced and should not be impacted by your `.inline` rule.

ayushidianapps:
Awesome, thanks for the update!

PR #2059, opened 2023-07-17
fix: provide default `members` filter param for channel search to avoid error response

MartinCupela:
### 🎯 Goal

Provide minimum default `members` filter param for channel search to avoid error response:

`queryChannels failed with error: \"1 channels match your query but cannot be returned because you don't have access to them. Did you forget to include {members: $in: [\"<user_id>\"]}?\"`


MartinCupela:
The error is returned, if at least 1 channel. that the user has not permission to read, matches the query string. This can be tackled by providing custom channel or user filters as additional search props.

Issue #1827, opened 2022-10-27
use tslib in package bundling process


MartinCupela:
**Motivation**

Configure properly the use of already installed `tslib` dependency in the build process and reduce the bundle size.

PR #2090, opened 2023-09-09
chore(deps): general build audit & tslib

arnautov-anton:
### 🎯 Goal

- remove CSS mappings from the build as they point to non-existing files within the `dist` folder
- refresh `tsconfig.json` (feel free to commit your changes)
- move `tslib` and actually use it

(I feel like we should also do something about all the Webpack packages and do some ESLint cleaning - it's way too strict)

Closes: #1540
Closes: #1827

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "react-virtuoso": "^2.16.5",
     "remark-gfm": "^3.0.1",
     "textarea-caret": "^3.1.0",
+    "tslib": "^2.6.2",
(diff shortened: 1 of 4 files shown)


oliverlaz:
I still see duplicated `function objectSpread$1(...)` declarations in the final bundles. I expect these to be coming from the `tslib` you configured. I didn't analyze in detail whether they come from a third party dependency or from us.

oliverlaz (approved):

PR #2104, opened 2023-09-25
feat: add messageDeliveryStatus prop to ChannelListPreview

MartinCupela:
### 🎯 Goal

Provide delivery status calculated value to for custom `ChannelPreview` UI component. The default UI component `ChannelPreviewMessenger` does not support the message delivery status visuals.

### 🛠 Implementation details

The `ChannelPreview` component does not have access to the information about a message being sent and not delivered yet and so can infer only the following two delivery states :

1. `DELIVERED`
2. `READ`

The states are exposed to the SDK users in an enum:

```tsx
export enum MessageDeliveryStatus {
  DELIVERED = 'delivered',
  READ = 'read',
}
```

The hook that performs that calculations *`useMessageDeliveryStatus`* is not exported for the SDK users to protect the SDK internals.

### Specs

- The delivery status is calculated only for the last message in the message set

- The delivery status is calculated only for own messages and not those sent by other users

- The delivery status is calculated from the field `read` (`ReadResponse` array) available on the `ChannelAPIResponse`.

- The status is undefined if the last message is not own message.

- The message is considered to be `DELIVERED` on `ChannelPreview` mount, if the channel’s `read` array does not contain `ReadResponse` newer than **the last message creation date**.

- The message is considered to be `DELIVERED` when the message creator receives the `message.new` WS event

- The message is considered to be `READ` on `ChannelPreview` mount, if the channel’s `read` array contains `ReadResponse` newer than **the last message creation date**.

- The message is considered to be `READ` when the message creator receives the the first `message.read` WS event

The status value changes are depicted on the state machine visualization below:

message_delivery_status_state_machine

### 🎨 UI Changes

None

### DIFF
--- a/src/components/ChannelPreview/ChannelPreviewMessenger.tsx
+++ b/src/components/ChannelPreview/ChannelPreviewMessenger.tsx
@@ -20,7 +20,6 @@ const UnMemoizedChannelPreviewMessenger = <
     displayImage,
     displayTitle,
     latestMessage,
-    messageDeliveryStatus,
     onSelect: customOnSelectChannel,
     setActiveChannel,
     unread,
@@ -71,7 +70,6 @@ const UnMemoizedChannelPreviewMessenger = <
               {unread}
             </div>
           )}
-          {messageDeliveryStatus && <span>{messageDeliveryStatus}</span>}
         </div>
         <div className='str-chat__channel-preview-messenger--last-message'>{latestMessage}</div>
       </div>
--- a/src/components/ChannelPreview/ChannelPreviewMessenger.tsx
+++ b/src/components/ChannelPreview/ChannelPreviewMessenger.tsx
@@ -20,6 +20,7 @@ const UnMemoizedChannelPreviewMessenger = <
     displayImage,
     displayTitle,
     latestMessage,
+    messageDeliveryStatus,
     onSelect: customOnSelectChannel,
     setActiveChannel,
     unread,
@@ -70,6 +71,7 @@ const UnMemoizedChannelPreviewMessenger = <
               {unread}
             </div>
           )}
+          {messageDeliveryStatus && <span>{messageDeliveryStatus}</span>}
         </div>
         <div className='str-chat__channel-preview-messenger--last-message'>{latestMessage}</div>
       </div>
--- a/src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
+++ b/src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
@@ -26,7 +26,9 @@ export const useMessageDeliveryStatus = <
   lastMessage,
 }: UseMessageStatusParamsChannelPreviewProps<StreamChatGenerics>) => {
   const { client } = useChatContext();
-  const [messageDeliveryStatus, setMessageDeliveryStatus] = useState<MessageDeliveryStatus | undefined>();
+  const [messageDeliveryStatus, setMessageDeliveryStatus] = useState<
+    MessageDeliveryStatus | undefined
+  >();
 
   const isOwnMessage = useCallback(
     (message?: StreamMessage<StreamChatGenerics>) =>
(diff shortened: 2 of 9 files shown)


src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+        return ignoreOwnReadStatus && lastMessageCreatedAtDate < channelLastMarkedReadDate;
+      },
+    );
+
+    setMessageDeliveryStatus(
+      channelReadByOthersAfterLastMessageUpdate
+        ? MessageDeliveryStatus.READ
+        : MessageDeliveryStatus.DELIVERED,
+    );
+  }, [channel.state.read, client, isOwnMessage, lastMessage?.created_at]);
```
oliverlaz, line 58: We don't have `lastMessage` in the list of dependencies here. Is that intentional?

src/components/ChannelPreview/ChannelPreviewMessenger.tsx
```diff
@@ -70,6 +71,7 @@ const UnMemoizedChannelPreviewMessenger = <
               {unread}
             </div>
           )}
+          {messageDeliveryStatus && <span>{messageDeliveryStatus}</span>}
```
oliverlaz, line 74: Do we intentionally want to write `read` or `delivered` in plain text here?

MartinCupela: Will remove this one. This PR should not introduce any change to `ChannelPreviewMessenger`.

src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
oliverlaz: Theoretically, we need the whole message, as it is used in `isOwnMessage` that internally checks for `message.user.id`.
Also, we should check whether there are accidental mutations of `lastMessage` that may break the `useEffect`.

src/components/ChannelPreview/ChannelPreviewMessenger.tsx
MartinCupela: Done

src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
MartinCupela: True

MartinCupela: Fixed

src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+    channel.on('message.new', handleMessageNew);
+    channel.on('message.read', handleMarkRead);
+
+    return () => {
+      channel.off('message.new', handleMessageNew);
+      channel.off('message.read', handleMarkRead);
+    };
+  }, [channel, client, lastMessage?.id, isOwnMessage]);
```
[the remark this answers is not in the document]
MartinCupela: I have added a fix meanwhile fixing the other dependency. We need `lastMessge` to determine, whether to subscribe to `message.read` event.

src/components/ChannelPreview/hooks/useMessageDeliveryStatus.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+export const useMessageDeliveryStatus = <
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>({
+  channel,
+  lastMessage,
+}: UseMessageStatusParamsChannelPreviewProps<StreamChatGenerics>) => {
+  const { client } = useChatContext();
+  const [messageDeliveryStatus, setMessageDeliveryStatus] = useState<MessageDeliveryStatus>();
```
arnautov-anton, line 29:
```suggestion
  const [messageDeliveryStatus, setMessageDeliveryStatus] = useState<MessageDeliveryStatus | undefined>();
```

oliverlaz (approved):

PR #2123, opened 2023-10-11
feat: allow complete channel list throttled reload on internet connection recovery

MartinCupela:
### 🎯 Goal

Allow to load the most recent updates to the ChannelList after the internet connection loss. The `StreamChat` client's recovery mechanism has the following shortcomings:

- re-queries only the channels that have already been loaded into the UI
-  does not throttle the channel list queries.

On the other side, the SDK's recovery mechanism just reloads the channel list based on the filters, sort and options parameters originally provided to `ChannelList` through props.

This is an opt-in feature that has to be enabled as described in the documentation.

Currently it is not possible to load both the latest updates that would include new channels at the top and the originally loaded channels due to the fact that after re-querying the originally loaded channels we loose track of the next page for the channels pagination.

### DIFF
--- a/src/components/ChannelList/hooks/usePaginatedChannels.ts
+++ b/src/components/ChannelList/hooks/usePaginatedChannels.ts
@@ -96,7 +96,7 @@ export const usePaginatedChannels = <
 
     lastRecoveryTimestamp.current = now;
     queryChannels('reload');
-  }, [error, queryChannels, lastRecoveryTimestamp]);
+  }, [error, queryChannels, recoveryThrottleInterval]);
 
   const loadNextPage = () => {
     queryChannels();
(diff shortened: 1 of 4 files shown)


src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -72,10 +83,34 @@ export const usePaginatedChannels = <
...
+      ? now - lastRecoveryTimestamp.current
+      : 0;
+
+    if (!isFirstRecovery && timeElapsedSinceLastRecoveryMs < recoveryThrottleInterval && !error) {
+      return;
+    }
+
+    lastRecoveryTimestamp.current = now;
+    queryChannels('reload');
+  }, [error, queryChannels, lastRecoveryTimestamp]);
```
oliverlaz, line 99:
```suggestion
  }, [error, queryChannels, recoveryThrottleInterval]);
```

- Refs shall not be put in dependencies
- `recoveryThrottleInterval` isn't listed here.

oliverlaz (approved):

oliverlaz (approved):

PR #2140, opened 2023-10-23
feat(VirtualizedMessageList): allow to merge custom virtuoso components with the SDK defaults

MartinCupela:
### 🎯 Goal

Allow integrators to add only some custom virtuoso components via `VirtualizedMessageListProps['additionalVirtuosoProps']['components']` and the rest to be filled with the defaults provided by the SDK.

### 🛠 Implementation details

The `MessageList/hooks` directory structure has changed. Now the hooks used exclusively by `VirtualizedMessageList` are placed in `hooks/VirtualizedMessageList` and those used exclusively by `MessageList` are stored in `hooks/MessageList`. All the hooks that used to be exported from the `MessageList/hooks` folder are still exported.

I have also extracted new hooks from `VirtualizedMessageList` code and placed them into `hooks/VirtualizedMessageList`. These are not exported for the integrators to use. These hooks are:

1. `useMessageSetKey`
2. `useScrollToBottomOnNewMessage`

### 🎨 UI Changes

None.


src/components/MessageList/hooks/VirtualizedMessageList/useVirtualizedMessageListComponents.tsx
```diff
@@ -0,0 +1,200 @@
...
+      ...virtuosoComponentsFromProps,
+    };
+  }, [
+    EmptyStateIndicator,
+    LoadingIndicator,
+    TypingIndicator,
+    loadingMore,
+    head,
+    threadList,
+    virtuosoComponentsFromProps,
```
oliverlaz, line 195: I have a feeling that this may re-introduce this bug: https://github.com/GetStream/stream-chat-react/pull/2067
as it can break the memoization, in case someone uses this:
```
<VirualizedMessageList additionalVirtuosoProps={{ components: { Footer: MyFooter }}} />
```

Can you please check that?

MartinCupela: Checked, whether the list rerenders:

1. add video
2. play video
3. add new messages

The video kept playing.

I have also added a current timestamp to the `components` objects to observe, whether the `components` object would change (if the timestamp changes). The timestamp does not change even when scrolling the list.

```
< VirualizedMessageList additionalVirtuosoProps={{ components: { Footer: MyFooter, d: new Date().toISOString() }}} />
```

PR #2142, opened 2023-10-25
feat(renderText): allow custom remark and rehype plugin composition

MartinCupela:
### 🎯 Goal

Add optional parameters to `renderText()` to enable configuration of rehype and remark plugins used to transform message.text into React elements.


oliverlaz (approved):

src/components/Message/renderText/renderText.tsx
```diff
@@ -0,0 +1,191 @@
...
+export const markDownRenderers: RenderTextOptions['customMarkDownRenderers'] = {
+  a: Anchor,
+  emoji: Emoji,
+  mention: Mention,
+};
+
+export type RenderTextOptions<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = {
+  allowedTagNames?: Array<keyof JSX.IntrinsicElements | 'emoji' | 'mention' | string>;
```
arnautov-anton, line 69:
```suggestion
  allowedTagNames?: Array<keyof JSX.IntrinsicElements | 'emoji' | 'mention' | (string & {})>;
```
This way you get code-completion to work as well. :)

https://stackoverflow.com/a/61048124

arnautov-anton (approved):

src/components/Message/renderText/renderText.tsx
MartinCupela: added and disabled applied eslint rule for now. thank you :)

PR #2170, opened 2023-11-13
feat: expose optional remark plugin to keep all line breaks and keep HTML in message text

MartinCupela:
### 🎯 Goal

Provide a remark plugin that would allow the integrators to keep the line breaks in the displayed message text. By default, a sequence of multiple line breaks in a markdown string leads to creation of a paragraph element by the `remark-parse` library used internally by `react-markdown` (used internally by `renderText`). To keep the line breaks, these have to be injected while transforming the mdast tree with a custom remark plugin. 

This PR introduces export of such plugin but does apply the plugin by default. Integrators that wish to keep those line breaks will have to opt in to use of the plugin as follows:

```tsx
import { UserResponse } from 'stream-chat';
import {
  htmlToStringPlugin,
  keepLineBreaksPlugin,
  MessageList,
  MessageListProps,
  renderText,
  RenderTextPluginConfigurator,
} from 'stream-chat-react';

const getRemarkPlugins: RenderTextPluginConfigurator = (plugins) => [
  keepLineBreaksPlugin,
  ...plugins,
];

function customRenderText(text?: string, mentionedUsers?: UserResponse[]) {
  return renderText(text, mentionedUsers, { getRemarkPlugins });
}

export const CustomMessageList = (props: MessageListProps) => (
  <MessageList {...props} renderText={customRenderText} />
);
```

**Display HTML in message text**

The PR introduces a fix that renders the html in the text as a part of the message text. Until now, the HTML tags were scraped and only the text inside was rendered. Now a remark plugin `htmlToTextPlugin` is available to render the HTML string as text.

```tsx
import React from 'react';
import { UserResponse } from 'stream-chat';
import {
  htmlToTextPlugin,
  keepLineBreaksPlugin,
  MessageList,
  MessageListProps,
  renderText,
  RenderTextPluginConfigurator,
} from 'stream-chat-react';

const getRemarkPlugins: RenderTextPluginConfigurator = (plugins) => [
  htmlToTextPlugin,
  keepLineBreaksPlugin,
  ...plugins,
];

function customRenderText(text?: string, mentionedUsers?: UserResponse[]) {
  return renderText(text, mentionedUsers, { getRemarkPlugins });
}

export const CustomMessageList = (props: MessageListProps) => (
  <MessageList {...props} renderText={customRenderText} />
);
```

### 🎨 UI Changes

**keep line breaks**

image

**display html as string**

image


oliverlaz (approved):

src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+  const ownStartLine = node.position.start.line;
+  const prevEndLine = prevSibling.position.end.line;
+
+  // the -1 is adjustment for the single line break into which multiple line breaks are converted
+  const countTruncatedLineBreaks = ownStartLine - prevEndLine - 1;
+  if (countTruncatedLineBreaks < 1) return;
+
+  const lineBreaks = Array.from({ length: countTruncatedLineBreaks }).map(() =>
+    u('break', { tagName: 'br' }),
+  ) as Break[];
```
arnautov-anton, line 25:
```suggestion
  const lineBreaks = Array.from<unknown, Break>({ length: countTruncatedLineBreaks }, () =>
    u('break', { tagName: 'br' }),
  );
```

PR #2175, opened 2023-11-16
fix: lift notifications above modal overlay

MartinCupela:
### 🎯 Goal

Depends on https://github.com/GetStream/stream-chat-css/pull/249


oliverlaz (approved):

arnautov-anton (approved):

arnautov-anton:
@MartinCupela @oliverlaz - question - why do we keep `@stream-io/stream-chat-css` in the dependencies if we only use it during the build process - wouldn't it make more sense to move it to `devDeps`? 🤔

PR #2187, opened 2023-11-24
feat: add channel list context

MartinCupela:
### 🎯 Goal

Allow to control and access the `ChannelList` state  with `channels` and `setChannels`.


arnautov-anton (approved):
I feel like at some point we should build a better ChannelList component which is composable through `children` instead. It can live alongside and this one will get deprecated and removed eventually.

oliverlaz (approved):

PR #2192, opened 2023-11-28
chore: Git Hooks & general cleanup

arnautov-anton:
### 🎯 Goal

Personally - not a big fan of Git Hooks - but I believe the general consensus is to keep them so I did some upgrading/cleaning up. I also removed the Python script as we're not commiting any binaries to our SDK and I'm not installing Python on my machine to run it.

Also - I'm considering ditching the `commitlint` extension as all of our PR's are squashed and that commit message is being checked by the CI (Check PR title).

Removed some of the other files that I believe are unused and no longer relevant.


MartinCupela:
I think that we can remove commitlint once it will not be possible to push directly to master :)

MartinCupela (approved):

oliverlaz (approved):

PR #2208, opened 2023-12-11
fix: schedule cooldown removal from useCooldownTimer hook instead of CooldownTimer

MartinCupela:
### 🎯 Goal

Schedule the cooldown removal in `useCooldownTimer` hook so that all the descendants can update their UI.
Properly update CooldownTimer state with setTimeout to avoid freezing of the countdown during higher message rate influx.

### DIFF
--- a/src/components/MessageInput/CooldownTimer.tsx
+++ b/src/components/MessageInput/CooldownTimer.tsx
@@ -4,20 +4,24 @@ export type CooldownTimerProps = {
   cooldownInterval: number;
   setCooldownRemaining: React.Dispatch<React.SetStateAction<number | undefined>>;
 };
-export const CooldownTimer = ({ cooldownInterval, setCooldownRemaining }: CooldownTimerProps) => {
-  const [seconds, setSeconds] = useState(cooldownInterval);
+export const CooldownTimer = ({ cooldownInterval }: CooldownTimerProps) => {
+  const [seconds, setSeconds] = useState<number | undefined>();
 
   useEffect(() => {
-    const countdownInterval = setInterval(() => {
-      if (seconds > 0) {
+    let countdownTimeout: ReturnType<typeof setTimeout>;
+    if (seconds) {
+      countdownTimeout = setTimeout(() => {
         setSeconds(seconds - 1);
-      } else {
-        setCooldownRemaining(0);
-      }
-    }, 1000);
+      }, 1000);
+    }
+    return () => {
+      clearTimeout(countdownTimeout);
+    };
+  }, [seconds]);
 
-    return () => clearInterval(countdownInterval);
-  });
+  useEffect(() => {
+    setSeconds(cooldownInterval ?? 0);
+  }, [cooldownInterval]);
 
   return (
     <div className='str-chat__message-input-cooldown' data-testid='cooldown-timer'>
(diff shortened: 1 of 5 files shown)


src/components/MessageInput/CooldownTimer.tsx
```diff
@@ -4,20 +4,24 @@ export type CooldownTimerProps = {
   cooldownInterval: number;
   setCooldownRemaining: React.Dispatch<React.SetStateAction<number | undefined>>;
```
oliverlaz, line 5: I think we don't need this prop anymore, right?

src/components/MessageInput/CooldownTimer.tsx
```diff
@@ -4,20 +4,24 @@ export type CooldownTimerProps = {
...
-export const CooldownTimer = ({ cooldownInterval, setCooldownRemaining }: CooldownTimerProps) => {
-  const [seconds, setSeconds] = useState(cooldownInterval);
+export const CooldownTimer = ({ cooldownInterval }: CooldownTimerProps) => {
+  const [seconds, setSeconds] = useState<number | undefined>();
 
   useEffect(() => {
-    const countdownInterval = setInterval(() => {
-      if (seconds > 0) {
+    let countdownTimeout: ReturnType<typeof setTimeout>;
+    if (seconds) {
```
oliverlaz, line 12:
```suggestion
    if (seconds > 0) {
```

I think we should still keep this condition, as in case it happens that someone provides `-1` as a `cooldownInterval` this loop will run forever :)

src/components/MessageInput/hooks/useCooldownTimer.tsx
```diff
@@ -40,13 +40,24 @@ export const useCooldownTimer = <
...
-    );
+        : 0;
+
+    setCooldownRemaining(remaining);
+
+    if (!remaining) return;
+
+    const timeout = setTimeout(() => {
+      setCooldownRemaining(0);
+    }, remaining * 1000);
```
oliverlaz, line 56: Is this timeout here to reset the "global" cooldown remaining? And the timeout in `CooldownTimer.tsx` is there only for updating the UI?

Can we move everything here and use `cooldownRemaining` in the Component? Are we avoiding this to save on re-renders?

src/components/MessageInput/CooldownTimer.tsx
MartinCupela: Removing it would be a breaking change as the interface is given also for the custom variations of the timer:

image

src/components/MessageInput/CooldownTimer.tsx
```diff
@@ -4,20 +4,24 @@ export type CooldownTimerProps = {
...
-export const CooldownTimer = ({ cooldownInterval, setCooldownRemaining }: CooldownTimerProps) => {
-  const [seconds, setSeconds] = useState(cooldownInterval);
+export const CooldownTimer = ({ cooldownInterval }: CooldownTimerProps) => {
+  const [seconds, setSeconds] = useState<number | undefined>();
 
   useEffect(() => {
-    const countdownInterval = setInterval(() => {
-      if (seconds > 0) {
+    let countdownTimeout: ReturnType<typeof setTimeout>;
+    if (seconds > 0) {
```
arnautov-anton, line 12:
```suggestion
    if (typeof seconds === 'number' && seconds > 0) {
```
Just to cover this error:
`'seconds' is possibly 'undefined'.`

src/components/MessageInput/CooldownTimer.tsx
oliverlaz: ok, I see... it is messed up :)

src/components/MessageInput/CooldownTimer.tsx
MartinCupela: @arnautov-anton I think we do not have to check for undefined here as you can compare undefined with number:

image

oliverlaz (approved):

PR #2211, opened 2023-12-12
fix: mark as read messages sent by current user

arnautov-anton:
### 🎯 Goal

Adjusts Channel behavior to mark as read messages sent by current user as well, this adjustment is made to match the behavior of Angular SDK.

### DIFF
--- a/src/components/Channel/__tests__/Channel.test.js
+++ b/src/components/Channel/__tests__/Channel.test.js
@@ -1347,6 +1347,20 @@ describe('Channel', () => {
         await waitFor(() => expect(markReadSpy).toHaveBeenCalledWith());
       });
 
+      it('should mark the channel as read if a new message current user comes in and the user is looking at the page', async () => {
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):

src/components/Channel/__tests__/Channel.test.js
```diff
@@ -1347,6 +1347,20 @@ describe('Channel', () => {
         await waitFor(() => expect(markReadSpy).toHaveBeenCalledWith());
       });
 
+      it('should mark the channel as read if a new message current user comes in and the user is looking at the page', async () => {
```
MartinCupela, line 1350:
```suggestion
      it('should mark the channel as read if the new message author is the current user and the user is looking at the page', async () => {
```

MartinCupela (approved):

szuperaz (approved):

PR #2213, opened 2023-12-12
fix: throttle all mark read requests

MartinCupela:
### 🎯 Goal

todo:
- [ ] verify that the current throttle function really throttles the requests.


MartinCupela:
Will be solved in #2238

PR #2214, opened 2023-12-12
Pass groupStyles through VirtualizedMessageList

TaylorBeeston:
# Submit a pull request

### 🎯 Goal

We are trying to switch to `VirtualizedMessageList` and found that our custom `groupStyles` stopped working!

When I dug around, I found that the prop is simply not getting passed through.

### 🛠 Implementation details

Inside of `messageRender` in `VirtualizedMessageListComponents`, I grab the `messageGroupStyles` object
off of the Virtuoso context, grab the `groupStyles` string that has been set for this message, then
add it as the `groupStyles` prop for the `Messsage` component.

### 🎨 UI Changes

_Add relevant screenshots_

I was having trouble getting this to build locally to get some screenshots =(
It would build fine, but I was getting errors about multiple versions of react when trying to link
the locally built package. I figured that's probably more headache than it's worth so I thought I'd
just get this PR up!

### DIFF
--- a/src/components/MessageList/VirtualizedMessageListComponents.tsx
+++ b/src/components/MessageList/VirtualizedMessageListComponents.tsx
@@ -108,6 +108,7 @@ export const messageRenderer = <
     lastReceivedMessageId,
     Message: MessageUIComponent,
     messageActions,
+    messageGroupStyles,
     MessageSystem,
     numItemsPrepended,
     ownMessagesReadByOthers,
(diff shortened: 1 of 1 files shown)


gerardopar:
🚀

MartinCupela:
Hey @TaylorBeeston, thank you for the submission. We will pick it up from here as there are tests missing :)

PR #2223, opened 2023-12-20
fix: add callback to report fetching active channel errors

Annastrin:
# Submit a pull request

### 🎯 Goal

When we added opening a channel via react router url to a channel, the user is able to open any channel. 
If the channel exists but the user doesn't have access (is not a member of this channel), there's an error in activeChannelHandler (see screenshot).

### 🛠 Implementation details

Added try/catch block and callback prop to report to ChannelList's parent component

### 🎨 UI Changes
Screenshot from 2023-12-21 00-36-10

### DIFF
--- a/src/components/ChannelList/ChannelList.tsx
+++ b/src/components/ChannelList/ChannelList.tsx
@@ -116,6 +116,7 @@ export type ChannelListProps<
     setChannels: React.Dispatch<React.SetStateAction<Array<Channel<StreamChatGenerics>>>>,
     event: Event<StreamChatGenerics>,
   ) => void;
+  onGetActiveChannelError?: (error: unknown) => void;
   /** An object containing channel query options */
   options?: ChannelOptions;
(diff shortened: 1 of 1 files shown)


arnautov-anton:
Hey, @Annastrin, I don't think we're going to proceed with such mechanism - instead, please try watching channel manually and handle error accordingly - set as active if no error pops up, here's the snippet:

```tsx
const Navigator = ({ channelType, channelId }) => {
  const { client, setActiveChannel } = useChatContext();

  useEffect(() => {
    if (!client) return;

    const channel = client.channel(channelType, channelId);

    channel
      .watch()
      .then(() => {
        setActiveChannel(channel);
      })
      .catch((error) => {
        // handle error

        console.error(error);
      });
  }, [client]);

  return null;
};
```

(you could look for it locally first as well - to save one extra request if it's already in local state)

Make sure to set `ChannelList` property `setActiveChannelOnMount` to `false` to make this work properly.

PR #2239, opened 2024-01-12
fix: use username if name does not exist

xylb-pro:


MartinCupela:
@xylb-pro we will not accept this PR as none of the other SDKs provided by Stream recognizes the `username` field. Please use your own customized components that will reflect your user model according to your needs. Thank you

PR #2241, opened 2024-01-15
fix: use popper to properly position message actions box

myandrienko:
### 🎯 Goal

Message actions box is currently relatively positioned inside MessageList, which means it sometimes gets clipped by MessageList's boundaries.

This PR implements proper positioning for the actions box, preventing it from being clipped in (almost) every case.

### 🛠 Implementation details

`stream-chat-react` already uses `react-popper` for tooltips, so it made sense to reuse it for the actions box as well.

My initial plan was to render the actions box into a portal and position it using popper, but using a portal turned out to be unnecessary: with the positioning strategies that popper implements, the action box is never clipped by it's parent container.

See also: https://github.com/GetStream/stream-chat-css/pull/260

### 🎨 UI Changes

Previously:

image

image

After the fixes:

The actions box automatically flips if it's close to the clipping boundary:

image

image

### ✅ To-Do

- [x] Check with all message types
- [x] Check with Angular app
- [x] ~~Deal with theming v1~~
- [x] ~~Fix E2E~~

### DIFF
--- a/src/components/MessageActions/MessageActionsBox.tsx
+++ b/src/components/MessageActions/MessageActionsBox.tsx
@@ -14,7 +14,6 @@ import {
 import type { DefaultStreamChatGenerics } from '../../types/types';
 
 import { CustomMessageActionsList as DefaultCustomMessageActionsList } from './CustomMessageActionsList';
-import { useMessageActionsBoxPopper } from './hooks';
 
 type PropsDrilledToMessageActionsBox =
   | 'getMessageActions'
@@ -30,7 +29,6 @@ export type MessageActionsBoxProps<
   isUserMuted: () => boolean;
   mine: boolean;
   open: boolean;
-  referenceElement: HTMLElement | null;
 };
 
 const UnMemoizedMessageActionsBox = <
@@ -46,9 +44,7 @@ const UnMemoizedMessageActionsBox = <
     handleMute,
     handlePin,
     isUserMuted,
-    mine,
     open = false,
-    referenceElement,
   } = props;
 
   const {
@@ -60,12 +56,6 @@ const UnMemoizedMessageActionsBox = <
   );
 
   const { t } = useTranslationContext('MessageActionsBox');
-
-  const { attributes, popperElementRef, styles } = useMessageActionsBoxPopper<HTMLDivElement>({
-    open,
-    placement: mine ? 'top-end' : 'top-start',
-    referenceElement,
-  });
 
   const messageActions = getMessageActions();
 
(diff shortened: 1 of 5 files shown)


src/components/MessageActions/MessageActionsBox.tsx
```diff
@@ -46,42 +48,27 @@ const UnMemoizedMessageActionsBox = <
...
   const { setQuotedMessage } = useChannelActionContext<StreamChatGenerics>('MessageActionsBox');
-  const { customMessageActions, message, messageListRect } = useMessageContext<StreamChatGenerics>(
+  const { customMessageActions, message } = useMessageContext<StreamChatGenerics>(
     'MessageActionsBox',
   );
 
   const { t } = useTranslationContext('MessageActionsBox');
 
-  const [reverse, setReverse] = useState(false);
+  const { attributes, popperElementRef, styles } = useMessageActionsBoxPopper<HTMLDivElement>({
```
MartinCupela, line 64: I am thinking whether these could be moved up to `MessageActions` and `MessageActionsBox` would only get `ComponentProps<'div'>`. I am concerned about new prop `referenceElement` being introduced only for popper to work, but the integrators using this component may not know what to do with it. Also `referenceElement` is not mandatory, so that would be a breaking change potentially? I am thinking whether `MessageActionsBox` could forwardRef and accept div props that would be spread to the root of `MessageActionsBox`.

Not sure what @arnautov-anton or @oliverlaz think.

When adding a new prop we should document

src/components/MessageActions/hooks/useMessageActionsBoxPopper.ts
```diff
@@ -0,0 +1,45 @@
...
+    placement,
+  });
+
+  useEffect(() => {
+    if (open) {
+      // Since the popper's reference element might not be (and usually is not) visible
+      // all the time, it's safer to force popper update before showing it.
+      update?.();
+    }
+  }, [open]);
```
oliverlaz, line 38: Shouldn't we have `update` in the deps array as well?
```suggestion
  }, [open, update]);
```

src/components/MessageActions/MessageActionsBox.tsx
oliverlaz: I think your proposal makes sense (although it would make the implementation less clean). Let's try it.

myandrienko: @MartinCupela Thanks for the suggestion! I agree we should try to avoid adding a new mandatory prop, and moving the logic up to the `MessageActions` makes sense.

src/components/MessageActions/hooks/useMessageActionsBoxPopper.ts
myandrienko: @oliverlaz `usePopper` hook keeps a stable instance of the popper object, so all methods have stable references as well - but I agree we should add `update` to the deps anyway!

src/components/MessageActions/MessageActionsBox.tsx
myandrienko: @MartinCupela Fixed in https://github.com/GetStream/stream-chat-react/pull/2241/commits/8b30c038057e4cd9aee772eee6a858aeb8189ade.

However, instead of spreading div props on `MessageActionsBox` I just wrapped in an other div that is used solely for positioning - that felt a bit cleaner to me.

src/components/MessageActions/hooks/useMessageActionsBoxPopper.ts
myandrienko: @oliverlaz Fixed in https://github.com/GetStream/stream-chat-react/pull/2241/commits/b5ce2301630dd123bff4d137ea6244276e06dadd.

PR #2242, opened 2024-01-16
fix: hasMore limitations

arnautov-anton:
### 🎯 Goal

A bug that would prevent users from loading older messages within message lists if they paginated on channel B, switched to channel A and then back to channel B and tried paginating again.

### 🛠 Implementation details

Moved from `startReached` to `atTopStateChange` and from `endReached` to `atBottomStateChange` callbacks for pagination as `startReached` [is distinct](https://github.com/petyosi/react-virtuoso/blob/efdf50b8c4ee18e1c9d2fa4866e47ae9fd64cae1/src/listStateSystem.ts#L386) and called only once per set of data (I assume, not sure why as scrolling back down should ideally reset it) - but if you were to hit the top - and the data load would fail for some reason - scrolling a bit down and back up would not trigger the data load again.
This was one of my ways to try to resolve the original problem before I was able to find reliable reproduction steps but in the end decided to keep it as "futureproofing".
Though I'm worried that some of our integrators migh've hooked into these properties overriding whatever we newly have making their pagination not work.

### DIFF
--- a/src/components/MessageList/VirtualizedMessageList.tsx
+++ b/src/components/MessageList/VirtualizedMessageList.tsx
@@ -151,7 +151,6 @@ const VirtualizedMessageListWithContext = <
     defaultItemHeight,
     disableDateSeparator = true,
     groupStyles,
-    hasMore,
     hasMoreNewer,
     head,
     hideDeletedMessages = false,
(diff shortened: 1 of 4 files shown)


src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -364,7 +356,9 @@ const VirtualizedMessageListWithContext = <
         <div className={customClasses?.virtualizedMessageList || 'str-chat__virtual-list'}>
           <Virtuoso<UnknownType, VirtuosoContext<StreamChatGenerics>>
             atBottomStateChange={atBottomStateChange}
-            atBottomThreshold={200}
+            atBottomThreshold={100}
+            atTopStateChange={atTopStateChange}
```
oliverlaz, line 360: In regards to your comment in the PR description. Why don't we protect ourselves from integrators overriding this prop?

arnautov-anton: With this change - we probably should, just wanted to see if there are any other drawbacks I could be missing and we'd be tossing this change altogether.

MartinCupela (approved):

PR #2243, opened 2024-01-17
feat: allow overriding the way MessageList renders messages

myandrienko:
### 🎯 Goal

Some users want to be able to append custom elements at the start/end of `MessageList`. For maximum flexibility, we're just giving (optional) full control over how `MessageList` renders its children.

### 🛠 Implementation details

New optional prop `renderMessages` is added to `MessageList`. It receives an array of messages to be rendered, all the components from the channel's component context, and some additional metadata. It returns an array of React elements.

The default implementation is identical to how `MessageList` currently renders its children. This default implementation is also exported so that users can reuse it in their current renderers.

### 🎨 UI Changes

None by default.


src/components/Channel/Channel.tsx
```diff
@@ -1011,7 +1013,7 @@ const ChannelInner = <
       BaseImage: props.BaseImage,
       CooldownTimer: props.CooldownTimer,
       CustomMessageActionsList: props.CustomMessageActionsList,
-      DateSeparator: props.DateSeparator,
+      DateSeparator: props.DateSeparator || DateSeparator,
```
myandrienko, line 1016: This default for `DateSeparator` and `MessageSystem` components was previously set up in `useMessageListElements`: https://github.com/GetStream/stream-chat-react/blob/master/src/components/MessageList/hooks/MessageList/useMessageListElements.tsx#L59-L63

I think it makes sense to set default in `Channel`, since it's where all other defaults are provided (e.g. `Message` defaults to `MessageSimple` here).

src/components/MessageList/renderMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+  lastReceivedId,
+  messageGroupStyles,
+  messageProps,
+  messages,
+  readData,
+  threadList,
+}: RenderMessagesOptions<StreamChatGenerics>) {
+  const { DateSeparator, HeaderComponent, MessageSystem } = components;
+
+  return messages.map((message) => {
```
myandrienko, line 52: This was just extracted from `useMessageListElements`.

src/components/MessageList/hooks/MessageList/useMessageListElements.tsx
```diff
@@ -20,6 +14,7 @@ import type { MessageProps } from '../../../Message/types';
 import type { StreamMessage } from '../../../../context/ChannelStateContext';
 
 import type { DefaultStreamChatGenerics } from '../../../../types/types';
+import { MessageRenderer } from '../../renderMessages';
 
 type MessagePropsToOmit =
```
MartinCupela, line 19: I propose to get rid of the duplicate `MessagePropsToOmit` type so that we do not have to maintain it in multiple places.

src/components/MessageList/renderMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+  components,
+  customClasses,
+  lastReceivedId,
+  messageGroupStyles,
+  messageProps,
+  messages,
+  readData,
+  threadList,
+}: RenderMessagesOptions<StreamChatGenerics>) {
+  const { DateSeparator, HeaderComponent, MessageSystem } = components;
```
MartinCupela, line 50: I am thinking whether we need `components` param as when we provide our own custom rendered, we can import our custom components there. I think that gradually the integrators may start to require add more and more components to the `components` param.

src/components/MessageList/renderMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+  | 'message'
+  | 'readBy'
+  | 'threadList';
+
+export function defaultRenderMessages<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>({
+  components,
+  customClasses,
+  lastReceivedId,
```
MartinCupela, line 43: I propose to call this `lastReceivedMessageId` for more clarity

src/components/MessageList/renderMessages.tsx
MartinCupela: Please ignore, now I see this is the whole component context

src/components/MessageList/renderMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+  messageGroupStyles,
+  messageProps,
```
MartinCupela, line 45: As we are rendering a list of messages it could be unclear, for which the `messageProps` param is destined. Maybe we could give it another name like `sharedMessageProps`.

MartinCupela:
@myandrienko could you please add tests for these changes resp. code that had no test coverage previously, please?

docusaurus/docs/React/components/core-components/message-list.mdx
```diff
@@ -218,6 +218,34 @@ const CustomMessageList = () => (
 );
 ```
 
+#### Custom message list rendering
```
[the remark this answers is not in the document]
myandrienko: Sure. I guess the best place to do that is in the props reference section below. If there's a template I can use - please point me to it :)

src/components/MessageList/hooks/MessageList/useMessageListElements.tsx
myandrienko: Ooops, that's actually a mistake - I'll get rid of the duplicate `MessagePropsToOmit`. Thanks for pointing it out!

myandrienko: Fixed in https://github.com/GetStream/stream-chat-react/pull/2243/commits/75cd10bbc4bd86025636d519943b4ff91253583a.

src/components/MessageList/renderMessages.tsx
myandrienko: Fixed in https://github.com/GetStream/stream-chat-react/pull/2243/commits/bd5a39997957c6c993add0446d5f744dbcdb0a2b.

src/components/MessageList/renderMessages.tsx
myandrienko: Fixed in https://github.com/GetStream/stream-chat-react/pull/2243/commits/bd5a39997957c6c993add0446d5f744dbcdb0a2b.

myandrienko:
> @myandrienko could you please add tests for these changes resp. code that had no test coverage previously, please?

Added in https://github.com/GetStream/stream-chat-react/pull/2243/commits/98287a8376869bbcfbe76a03e42cfeb73b910b4c

docusaurus/docs/React/components/core-components/message-list.mdx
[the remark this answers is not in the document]
myandrienko: Added in https://github.com/GetStream/stream-chat-react/pull/2243/commits/991de501397a366aa0ffb3da3e4207f2e8798a55.

MartinCupela (approved):

oliverlaz (approved):

PR #2246, opened 2024-01-18
fix: get rid of positioning wrapper for message actions box

myandrienko:
### 🎯 Goal

In #2241 we've added a new positioning wrapper for `MessageActionsBox`. That was not the greatest idea, since adding new wrappers can break custom CSS for the users.

Having a wrapper is not necessary in this case anyway. This PR implements the same positioning as #2241, but without additional wrappers.


szuperaz (approved):
My review doesn't count for React changes, but it looks good to me :)

MartinCupela (approved):

oliverlaz (approved):

PR #2249, opened 2024-01-23
feat: show full list of reactions in a modal

myandrienko:
### 🎯 Goal

We're updating UI for displaying reactions. Instead of showing the latest reactions only, we're adding a separate modal window full the full list of reactions.

Although not a breaking change in terms of API, it's a breaking change in terms of UI and styling.

### 🛠 Implementation details

1. New component `ReactionsListModal` is added (rendered by default `ReactionsList`).
2. New handler added on message context level: `handleFetchReactions`. It's expected to load all reactions for current message (e.g. using our paged endpoint for fetching reactions). As usual, it can be overriden on `Message` or `ReactionsList` level.
3. Instead of showing counts for the latest reactions only, we show counts for all supported reactions in the `ReactionsList`.

### 🎨 UI Changes

image

### To-Do

- [x] Tests for ReactionsListModal
- [x] Add translations

### DIFF
--- a/docusaurus/docs/React/components/contexts/message-context.mdx
+++ b/docusaurus/docs/React/components/contexts/message-context.mdx
@@ -167,6 +167,8 @@ Function that loads all reactions for a message.
 | ------------------------------------- |
 | () => Promise<ReactionResponse[]\> \  |
 
+This function limits the number of loaded reactions to 1200. To customize this behavior, you can pass [a custom `ReactionsList` component](../message-components/reactions.mdx#handlefetchreactions).
+
 ### handleFlag
 
 Function that flags a message.
@@ -347,8 +349,8 @@ An array of users that have read the current message.
 
 Custom function to render message text content.
 
-| Type     | Default                                                                                |
-| -------- | -------------------------------------------------------------------------------------- |
+| Type     | Default                                                                        |
+| -------- | ------------------------------------------------------------------------------ |
 | function | <GHComponentLink text='renderText' path='/Message/renderText/renderText.tsx'/> |
 
 ### setEditingState
--- a/docusaurus/docs/React/components/contexts/message-context.mdx
+++ b/docusaurus/docs/React/components/contexts/message-context.mdx
@@ -161,7 +161,7 @@ Function that edits a message.
 
 ### handleFetchReactions
 
-Function that loads all reactions for a message.
+Function that loads the reactions for a message.
 
 | Type                                  |
 | ------------------------------------- |
--- a/docusaurus/docs/React/components/contexts/message-context.mdx
+++ b/docusaurus/docs/React/components/contexts/message-context.mdx
@@ -159,6 +159,14 @@ Function that edits a message.
 | ----------------------------------------------------------- |
 | (event: React.BaseSyntheticEvent) => Promise<void\> \| void |
 
+### handleFetchReactions
+
+Function that loads all reactions for a message.
+
+| Type                                  |
+| ------------------------------------- |
+| () => Promise<ReactionResponse[]\> \  |
+
 ### handleFlag
 
 Function that flags a message.
--- a/docusaurus/docs/React/components/message-components/message-ui.mdx
+++ b/docusaurus/docs/React/components/message-components/message-ui.mdx
@@ -289,7 +289,7 @@ Function that edits a message (overrides the function stored in `MessageContext`
 
 ### handleFetchReactions
 
-Function that loads all reactions for a message (overrides the function stored in `MessageContext`).
+Function that loads the reactions for a message (overrides the function stored in `MessageContext`).
 
 | Type                                                        | Default                                                                                                  |
 | ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
--- a/docusaurus/docs/React/components/message-components/message-ui.mdx
+++ b/docusaurus/docs/React/components/message-components/message-ui.mdx
@@ -287,6 +287,14 @@ Function that edits a message (overrides the function stored in `MessageContext`
 | ----------------------------------------------------------- | ------------------------------------------------------------------------------- |
 | (event: React.BaseSyntheticEvent) => Promise<void\> \| void | [MessageContextValue['handleEdit']](../contexts/message-context.mdx#handleedit) |
 
+### handleFetchReactions
+
+Function that loads all reactions for a message (overrides the function stored in `MessageContext`).
+
+| Type                                                        | Default                                                                                                  |
+| ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
+| (event: React.BaseSyntheticEvent) => Promise<void\> \| void | [MessageContextValue['handleFetchReactions']](../contexts/message-context.mdx#handlhandlefetchreactions) |
+
 ### handleFlag
 
 Function that flags a message (overrides the function stored in `MessageContext`).
--- a/docusaurus/docs/React/components/message-components/message-ui.mdx
+++ b/docusaurus/docs/React/components/message-components/message-ui.mdx
@@ -295,6 +295,8 @@ Function that loads all reactions for a message (overrides the function stored i
 | ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
 | (event: React.BaseSyntheticEvent) => Promise<void\> \| void | [MessageContextValue['handleFetchReactions']](../contexts/message-context.mdx#handlhandlefetchreactions) |
 
+This function limits the number of loaded reactions to 1200. To customize this behavior, you can pass [a custom `ReactionsList` component](./reactions.mdx#handlefetchreactions).
+
 ### handleFlag
 
 Function that flags a message (overrides the function stored in `MessageContext`).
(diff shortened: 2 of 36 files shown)


src/components/Message/Message.tsx
```diff
@@ -183,6 +186,11 @@ export const Message = <
   const handleRetry = useRetryHandler(propRetrySendMessage);
   const userRoles = useUserRole(message, onlySenderCanEdit, disableQuotedMessages);
 
+  const handleFetchReactions = useReactionsFetcher(message, {
+    getErrorNotification: getFetchReactionsErrorNotification,
+    notify: addNotification,
+  });
```
myandrienko, line 192: This works similar to other handlers we define on the message context (see, for example, `handleMute`).

src/components/Message/hooks/useReactionsFetcher.ts
```diff
@@ -0,0 +1,57 @@
...
+async function fetchMessageReactions<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>(channel: Channel<StreamChatGenerics>, messageId: string) {
+  const reactions: ReactionResponse<StreamChatGenerics>[] = [];
+  const limit = 300;
+  let offset = 0;
+  const reactionsLimit = MAX_MESSAGE_REACTIONS_TO_FETCH;
+  let lastPageSize = limit;
+
+  while (lastPageSize === limit && reactions.length < reactionsLimit) {
```
myandrienko, line 44: This implementation is gratefully adopted from our Angular SDK :)

src/components/Reactions/ReactionsList.tsx
```diff
@@ -30,113 +29,79 @@ export type ReactionsListProps<
...
+                  </button>
+                </li>
+              ),
+          )}
+          <li>
+            <span className='str-chat__reaction-list--counter'>{totalReactionCount}</span>
+          </li>
+        </ul>
+      </div>
+      <ReactionsListModal
```
myandrienko, line 96: 🍓

src/components/Message/utils.tsx
```diff
@@ -305,7 +305,7 @@ export const messageHasReactions = <
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
 >(
   message?: StreamMessage<StreamChatGenerics>,
-) => !!message?.latest_reactions && !!message.latest_reactions.length;
+) => Object.values(message?.reaction_counts ?? {}).some((count) => count > 0);
```
myandrienko, line 308: This check has never been strictly correct (we might have some reactions, but all of them may be unsupported) - but at least it's more in line with the behavior of the `ReactionsList` component.

src/components/Reactions/SimpleReactionsList.tsx
```diff
@@ -18,8 +18,6 @@ type WithTooltipProps = {
   title: React.ReactNode;
 };
 
-// todo: merge with ReactionsList/ButtonWithTooltip
-// avoiding breaking change of replacing <span> with <button>
 const WithTooltip = ({
```
myandrienko, line 21: No longer relevant, since `ReactionsList` doesn't use tooltips anymore.

src/components/Message/hooks/useReactionsFetcher.ts
```diff
@@ -0,0 +1,57 @@
...
+  };
+}
+
+async function fetchMessageReactions<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>(channel: Channel<StreamChatGenerics>, messageId: string) {
+  const reactions: ReactionResponse<StreamChatGenerics>[] = [];
+  const limit = 300;
+  let offset = 0;
+  const reactionsLimit = MAX_MESSAGE_REACTIONS_TO_FETCH;
```
MartinCupela, line 41: I am thinking whether integrators would want to customize this param.

src/components/Message/hooks/useReactionsFetcher.ts
MartinCupela: Would it make sense to cache the results in case somebody clicks a lot on those buttons that trigger the pagination over all the reactions up to `MAX_MESSAGE_REACTIONS_TO_FETCH`.

docusaurus/docs/React/hooks/message-hooks.mdx
```diff
@@ -179,6 +179,26 @@ const MyCustomMessageComponent = () => {
 };
 ```
 
+### useReactionsFetcher
+
+A [custom hook](https://github.com/GetStream/stream-chat-react/blob/master/src/components/Message/hooks/useReactionsFetcher.ts) to handle loading message reactions. Returns an async function that loads and returns message reactions.
```
MartinCupela, line 184: Would it make sense to mention there is a limit how many reactions are returned at max? (`MAX_MESSAGE_REACTIONS_TO_FETCH`)

docusaurus/docs/React/hooks/message-hooks.mdx
```diff
@@ -179,6 +179,26 @@ const MyCustomMessageComponent = () => {
...
+
+```jsx
+const MyCustomMessageComponent = () => {
+  const { message } = useMessageContext();
+  const { addNotification } = useChannelActionContext();
+  const handleFetchReactions = useReactionsFetcher(message, { notify: addNotification });
+  const { data } = useQuery(['message', message.id, 'reactions'], handleFetchReactions);
```
MartinCupela, line 191: It is not clear where `useQuery` comes from. Maybe we could add a comment above.

docusaurus/docs/React/hooks/message-hooks.mdx
```diff
@@ -179,6 +179,26 @@ const MyCustomMessageComponent = () => {
 };
 ```
 
+### useReactionsFetcher
+
+A [custom hook](https://github.com/GetStream/stream-chat-react/blob/master/src/components/Message/hooks/useReactionsFetcher.ts) to handle loading message reactions. Returns an async function that loads and returns message reactions.
+
+```jsx
+const MyCustomMessageComponent = () => {
```
MartinCupela, line 187: Maybe I would be showing how to override `ReactionsList` rather than a whole message.

src/components/Reactions/ReactionsList.tsx
MartinCupela: Would it make sense to make this customizable through component context? I am not sure as ReactionsList and ReactionsListModal do not contain complicated logic.

src/components/Message/hooks/useReactionsFetcher.ts
myandrienko: Increasing this limit can quickly cause rate limit errors. It's possible to do (`ReactionsList` accepts an override for `handleFetchReactions`), but I think it's a good thing that it's not too easy :)

docusaurus/docs/React/hooks/message-hooks.mdx
myandrienko: I don't think that's possible - AFAIK you cannot override a component used for `ReactionsList`, you have to provide a whole custom message component instead.

myandrienko: @MartinCupela I stand corrected, will update the doc.

docusaurus/docs/React/components/message-components/reactions.mdx
```diff
@@ -263,6 +286,14 @@ Additional props to be passed to the [`NimbleEmoji`](https://github.com/missive/
 | ------ |
 | object |
 
+### handleFetchReactions
+
+Function that loads all message reactions (overrides the function stored in `MessageContext`).
```
MartinCupela, line 291:
```suggestion
Function that loads the message reactions (overrides the function stored in `MessageContext`).
```

docusaurus/docs/React/components/message-components/message-ui.mdx
```diff
@@ -287,6 +287,16 @@ Function that edits a message (overrides the function stored in `MessageContext`
 | ----------------------------------------------------------- | ------------------------------------------------------------------------------- |
 | (event: React.BaseSyntheticEvent) => Promise<void\> \| void | [MessageContextValue['handleEdit']](../contexts/message-context.mdx#handleedit) |
 
+### handleFetchReactions
+
+Function that loads all reactions for a message (overrides the function stored in `MessageContext`).
```
MartinCupela, line 292:
```suggestion
Function that loads the reactions for a message (overrides the function stored in `MessageContext`).
```

docusaurus/docs/React/components/contexts/message-context.mdx
```diff
@@ -159,6 +159,16 @@ Function that edits a message.
 | ----------------------------------------------------------- |
 | (event: React.BaseSyntheticEvent) => Promise<void\> \| void |
 
+### handleFetchReactions
+
+Function that loads all reactions for a message.
```
MartinCupela, line 164:
```suggestion
Function that loads the reactions for a message.
```

src/components/Message/utils.tsx
```diff
@@ -306,6 +306,7 @@ export const messageHasReactions = <
 >(
   message?: StreamMessage<StreamChatGenerics>,
 ) => !!message?.latest_reactions && !!message.latest_reactions.length;
+// tktktk
```
arnautov-anton:
```suggestion
```

arnautov-anton (approved):

src/components/Message/utils.tsx
myandrienko: @arnautov-anton Looks like this is from an older commit :)

PR #2261, opened 2024-02-06
fix: show UnreadMessagesNotification when MessageList is scrolled to bottom via Element.scrollTo

MartinCupela:
### 🎯 Goal

Handle situation when `scrollToBottom` is called from another component when the msg list is scrolled above the observed target (unread separator).The intersection observer is not triggered when Element.scrollTo() is called and so we are left with old position information - being above the `UnreadMessagesSeparator`. So we end up in a situation when we are scrolled to the bottom and at the same time scrolled above the observed target.

### DIFF
--- a/src/components/MessageList/hooks/MessageList/useUnreadMessagesNotification.ts
+++ b/src/components/MessageList/hooks/MessageList/useUnreadMessagesNotification.ts
@@ -70,6 +70,7 @@ export const useUnreadMessagesNotification = ({
 
     if (unreadCount && isMessageListScrolledToBottom && isScrolledAboveTargetTop.current) {
       setShow(true);
+      isScrolledAboveTargetTop.current = false;
     }
   }, [isMessageListScrolledToBottom, unreadCount]);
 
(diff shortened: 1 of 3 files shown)


src/components/MessageList/hooks/MessageList/useUnreadMessagesNotification.ts
```diff
@@ -54,5 +61,17 @@ export const useUnreadMessagesNotification = ({
...
 
+  useEffect(() => {
+    /**
+     * Handle situation when scrollToBottom is called from another component when the msg list is scrolled above the observed target (unread separator).
+     * The intersection observer is not triggered when Element.scrollTo() is called. So we end up in a situation when we are scrolled to the bottom
+     * and at the same time scrolled above the observed target.
+     */
+
+    if (unreadCount && isMessageListScrolledToBottom && isScrolledAboveTargetTop.current) {
+      setShow(true);
```
myandrienko, line 72: Looks like this can be done safely outside of the effect. Plus, you'll save yourself one commit to DOM, which is always nice :)

Also, should we update `isScrolledAboveTargetTop.current = false` here, or am I misunderstanding the problem?

myandrienko (approved):

PR #2263, opened 2024-02-07
fix: avatar alignment

myandrienko:
### 🎯 Goal

Currently `` in the Avatar component is an inline-block element with line-height equal to the avatar size. Since the image is vertically aligned to the baseline, the actual height occupied by it is a bit more than the avatar size, with a little additional spacing below the baseline.

That doesn't affect the positioning, but can cause unexpected overflows. I encountered this when updating our demos:

https://github.com/GetStream/stream-chat-react/assets/975978/1ab31832-010c-49f6-8a27-13a0f89ad434

### 🛠 Implementation details

Fixing this just by making `` block - no need to deal with vertical alignment that way.

### 🎨 UI Changes

No visible changes.

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -80,6 +80,7 @@ export const Avatar = <
           onLoad={() => setLoaded(true)}
           src={image}
           style={{
+            display: 'block',
             flexBasis: `${size}px`,
             height: `${size}px`,
             objectFit: 'cover',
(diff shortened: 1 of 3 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -80,6 +80,7 @@ export const Avatar = <
           onLoad={() => setLoaded(true)}
           src={image}
           style={{
+            display: 'block',
```
oliverlaz, line 83: Why don't you add this to the `.str-chat__avatar-image` rule? It would be easier for a customer to override it in case they would have to.

myandrienko:
Superseded by https://github.com/GetStream/stream-chat-css/pull/267.

PR #2265, opened 2024-02-07
chore: bump react-scripts in typescript example

myandrienko:
### 🎯 Goal

Typescript app is our go-to example for React SDK, however it's currently incompatible with Node 18+ (which is a requirement for `stream-chat-react`).

### 🛠 Implementation details

Fixed by bumping react-scripts to latest.


oliverlaz (approved):

oliverlaz:
On the long run, we should switch to Vite :)

PR #2267, opened 2024-02-08
feat: keep unread channel UI when unread channel is marked read on mount

MartinCupela:
### 🎯 Goal

Change the default behavior when a channel is marked read. This PR introduces change that will mark channel read on mount, but at the same time will keep the original unread channel state from before when the channel was marked read. This copies the behavior of other chat applications.

### 🛠 Implementation details

The PR introduces new state variable `ownChannelReadState` that maintaines the channel read state for own user apart from the `channel.state.read` mapping. This allows us to keep the unread UI even though a channel was marked read on mount.


src/components/Channel/Channel.tsx
```diff
@@ -411,29 +415,42 @@ const ChannelInner = <
...
-          doMarkReadRequest(channel);
-        } else {
-          logChatPromiseExecution(channel.markRead(), 'mark read');
-        }
+        try {
+          if (doMarkReadRequest) {
+            doMarkReadRequest(channel, updateLocalReadState ? setOwnChannelReadState : undefined);
+          } else {
+            const markReadResponse = await channel.markRead();
+            console.log('markReadResponse', updateLocalReadState, markReadResponse);
```
MartinCupela, line 431: Remove this please

src/components/ChannelPreview/ChannelPreviewMessenger.tsx
```diff
@@ -71,8 +70,6 @@ const UnMemoizedChannelPreviewMessenger = <
     } else if (setActiveChannel) {
       // eslint-disable-next-line default-case
       switch (markActiveChannelReadOn) {
-        case MarkChannelReadOn.never:
-          break;
         case MarkChannelReadOn.leave:
           handleMarkReadOnLeave();
           break;
```
MartinCupela, line 75: change to if conditional

oliverlaz (approved):

PR #2278, opened 2024-02-14
fix: trap focus in opened modal

myandrienko:
### 🎯 Goal

When a modal dialog is displayed in chat, it's still possible to `Tab` your way through the focusable elements outside of the modal. The common practice is to "trap" the focus inside the modal, so that `Tab` cycles through the focusable elements inside the modal only.

### 🛠 Implementation details

Implementing focus trap properly is not as easy as it seems - so it's better to use a proven solution. This PR uses `FocusScope` from React Aria.

### 🎨 UI Changes

https://github.com/GetStream/stream-chat-react/assets/975978/a6ac10f0-b4a9-44b2-9beb-2a822ceb0ab2


package.json
```diff
@@ -157,10 +158,10 @@
     "@types/lodash.throttle": "^4.1.7",
     "@types/lodash.uniqby": "^4.7.7",
     "@types/moment": "^2.13.0",
-    "@types/react": "^18.0.8",
-    "@types/react-dom": "^18.0.3",
-    "@types/react-image-gallery": "^1.0.5",
-    "@types/react-is": "^17.0.3",
+    "@types/react": "^18.2.55",
```
myandrienko, line 161: React Aria packages use `React.JSX` in their typings, instead of the deprecated global `JSX`. That's why we needed to bump `@types/react` to the latest version in v18 branch: our previous version `@types/react@^18.0.3` didn't have `React.JSX` (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64464).

package.json
```diff
@@ -157,10 +158,10 @@
...
+    "@types/react-dom": "^18.2.19",
+    "@types/react-image-gallery": "^1.2.4",
+    "@types/react-is": "^18.2.4",
```
myandrienko, line 164: Bringing those up-to-date as well :)

oliverlaz (approved):

MartinCupela (approved):

arnautov-anton:
Without running this explicitly - is the exit button tabbable as well (top right corner) - would it be also expected behavior cosidering elements are on the same one visible layer?

myandrienko:
> Without running this explicitly - is the exit button tabbable as well (top right corner) - would it be also expected behavior cosidering elements are on the same one visible layer?

@arnautov-anton Yes, you can tab into the close button! I think this is the correct behavior, as you said: both the close button and the rest of the modal belong to the same "layer".

PR #2282, opened 2024-02-15
feat: make aria-labels localizable

### LINKED ISSUE #1931: Request MessageActions to allow MessageActionProps to accept custom aria label for button
**Motivation**
I'd like to request the ability to pass in a custom aria-label for the button found in 
https://github.com/GetStream/stream-chat-react/blob/03fbe2b5ab1f71d82880586bada6ed2cba156db4/src/components/MessageActions/MessageActions.tsx#L134

**Proposed solution**
Optional MessageActionProp that accepts a string that can be passed to alter button aria-label

**Acceptance Criteria**
- If prop is not passed in, then it defaults to "Open Message Actions Menu"
- If a custom aria-label prop string is passed in, the button takes that custom string

### LINKED ISSUE #1994: Allow Aria Label Configurability to improve Accessibility
**Motivation**
With accessibility being more of a focus lately, we have been looking to improve areas that are within our control in regards to GetStream. We have noticed there are a few concerns we'd like to address that are currently out of our hands as GetStream is hard-coding values/not allowing configurability in some of these areas. 

One thing we find ourselves needing to do is set the `aria-label` for things such as the Emoji and Upload buttons. Although you are able to define custom components for these icons, the parent contains hard-coded labels that can only be addressed by creating an entirely new MessageInput component. 
HardCodedAriaLabels

 
**Proposed solution**
This could just be an object that is taken in as a prop into Channel, with each key representing the label value you'd like to overwrite (ex. emojiPickerAriaLabel). This would allow us to define custom aria labels depending on user locale.

**Acceptance Criteria**
Please allow for configurability of the aria-labels for the controls/icons within the MessageInput.

myandrienko:
### 🎯 Goal

We had several requests related to customization of `aria-label` attributes. The best approach is to rely on existing internationalization support for customization and overrides.

Fixes #1931, fixes #1994.

### 🛠 Implementation details

1. Wrapped all ARIA labels in `t(...)` translation function. The keys used for ARIA labels are prefixed with `aria/...` to distinguish them from other texts.
2. Added default translations for supported languages.

### 🎨 UI Changes

No visible changes.

### To-Do

- [x] Update docs


i18next-parser.config.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   createOldCatalogs: false,
   input: ['./src/**/*.{tsx,ts}'],
   keepRemoved: true,
-  keySeparator: false,
+  keySeparator: '/',
```
myandrienko, line 6: Hopefully this will not conflict with any future translations, but just in case we can also use a more exotic symbol. Possible a backslash \?

src/mock-builders/translator.js
```diff
@@ -3,3 +3,9 @@ export const mockTranslatorFunction = (value, params = {}) =>
     const regexp = new RegExp(`\\{\\{\\s${key}\\s\\}\\}`, 'g');
     return acc.replace(regexp, params[key]);
   }, value);
+
+export const mockTranslationContext = {
+  // Mock translation function that always falls back to the key.
+  // It also handles nested keys, e.g. 'aria/Send' will fall back to 'Send'.
+  t: (key) => key.split('/').pop(),
```
myandrienko, line 10: We don't load translations in tests, so all translations fall back to the key. However, we don't want translations like `t('aria/Send')` to fall back to `'Send'` and not ~`'aria/Send'`~.

For test cases where that's an issue, `mockTranslationContext` can be provided.

src/i18n/Streami18n.ts
```diff
@@ -532,10 +537,10 @@ export class Streami18n {
...
           this.translations[this.currentLanguage] &&
           this.translations[this.currentLanguage][defaultNS]
-            ? {
-                ...this.translations[this.currentLanguage][defaultNS],
-                ...translationsForLanguage,
-              }
+            ? defaultsDeep(
+                translationsForLanguage,
+                this.translations[this.currentLanguage][defaultNS],
+              )
```
myandrienko, line 543: Translations JSON is now nested, so we need to merge in the defaults recursively.

i18next-parser.config.js
oliverlaz: If we go this way, I think we should find another, less commonly used symbol.
Another option is to keep the translations list flat (as today) and use `aria/*` or `aria.*` as a prefix. What do you think about this?

MartinCupela: I agree that separator should be a bit more crazy. Otherwise I approve the PR.

MartinCupela (approved):

i18next-parser.config.js
myandrienko: I like the idea of keeping the translations list flat. So I'm going to keep the `aria/` prefix, but disable the key separator in i18next 👍

src/i18n/Streami18n.ts
myandrienko: No longer relevant.

PR #2289, opened 2024-02-21
feat: add customizable reactions sorting

myandrienko:
### 🎯 Goal

Currently the order of reactions in the `ReactionsList` depends of the order of keys in the `reaction_counts` object returned by the backend. This leads to unstable reaction order, which can change on page reloads, or when the list of reactions changes.

We want to provide a way to customize this behavior, and to have a stable order of reactions by default.

### 🛠 Implementation details

Added a new `MessageContext` property: `sortReactions`. It accepts two reaction objects, and should return a positive/negative/zero numeric value, similar to the `Array.prototype.sort` method.

The `sortReactions` prop can be passed both to `MessageList` and `VirtualizedMessageList`, as well as to the `Message` component directly. It will then be available via the `MessageContext`.

### DIFF
--- a/src/components/Reactions/hooks/useProcessReactions.tsx
+++ b/src/components/Reactions/hooks/useProcessReactions.tsx
@@ -4,7 +4,7 @@ import { useComponentContext, useMessageContext } from '../../../context';
 
 import type { ReactionsListProps } from '../ReactionsList';
 import type { DefaultStreamChatGenerics } from '../../../types/types';
-import { ReactionSummary } from '../types';
+import type { ReactionsComparator, ReactionSummary } from '../types';
 
 type SharedReactionListProps =
   | 'own_reactions'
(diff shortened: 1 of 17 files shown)


src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -70,27 +79,32 @@ export const useProcessReactions = <
...
+            isOwnReaction: isOwnReaction(reactionType),
+            latestReactedUserNames: getLatestReactedUserNames(reactionType),
+            reactionCount,
+            reactionType,
+          },
+        ];
+      },
+    );
+
+    return unsortedReactions.sort(sortReactions);
```
myandrienko, line 100: 🍓

MartinCupela (approved):

PR #2290, opened 2024-02-21
feat: add customizable reaction details sorting

myandrienko:
This is a follow-up to 🚂 #2289. Same thing, but for the list of reacted users inside the reactions list modal.

### DIFF
--- a/src/components/Reactions/ReactionsListModal.tsx
+++ b/src/components/Reactions/ReactionsListModal.tsx
@@ -1,12 +1,13 @@
 import React, { useMemo } from 'react';
 import clsx from 'clsx';
 
+import type { ReactionDetailsComparator, ReactionSummary } from './types';
+
 import { Modal, ModalProps } from '../Modal';
-import { ReactionSummary } from './types';
 import { useFetchReactions } from './hooks/useFetchReactions';
 import { LoadingIndicator } from '../Loading';
 import { Avatar } from '../Avatar';
-import { MessageContextValue } from '../../context';
+import { MessageContextValue, useMessageContext } from '../../context';
 import { DefaultStreamChatGenerics } from '../../types/types';
 
 type ReactionsListModalProps<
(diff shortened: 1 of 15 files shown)


src/components/Reactions/__tests__/ReactionsListModal.test.js
```diff
@@ -180,4 +180,55 @@ describe('ReactionsListModal', () => {
...
+      sortReactionDetails: (a, b) => -a.user.name.localeCompare(b.user.name),
+    });
+
+    await act(() => {
+      fireEvent.click(getByTestId('reactions-list-button-haha'));
+    });
+
+    expect(
+      getByText('Mark Number 2').compareDocumentPosition(getByText('Mark Number 1')),
+    ).toStrictEqual(Node.DOCUMENT_POSITION_FOLLOWING);
```
arnautov-anton, line 228: Nice, this is an insteresting API I haven't seen before. TIL 🌮

src/components/Reactions/ReactionsListModal.tsx
```diff
@@ -16,13 +17,21 @@ type ReactionsListModalProps<
     reactions: ReactionSummary[];
     selectedReactionType: string | null;
     onSelectedReactionTypeChange?: (reactionType: string) => void;
+    sortReactionDetails?: ReactionDetailsComparator;
   };
 
+const defaultSortReactionDetails: ReactionDetailsComparator = (a, b) => {
+  const aName = a.user?.name ?? a.user?.id;
+  const bName = b.user?.name ?? b.user?.id;
+  return aName ? (bName ? aName.localeCompare(bName, 'en') : 1) : -1;
```
MartinCupela, line 26: Do I understand it well that if `bName` is `undefined, then the resulting order will be `[b,a]` because the return value is > 0?

myandrienko: Yes. But really, that's never going to happen, since we fallback to `user.id`, and `user`, although marked optional, is always defined.

Just a precaution, I've changed the order so that users with undefined names go to the bottom of the list, instead of to the top.

oliverlaz (approved):

PR #2292, opened 2024-02-22
fix: initial message load issues

myandrienko:
### 🎯 Goal

This PR covers a range of issues that originate from incorrect initialization of message list state. All the cases are covered below.

#### `ChannelList`, `Channel`, `MessageList`, `VirtualizedMessageList` not respecting the message limits

We have several default limits listed in the `limit.ts` file. However they were not always respected:

1. `MessageList`, `VirtualizedMessageList` didn't use `DEFAULT_NEXT_CHANNEL_PAGE_SIZE` and defaulted to the hard-coded message limit of 100.
2. `Channel` didn't use `DEFAULT_INITIAL_CHANNEL_PAGE_SIZE` when fetching the first page of messages.
3. `ChannelList` didn't use `DEFAULT_INITIAL_CHANNEL_PAGE_SIZE` as the message limit per channel when fetching channels.

This is now fixed.

#### False-negative `hasMore` when messages are fetched by `ChannelList`

We have two main ways to initialize the message list: it can be initialized by the `Channel` component; or it can be initialized by the `ChannelList` component (using `usePaginatedChannels` hook). If the message list is initialized by the `ChannelList`, the `Channel` component will not try to query the channel, and will use messages fetched by the `ChannelList` instead.

The problem here is that to determine if the channel has more messages to load, we used to compare the number of messages currently in the channel with the limit set on the `Channel`. But since it was not the `Channel` but the `ChannelList` that fetched the messages, this check makes no sense: `ChannelList` has its own separate limits.

Consider this example:

```tsx
const App = () => (
  <Chat client={chatClient}>
    <ChannelList options={{ message_limit: 5 }} />
    <Channel channelQueryOptions={{ members: { limit: 10 } }}>
      {/* ... */}
    </Channel>
  </Chat>
);
```

The `Channel` component will compare the number of messages fetched by the `ChannelList` (5) with its own limit (10) and determine that the channel has no more messages.

This is fixed by always setting `hasMore` to false in channel if it was not the `Channel` component that fetched the messages. In the worst case we'll just make one redundant query.

#### The "tall window" problem

In case the message container is very tall, or the message limit for the initial load is very small, we can run into a situation when the first page of messages is not long enough to trigger overflow in the message list. Without overflow, scroll events don't fire, so the next page of messages is never fetched.

This is not usually a problem in the `VirtualizedMessageList`, since it will immediately load the next page _once_ in this case. However, the `MessageList` didn't have this logic in place. It is now added (by invoking the `scrollListener` _once_ on mount).

**Note:** both `VirtualizedMessageList` and `MessageList` will only try loading the next page _once_ if the first page of messages is not long enough to cause overflow. A better solution would be to keep loading pages until there's overflow or there is no more pages. This solution is not currently possible to implement in `VirtualizedMessageList`, so for the sake of consistency I didn't implement it `MessageList` (event though it is possible).

In most cases this not an issue, since our default page size is 100 messages which is enough to fill even very tall containers.


src/components/Channel/Channel.tsx
```diff
@@ -366,6 +367,16 @@ const ChannelInner = <
...
 
+  const channelQueryOptions: ChannelQueryOptions<StreamChatGenerics> & {
+    messages: { limit: number };
+  } = useMemo(
+    () =>
+      defaultsDeep(propChannelQueryOptions, {
+        messages: { limit: DEFAULT_INITIAL_CHANNEL_PAGE_SIZE },
+      }),
+    [propChannelQueryOptions],
+  );
```
myandrienko, line 378: Use `DEFAULT_INITIAL_CHANNEL_PAGE_SIZE` when loading the first page of messages.

src/components/Channel/Channel.tsx
```diff
@@ -583,10 +596,12 @@ const ChannelInner = <
...
-          hasMore: hasMoreMessagesProbably(
-            channel.state.messages.length,
-            channelQueryOptions?.messages?.limit ?? DEFAULT_INITIAL_CHANNEL_PAGE_SIZE,
-          ),
+          hasMore:
+            channelInitializedExternally ||
+            hasMoreMessagesProbably(
+              channel.state.messages.length,
+              channelQueryOptions.messages.limit,
+            ),
```
myandrienko, line 604: If it wasn't us that loaded the first page of messages, don't trust the message count to determine if there's more messages to load, since the function that loaded messages could have totally different limits.

src/components/MessageList/MessageList.tsx
```diff
@@ -68,7 +69,7 @@ const MessageListWithContext = <
     headerPosition,
     read,
     renderMessages = defaultRenderMessages,
-    messageLimit = 100,
+    messageLimit = DEFAULT_NEXT_CHANNEL_PAGE_SIZE,
```
myandrienko, line 72: Remove hard-coded limit for message page size.

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -184,7 +185,7 @@ const VirtualizedMessageListWithContext = <
     loadMoreNewer,
     Message: MessageUIComponentFromProps,
     messageActions,
-    messageLimit = 100,
+    messageLimit = DEFAULT_NEXT_CHANNEL_PAGE_SIZE,
```
myandrienko, line 188: Remove hard-coded limit for message page size.

src/components/InfiniteScrollPaginator/InfiniteScroll.tsx
```diff
@@ -130,14 +123,17 @@ export const InfiniteScroll = (props: PropsWithChildren<InfiniteScrollProps>) =>
     const scrollElement = scrollComponent.current?.parentNode;
     if (!scrollElement) return;
 
+    const scrollListener = () => scrollListenerRef.current?.();
+
     scrollElement.addEventListener('scroll', scrollListener, useCapture);
     scrollElement.addEventListener('resize', scrollListener, useCapture);
+    scrollListener();
```
myandrienko, line 130: Note that we'll only load a single additional page of messages in case the first page is not enough to cause overflow of the message list.

To keep loading pages until there is overflow, or there is no more pages to load, we can do something like this:

```ts
  useLayoutEffect(() => {
    if (!isLoading) {
      scrollListenerRef.current?.();
    }
  }, [isLoading]);
```

However, I don't see a way to have the same behavior in the `VirtualizedMessageList`, and since our default page size is rather big (100 message), I think it's better to keep things consistent.

arnautov-anton:
Thank you, @myandrienko, closing #2275.

arnautov-anton (approved):

PR #2309, opened 2024-03-05
chore: revive e2e

myandrienko:
### 🎯 Goal

Bring our e2e tests to at least minimally useful state.

1. Added Playwright browser binaries to cache so that they are not downloaded each time.
2. Removed flaky tests.
3. Removed all browsers except for Chromium - hopefully temporary, but some tests were flaky particularly on WebKit.
4. Playwright will now stop after the first failed tests, so that tests don't hang for 30 mins.
5. Added the whole job to a single concurrency group, because two tests running at once on the same app will conflict.


MartinCupela (approved):

.github/workflows/e2e.yml
```diff
@@ -6,27 +6,33 @@ jobs:
...
       - uses: actions/checkout@v3
 
       - name: 💾 Cache Dependencies
         uses: actions/cache@v3
         with:
-          path: ./node_modules
+          path: |
+            ./node_modules
+            ./pw-browsers
           key: ${{ runner.os }}-${{ matrix.node }}-modules-${{ hashFiles('**/yarn.lock') }}
```
arnautov-anton, line 23:
```suggestion
          key: ${{ runner.os }}-${{ matrix.node }}-modules-${{ hashFiles('./yarn.lock') }}
```
Let's also get rid of this wildcard, I don't think we care about other lock files at this point in the pipeline.

.github/workflows/e2e.yml
```diff
@@ -6,27 +6,33 @@ jobs:
...
 
       - name: 🔨 Install Dependencies
-        run: yarn install --frozen-lockfile --ignore-engines --ignore-scripts
-
-      - name: ⚗️ End-to-end tests
         run: |
+          yarn install --frozen-lockfile --ignore-engines --ignore-scripts
           npx playwright install
```
arnautov-anton, line 28:
```suggestion
          npx playwright install chromium
```
Let's skip the installation of the other two as those are not needed (for the time being).

arnautov-anton: Other thing (consider it a nit) - would it be faster to move this command to its own step and running it only if the cache-hit is `false`?

.github/workflows/e2e.yml
myandrienko: Good point!

.github/workflows/e2e.yml
myandrienko: Playwright actually checks if binaries are already up-to-date and only downloads them as necessary. So adding the path with browser binaries to cache is enough.

PR #2315, opened 2024-03-07
fix: jump to first unread message when the last read and first unread message id is unknown

MartinCupela:
### 🎯 Goal

Fix the shortcoming in the `jumpToFirstUnreadMessage()` function's logic. This PR allows us to jump to the first unread message even though the the last read and first unread message ids are unknown. The `last_read` value is always available and so we can query the message by last read timestamp. 

This PR provides two more fixes:

- fix: render UnreadMessagesSeparator before the first message in channel
- fix: reset channel unread UI state on channel.truncated

### 🛠 Implementation details

The `jumpToFirstUnreadMessage` logic accounts for scenarios:

1. first unread message id is known - can be in local state / queried from the API
2. last read message id is know - can be in local state / queried from the API - the first unread message is then any message following the last read message
3. last read message can be retrieved by `last_read` date - can be in local state / queried from the API by message `created_at` date - the first unread message is then any message following the last read message


src/components/Channel/Channel.tsx
```diff
@@ -781,54 +785,113 @@ const ChannelInner = <
   const jumpToFirstUnreadMessage = useCallback(
     async (queryMessageLimit = 100) => {
```
arnautov-anton, line 786:
```suggestion
    async (queryMessageLimit = 100) => {
```
Let's move this default to a constant.

src/components/Channel/Channel.tsx
```diff
@@ -781,54 +785,113 @@ const ChannelInner = <
...
       if (!(client.user && channelUnreadUiState?.unread_messages)) return;
```
arnautov-anton, line 787:
```suggestion
      if (!channelUnreadUiState?.unread_messages) return;
```
Can we expect `client.user` to be always available at this point in time? I don't see it being used anywhere within the scope of the function so I feel like it'd be safe to remove it if it's not being used for type narrowing.

src/components/Channel/Channel.tsx
```diff
@@ -781,54 +785,113 @@ const ChannelInner = <
...
+      if (!firstUnreadMessageId) {
+        addNotification(t('Failed to jump to the first unread message'), 'error');
+        return;
+      }
+      if (!channelUnreadUiState.first_unread_message_id)
+        _setChannelUnreadUiState({
+          ...channelUnreadUiState,
+          first_unread_message_id: firstUnreadMessageId,
+          last_read_message_id: lastReadMessageId,
+        });
```
arnautov-anton, line 890:
```suggestion
        _setChannelUnreadUiState((currentChannelUnreadUiState) => ({
          ...currentChannelUnreadUiState,
          first_unread_message_id: firstUnreadMessageId,
          last_read_message_id: lastReadMessageId,
        }));
```

src/components/Channel/Channel.tsx
```diff
@@ -781,54 +785,113 @@ const ChannelInner = <
...
+          channelUnreadUiState.last_read,
+          channel.state.messages,
+          true,
+        );
+
+        if (lastReadMessage) {
+          firstUnreadMessageId = channel.state.messages[lastReadMessageIndex + 1]?.id;
+          isInCurrentMessageSet = !!firstUnreadMessageId;
+          lastReadMessageId = lastReadMessage.id;
+        } else {
```
arnautov-anton, line 813: In this block there's a dispatch for "setLoadingMore" but  no "loadMoreFinished" apart from the one in the `catch` block - is that expected? I'd add it to a `finally` block to have it run either way (success or error).

src/components/Channel/Channel.tsx
MartinCupela: `currentChannelUnreadUiState` could be undefined according to Typescript so we would have to provide default for `last_read`.

src/components/Channel/Channel.tsx
MartinCupela: cannot use finally as hasMoreMessages is updated later on. I have adjusted the code so that it calls `loadMoreFinished` for all the paths in the given condition branch.

src/components/Channel/Channel.tsx
```diff
@@ -781,54 +785,113 @@ const ChannelInner = <
...
+                'new',
+              )
+            ).messages;
+          } catch (e) {
+            addNotification(t('Failed to jump to the first unread message'), 'error');
+            loadMoreFinished(hasMoreMessages, channel.state.messages);
+            return;
+          }
+          const firstMessageWithCreationDate = messages.find((msg) => msg.created_at);
+          if (!(messages.length && firstMessageWithCreationDate)) {
```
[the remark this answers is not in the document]
MartinCupela: It is important to check, whether some messages were returned here. If not, then it is unexpected and error notification shown.

src/components/Channel/Channel.tsx
MartinCupela: Added constant + to other default params in the Channel.tsx file.

src/components/Channel/Channel.tsx
MartinCupela: Removed client as dependency from the memoized function.

src/components/Channel/Channel.tsx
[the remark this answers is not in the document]
arnautov-anton: But at the same time - if the messages are empty, `firstMessageWithCreationDate` would be `undefined` thus checking messages seems unnecessary.

src/components/Channel/Channel.tsx
arnautov-anton: We could do something like this:
```tsx
_setChannelUnreadUiState((currentChannelUnreadUiState = DEFAULT) => ({
```
Or - set a default when the state is initialized - depends though. Not that important either way but this feels "cleaner", feel free to ignore. 👍

MartinCupela: I do not have resonabled default for `last_read`.

PR #2325, opened 2024-03-14
chore: streamline build pipeline

myandrienko:
### 🎯 Goal

Our build pipeline contained some quirks and the remains of previous setups. It was time to clean it up :)

### 🛠 Implementation details

The main changes are:

**No rollup.** Previously our build pipeline was implemented in rollup. To simplify things up, we now just run a couple of processes in parallel: tsc to build our normal distribution and type declarations, esbuild to build bundles, and a shell script to include assets in the distribution.

For development, just `tsc --watch` is enough.

**No babel.** Previously our rollup config included babel transpilation with babel-preset-env. That means we included babel-runtime in our distribution, and transpiled async functions into generators with the notorious `regenerator-runtime`.

Babel is now excluded from the build process (but still used as a parser for JS files in eslint). Targeting ES2020 in tsc and esbuild should be enough. And we finally have regular async functions in our distribution :)

**Esbuild for bundles.** We now use esbuild instead of rollup for creating a CJS bundles. And we no longer create UDM browser bundles.

**TS5.** I used this opportunity to bump TypeScript version as well :)

Also, removed webpack and postcss dependencies - they were not used at all.

### Some ideas for the future

1. Bump prettier, eslint and eslint plugin versions. Will result in *a lot* of warnings, but good for future-proofing.
2. Review our eslintrc, it's currently a bit of a mess.
3. Remove babel completely, use typescript-eslint instead.
4. Switch to `"module": "NodeNext"` in tsconfig. This is the new recommended option for libraries, but will require us to add extensions to all of our import statements, to make them fully qualified.
5. (breaking change) With `"module": "NodeNext"` we will be ready to add `"type": "module"` in our package.json. We can't do that now, because fully specified import statements are expected from module packages.
6. (breaking change) Drop all bundles. CJS bundle is kinda weird anyway. Browser bundle can be replaced by `<script type="module">`.


package.json
```diff
@@ -208,49 +204,33 @@
...
     "coverage": "jest --collectCoverage && codecov",
     "eslint": "eslint '**/*.{js,md,ts,jsx,tsx}' --max-warnings 0",
     "lint": "prettier --list-different 'src/**/*.{js,ts,tsx,md,json}' .eslintrc.json .prettierrc babel.config.js && eslint 'src/**/*.{js,ts,tsx,md}' --max-warnings 0 && yarn validate-translations",
     "lint-fix": "prettier --write 'src/**/*.{js,ts,tsx,md,json}' .eslintrc.json .prettierrc babel.config.js && eslint --fix 'src/**/*.{js,ts,tsx,md}' --max-warnings 0",
     "prettier": "prettier --list-different '**/*.{js,ts,tsx,md,json}' .eslintrc.json .prettierrc babel.config.js",
     "prettier-fix": "prettier --write '**/*.{js,ts,tsx,md,json}' .eslintrc.json .prettierrc babel.config.js",
     "fix-staged": "lint-staged --config .lintstagedrc.fix.json --concurrent 1",
-    "start": "tsc --watch",
-    "prepare": "husky install && yarn run build",
```
myandrienko, line 250: Previously we always built the package on post-install. That was not always needed, and slowed down our CI significantly.

Removing build from `prepare` script, and adding `yarn run build` in CI where really necessary.

package.json
```diff
@@ -208,49 +204,33 @@
...
+    "start": "tsc --watch --declarationMap",
```
myandrienko, line 229: In "dev mode" it's nice to have declaration maps, so that clicking on types in example apps takes you straight to the source TS file (instead of `.d.ts` from dist).

scripts/copy-version.sh
myandrienko, line 1: This file was extracted from inline script in package.json for readability.

scripts/copy-css.sh
myandrienko, line 1: This script replaces rollup-plugin-copy, it copies CSS from stream-chat-css to the distribution.

Also useful when debugging styles - no need to rebundle the whole distribution, just run:

```sh
./scripts/copy-css.sh
```

src/index_UMD.ts
```diff
@@ -6,3 +6,19 @@ export * from './i18n';
...
+    logChatPromiseExecution: typeof import('stream-chat').logChatPromiseExecution;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    StreamChat: typeof import('stream-chat') & { StreamChat: any };
+  }
+}
+
+window.StreamChat.StreamChat = window.StreamChat;
+window.StreamChat.logChatPromiseExecution = window.logChatPromiseExecution;
+window.StreamChat.Channel = window.Channel;
+window.ICAL = window.ICAL || {};
```
myandrienko, line 24: This used to be done in rollup. Since this fixup is only required for the browser bundle, it makes sense to just add it to bundle entrypoint. This file is not used anywhere else.

BTW, let me know if you have any idea what is ICAL.

tsconfig.json
```diff
@@ -1,35 +1,35 @@
...
-    "lib": ["ESNext", "DOM"],
-    "module": "ESNext",
+    "module": "ES2020",
+    "esModuleInterop": true,
     "moduleResolution": "Node",
+    "moduleDetection": "force",
+    "allowSyntheticDefaultImports": true,
+    "isolatedModules": true,
+    "allowJs": true,
+    "resolveJsonModule": true,
```
myandrienko, line 14: No big changes here, except aligning everything to ES2020.

package.json
```diff
@@ -208,49 +204,33 @@
...
     "stream-chat": "^8.21.0",
     "style-loader": "^2.0.0",
     "ts-jest": "^28.0.8",
-    "typescript": "^4.7.4",
-    "url-loader": "^4.1.1",
-    "webpack": "4.44.2",
-    "webpack-cli": "^3.3.12",
-    "webpack-dev-server": "~3.11.0"
+    "typescript": "^5.4.2",
+    "url-loader": "^4.1.1"
```
oliverlaz, line 216: `url-loader` can go away as well :)

package.json
```diff
@@ -208,49 +204,33 @@
...
-    "rollup-plugin-copy": "^3.4.0",
-    "rollup-plugin-insert": "^1.3.1",
-    "rollup-plugin-node-globals": "^1.4.0",
-    "rollup-plugin-peer-deps-external": "^2.2.4",
-    "rollup-plugin-terser": "^7.0.2",
-    "rollup-plugin-url": "^3.0.1",
-    "rollup-plugin-visualizer": "^4.2.0",
     "semantic-release": "^19.0.5",
     "stream-chat": "^8.21.0",
     "style-loader": "^2.0.0",
```
oliverlaz, line 213: `style-loader` too :)

rollup.config.js
```diff
@@ -1,158 +0,0 @@
...
-  'react-virtuoso',
-  'textarea-caret',
-  /uuid/,
-];
-
-const basePlugins = ({ useBrowserResolve = false }) => [
-  replace({
-    preventAssignment: true,
-    'process.env.NODE_ENV': JSON.stringify('production'),
-  }),
```
oliverlaz, line 69: Are we missing this in the esbuild pipeline?

package.json
oliverlaz: I think every `*-loader` package should be removed, as they are typically used with weback.

rollup.config.js
arnautov-anton: I think that is being set by default when building with `esbuild`:
https://esbuild.github.io/api/#platform

oliverlaz: Nice! TIL :)

package.json
myandrienko: All gone!

scripts/copy-css.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+mkdir -p dist/assets dist/css/v2 dist/scss/v2
+
+cp -r node_modules/@stream-io/stream-chat-css/dist/assets/* dist/assets
+cp -r node_modules/@stream-io/stream-chat-css/dist/css/*.css dist/css
```
MartinCupela, line 6: The `master` branch does not count with legacy styles anymore. Only `dist/v2/*` will be available.

arnautov-anton (approved):

PR #2339, opened 2024-03-22
feat: async voice messages recording

MartinCupela:
### 🎯 Goal

Introduce 
- media recording capability focused on recording audio at first.
- possibility to add any attachment to the MessageInput state
- audio transcoding to MP3

### 🛠 Implementation details

### 🎨 UI Changes

Please see the attached documentation files.


src/components/Attachment/components/WaveProgressBar.tsx
```diff
@@ -55,10 +69,14 @@ export const WaveProgressBar = ({
...
       data-testid='wave-progress-bar-track'
       onClick={seek}
-      onMouseDown={handleMouseDown}
+      onMouseDown={handleDragStart}
       onMouseLeave={handleMouseUp}
       onMouseMove={handleMouseMove}
       onMouseUp={handleMouseUp}
+      onTouchEnd={handleMouseUp}
+      onTouchMove={handleTouchMove}
+      onTouchStart={handleDragStart}
```
myandrienko, line 78: Don't want to send you down the rabbit hole, but if you want - check out pointer events instead of mouse and touch. Having a single interface might help to reduce code duplication.

src/components/Attachment/hooks/useAudioController.ts
```diff
@@ -56,8 +111,8 @@ export const useAudioController = ({
       const currentTime = ratio * audioRef.current.duration;
       setSecondsElapsed(currentTime);
       audioRef.current.currentTime = currentTime;
-    },
-    [],
+    }, 60),
```
myandrienko, line 114: Having this function throttled means that the `secondsElapsed` value only updates about 16 times per seconds. That means that the dragging interaction on the progress bar will feel laggy.

Do you think it's possible to either:
1. Separate the displayed progress (which will update optimistically) and the actual progress (which will update with a throttle)
2. Reduce the throttle duration to at least 16ms (for 60fps)
3. Or just remove throttling here?

src/components/MessageInput/AttachmentPreviewList.tsx
```diff
@@ -28,6 +65,113 @@ export const AttachmentPreviewList = () => {
...
+    >
+      <audio ref={audioRef}>
+        <source data-testid='audio-source' src={attachment.asset_url} type={attachment.mime_type} />
+      </audio>
+      <PlayButton isPlaying={isPlaying} onClick={togglePlay} />
+
+      <button
+        className='str-chat__attachment-preview-delete'
+        data-testid='file-preview-item-delete-button'
+        disabled={attachment.$internal?.uploadState === AttachmentUploadState.UPLOADING}
```
myandrienko, line 94: Can you explain this `$internal` stuff? Is it something voice message related? I haven't seen it for attachments before.

src/components/MessageInput/hooks/useTimer.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
+
+export const useTimer = ({ startFrom }: { startFrom: number }) => {
+  const [secondsLeft, setSecondsLeft] = useState<number | undefined>();
+
+  useEffect(() => {
+    let countdownTimeout: ReturnType<typeof setTimeout>;
+    if (typeof secondsLeft === 'number' && secondsLeft > 0) {
+      countdownTimeout = setTimeout(() => {
+        setSecondsLeft(secondsLeft - 1);
```
myandrienko, line 10: You can use the callback variant of setState and remove effect dependency.

src/components/MessageInput/hooks/useTimer.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
+
+export const useTimer = ({ startFrom }: { startFrom: number }) => {
```
myandrienko, line 3: Can you check if this can be reused in `CooldownTimer` component? Not necessarily in this PR :)

src/components/MessageInput/types.ts
```diff
@@ -3,6 +3,12 @@ import type { DefaultStreamChatGenerics } from '../../types/types';
 
 type AttachmentLoadingState = 'uploading' | 'finished' | 'failed';
 
+export enum AttachmentUploadState {
```
myandrienko, line 6: Is this different from `UploadState`?

MartinCupela:
@myandrienko thank you for the review. I forgot to put this into draft state when creating this PR. I plan to clean up the stuff still 🙂

src/components/MessageInput/hooks/useTimer.ts
MartinCupela: This code has been extract from CooldownTimer and imported there :)

src/components/MessageInput/hooks/useTimer.ts
MartinCupela: How would I then decide to prevent setting timeout, when the value is 0? I understand that the author wanted to avoid setInterval due to throttling. Recursion with setTimeout may be a problem.

src/components/MessageInput/AttachmentPreviewList.tsx
MartinCupela: This are internal metadata that are not shared with the back-end. I decided to keep them "isolated" under this key and just remove them before sending the message with the attachment to the back-end. Current implementation with image and file uploads seems to me clunky in that it does not isolate these locally important metadata and even use different names for those attributes, that are later removed before the submission. 

I would like to propose this approach also to other types of attachments gradually to make our code more readable.

src/components/MessageInput/types.ts
MartinCupela: I wanted to gradually replace the union type with enum that could be used by the integrators too - if they want to do some customizations. Removing the original would be breaking. WDYT, how could this be approached?

src/components/Attachment/hooks/useAudioController.ts
MartinCupela: I decided to set it to 60Hz :)

src/components/Attachment/components/WaveProgressBar.tsx
MartinCupela: Thank you for the idea Matvei. I have made changes replacing those event listeners for pointer events listeners 👍

src/components/MessageInput/hooks/useTimer.ts
myandrienko: Ah, I see, I haven't noticed that `secondsLeft` is used outside of the state update. We can keep it like this then :)

src/components/MessageInput/AttachmentPreviewList.tsx
myandrienko: I see, here's what worries me: integrators implementing custom attachment previews shouldn't care about what is and what isn't part of our backend schema. But in this case they will have to use this `$internal` property, and they may find it a bit weird.

src/components/MessageInput/types.ts
myandrienko: I actually don't know, the union type doesn't sound bad, can we just keep using the union in this case?

src/components/MediaRecorder/classes/MediaRecorderController.ts
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    if (this.permission.state.value === 'denied') {
+      logError(new Error('Permission denied'));
+      return;
+    }
+
+    try {
+      const stream = await navigator.mediaDevices.getUserMedia(
+        this.mediaType === 'video' ? { audio: true, video: true } : { audio: true },
```
oliverlaz, line 286: Why do we support/need video recordings?

src/components/MediaRecorder/classes/MediaRecorderController.ts
```diff
@@ -0,0 +1,353 @@
...
+  AmplitudeRecorder,
+  AmplitudeRecorderConfig,
+} from './AmplitudeRecorder';
+
+import type { LocalVoiceRecordingAttachment } from '../../MessageInput';
+
+const RECORDED_MIME_TYPE_BY_BROWSER = {
+  audio: {
+    others: 'audio/webm',
+    safari: 'audio/mp4;codecs=mp4a.40.2',
```
oliverlaz, line 29: Do we convert mp4/aac to mp3 as well?

src/components/MediaRecorder/transcode/audioProcessing.ts
```diff
@@ -0,0 +1,35 @@
+import { readFileAsArrayBuffer } from '../../ReactFileUtilities';
+
+/**
+ * In the context of resampling audio data, AudioContext is used to decode the input audio file into an AudioBuffer,
+ * which is a fundamental data structure representing audio data.
+ * @param file
+ */
+export const toAudioBuffer = async (file: File) => {
+  const audioCtx = new AudioContext();
```
oliverlaz, line 9: Should we close this context eventually? Does this create a leak?

src/components/MediaRecorder/classes/MediaRecorderController.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    this.handleErrorEvent = this.handleErrorEvent.bind(this);
+    this.handleDataavailableEvent = this.handleDataavailableEvent.bind(this);
+    this.resetRecordingState = this.resetRecordingState.bind(this);
+    this.cleanUp = this.cleanUp.bind(this);
+    this.start = this.start.bind(this);
+    this.pause = this.pause.bind(this);
+    this.resume = this.resume.bind(this);
+    this.stop = this.stop.bind(this);
+    this.cancel = this.cancel.bind(this);
```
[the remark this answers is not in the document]
MartinCupela: thank you fixed

src/components/MediaRecorder/classes/MediaRecorderController.ts
MartinCupela: true, that is unnecessary

src/components/MediaRecorder/classes/MediaRecorderController.ts
MartinCupela: I will throw an error if video MIME type is passed.

src/components/MediaRecorder/classes/MediaRecorderController.ts
MartinCupela: https://github.com/GetStream/stream-chat-react/pull/2339/commits/f78ef717ce8d34113572dd3aa8ba373fdb7be682

src/components/MediaRecorder/transcode/audioProcessing.ts
MartinCupela: Fixed by closing the context: https://github.com/GetStream/stream-chat-react/pull/2339/commits/dab90e282f0c14656d64191e05c1175189b19d8a

src/components/MediaRecorder/classes/MediaRecorderController.ts
MartinCupela: https://github.com/GetStream/stream-chat-react/pull/2339/commits/3577a09da5384ac83bee2aeefd21e538ecd9b3c9

src/components/MediaRecorder/transcode/audioProcessing.ts
oliverlaz: Check this: https://github.com/GetStream/stream-chat-react/commit/dab90e282f0c14656d64191e05c1175189b19d8a#comments

docusaurus/docs/React/components/message-input-components/audio-recorder.mdx
```diff
@@ -0,0 +1,173 @@
...
+**3. Transcoding configuration**
+
+As per the mentioned problem with playback incompatibility among the browsers, the audio recorder has to transcode (convert) the original recording to a different format that can be reproduced in the majority of the browsers on majority of the OS, e.g. `audio/webm` to `audio/mp3`.
+
+The supported transcoded MIME types are
+
+| MIME                  | Description                                    |
+| --------------------- | ---------------------------------------------- |
+| `audio/mp3` (default) | Lossy audio format leading to smaller files.   |
+| `audio/wav` (default) | Lossless audio format leading to bigger files. |
```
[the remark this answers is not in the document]
MartinCupela: Removed

docusaurus/docs/React/components/message-input-components/audio-recorder.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+**2. `MediaRecorder` configuration**
+
+At the moment the configuration object allows to change the default recording MIME type. The browser landscape does not provide a unified support to recording and playback formats. Specific browsers support limited array of recording formats (Chrome only `audio/webm`) and not all the formats recorded in one browser can be played back in another browser (`audio/webm` is not supported on native iOS apps, iOS 17.4 and lower). The default recording MIME type configuration takes care of this problem and should not be changed unless we want to record only for a limited set of browsers.
+
+**3. Transcoding configuration**
+
+As per the mentioned problem with playback incompatibility among the browsers, the audio recorder has to transcode (convert) the original recording to a different format that can be reproduced in the majority of the browsers on majority of the OS, e.g. `audio/webm` to `audio/mp3`.
```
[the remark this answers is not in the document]
MartinCupela: Removed

oliverlaz (approved):

PR #2343, opened 2024-03-26
fix: suggestion onClickHandler type

myandrienko:
Noticed that an example in our docs is broken. Turns out, we provided wrong typings for the `onClickHandler` in `ChatAutoComplete`.

The reason it passed type checks before is that the suggestion list component is in a JSX file (not TSX). I'm not going to switch it TSX right now, but the types are worth fixing quickly.

Also removed outdated demos that will break because of this change.


src/components/ChatAutoComplete/ChatAutoComplete.tsx
```diff
@@ -24,31 +24,32 @@ export type SuggestionUser<
...
   key: React.Key;
-  onClickHandler: (event: React.BaseSyntheticEvent) => void;
-  onSelectHandler: (
-    item:
-      | Awaited<ReturnType<EmojiSearchIndex<T>['search']>>
-      | SuggestionUser<StreamChatGenerics>
-      | SuggestionCommand<StreamChatGenerics>,
+  onClickHandler: (
+    event: React.BaseSyntheticEvent,
+    item: SuggestionItem<StreamChatGenerics, T>,
```
myandrienko, line 50: 🍒 This second parameter was missing

examples/website-demos/README.md
```diff
@@ -0,0 +1 @@
+The demos have a new home. Check out the [website-react-examples](https://github.com/GetStream/website-react-examples) repository for the latest React sample apps.
```
myandrienko, line 1: Adding these READMEs to keep the links to old demos working.

arnautov-anton (approved):

PR #2350, opened 2024-04-02
chore(deps): drop @react-aria/focus version

arnautov-anton:
### 🎯 Goal

Drop the version of the `@react-aria/focus` package.


package.json
```diff
@@ -62,7 +62,7 @@
     "@babel/runtime": "^7.23.6",
     "@braintree/sanitize-url": "^6.0.4",
     "@popperjs/core": "^2.11.5",
-    "@react-aria/focus": "^3.16.1",
+    "@react-aria/focus": "^3.12.1",
```
myandrienko, line 65: We can do just `^3` if you want.

myandrienko (approved):

package.json
arnautov-anton: Seems a bit too broad but if the `FocusScope` thing works exactly the same then I'm all for it.

oliverlaz (approved):

PR #2358, opened 2024-04-08
fix: make setChannels prop optional for ChannelSearch

myandrienko:
### 🎯 Goal

`setChannels` is used by the `ChannelSearch` component to update the channel list state to include the selected search result. When rendered by the `ChannelList`, it actually passes its own `setChannels` setter there.

But this behavior should be optional, because it's totally valid to use search outside of the `ChannelList`.


MartinCupela:
@myandrienko why is it important to make this optional for the hook. Do we want to make this hook publicly available?

myandrienko:
> @myandrienko why is it important to make this optional for the hook. Do we want to make this hook publicly available?

@MartinCupela the hook `useChannelSearch` is not public. However, this type is also used for props of the `ChannelSearch` component which is public.

Currently, if you want to use `ChannelSearch` outside of `ChannelList`, you have to provide the `setChannels` setter somehow, even though it doesn't always makes sense. That makes the standalone use of `ChannelSearch` unnecessarily confusing.

oliverlaz (approved):

MartinCupela (approved):

PR #2359, opened 2024-04-09
fix: search results list props

myandrienko:
Small fixups for search results list props:

1. `SearchResultItem` is not an optional prop is practice (it's defaulted in the parent component). Having it optional makes it hard to override `SearchResultsList`. (Technically a breaking change, but let's consider it a fix instead.)
2. `| []` is not required 😅 

## Migration

The only place realistically that may require updates is the custom `SearchResultsList` component.

Before:

```tsx
const CustomSearchResultsList = (props: SearchResultsListProps) => {
  const { results, SearchResultItem = DefaultSearchResultItem } = props;
  return /* ... */
};

<ChannelSearch SearchResultsList={CustomSearchResultsList} />
```

After:

```tsx
const CustomSearchResultsList = (props: SearchResultsListProps) => {
                              // 👇 
  const { results, SearchResultItem } = props;
  return /* ... */
};
```

The change is not strictly required, but may cause linter errors otherwise.

### DIFF
--- a/src/components/ChannelSearch/SearchResults.tsx
+++ b/src/components/ChannelSearch/SearchResults.tsx
@@ -44,9 +44,8 @@ const DefaultSearchResultsHeader = <
 
 export type SearchResultsListProps<
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
-> = Pick<
-  SearchResultsProps<StreamChatGenerics>,
-  'results' | 'SearchResultItem' | 'selectResult'
+> = Required<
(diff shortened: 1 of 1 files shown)


oliverlaz (commented):
For sake of completeness, can you provide instructions in the PR on what needs to be done in case someone had this prop considered as optional?
Before and After code is good.

myandrienko:
> For sake of completeness, can you provide instructions in the PR on what needs to be done in case someone had this prop considered as optional? Before and After code is good.

@oliverlaz Updated PR description.

oliverlaz (approved):

PR #2370, opened 2024-04-23
feat: support reaction groups

myandrienko:
### 🎯 Goal

Adds support for a new `reaction_groups` property on message, that contains more in-depth summary of message reactions than the soon-to-be-deprecated `reaction_counts`.

### 🛠 Implementation details

The new property allows to implement a Slack-like sorting order for reactions (sorted by first reaction timestamp) which is now the default.

### To-Do:

- [x] Bump LLC

### DIFF
--- a/src/components/Message/hooks/useReactionHandler.ts
+++ b/src/components/Message/hooks/useReactionHandler.ts
@@ -31,17 +31,28 @@ export const useReactionHandler = <
       reaction: ReactionResponse<StreamChatGenerics>,
       message: StreamMessage<StreamChatGenerics>,
     ): StreamMessage<StreamChatGenerics> => {
-      const newReactionCounts = message?.reaction_counts || {};
+      const newReactionGroups = message?.reaction_groups || {};
       const reactionType = reaction.type;
-      const hasReaction = !!newReactionCounts[reactionType];
+      const hasReaction = !!newReactionGroups[reactionType];
 
       if (add) {
-        newReactionCounts[reactionType] = hasReaction ? newReactionCounts[reactionType] + 1 : 1;
+        const timestamp = new Date().toISOString();
+        newReactionGroups[reactionType] = hasReaction
+          ? { ...newReactionGroups[reactionType], count: newReactionGroups[reactionType].count + 1 }
+          : {
+              count: 1,
+              first_reaction_at: timestamp,
+              last_reaction_at: timestamp,
+              sum_scores: 1,
+            };
       } else {
-        if (hasReaction && newReactionCounts[reactionType] > 1) {
-          newReactionCounts[reactionType]--;
+        if (hasReaction && newReactionGroups[reactionType].count > 1) {
+          newReactionGroups[reactionType] = {
+            ...newReactionGroups[reactionType],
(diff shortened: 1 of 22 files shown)


src/components/Message/hooks/useReactionsFetcher.ts
```diff
@@ -32,25 +32,24 @@ export function useReactionsFetcher<
...
-    }
-    offset += lastPageSize;
+  while (hasNext) {
+    const page = await client.queryReactions(
+      messageId,
+      reactionType ? { type: reactionType } : {},
+      { created_at: -1 },
+      { limit, next },
+    );
+    reactions.push(...page.reactions);
```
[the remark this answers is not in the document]
oliverlaz: >This code is transpiled down to: `reactions.push.apply(reactions, page.reactions)`

Ok, we should be good then!

src/components/Message/hooks/useReactionHandler.ts
```diff
@@ -31,17 +31,28 @@ export const useReactionHandler = <
...
-        newReactionCounts[reactionType] = hasReaction ? newReactionCounts[reactionType] + 1 : 1;
+        const timestamp = new Date().toISOString();
+        newReactionGroups[reactionType] = hasReaction
+          ? { ...newReactionGroups[reactionType], count: newReactionGroups[reactionType].count + 1 }
+          : {
+              count: 1,
+              first_reaction_at: timestamp,
+              last_reaction_at: timestamp,
+              sum_scores: 1,
+            };
```
myandrienko, line 47: Optimistic updates for reaction groups. Not technically correct (because score defaults to 1, which might not be the case), but no more incorrect than an existing implementation ;)

src/components/Reactions/hooks/useProcessReactions.tsx
```diff
@@ -9,15 +9,21 @@ import type { ReactionsComparator, ReactionSummary } from '../types';
...
 
-export const defaultReactionsSort: ReactionsComparator = (a, b) =>
-  a.reactionType.localeCompare(b.reactionType, 'en');
+export const defaultReactionsSort: ReactionsComparator = (a, b) => {
+  if (a.firstReactionAt && b.firstReactionAt) {
+    return +a.firstReactionAt - +b.firstReactionAt;
+  }
+
+  return a.reactionType.localeCompare(b.reactionType, 'en');
+};
```
myandrienko, line 26: New default reaction order is chronological, sorted by first time someone reacted. Falls back to alphabetical for older messages without reaction groups.

src/components/Reactions/types.ts
```diff
@@ -3,10 +3,13 @@ import type { ReactionResponse } from 'stream-chat';
 
 export interface ReactionSummary {
   EmojiComponent: ComponentType | null;
+  firstReactionAt: Date | null;
   isOwnReaction: boolean;
+  lastReactionAt: Date | null;
   latestReactedUserNames: string[];
   reactionCount: number;
   reactionType: string;
+  unlistedReactedUserCount: number;
```
myandrienko, line 12: Just useful in case someone wants to implement "John, Jane, and 5 others" style tooltips.

PR #2373, opened 2024-04-25
test: add AudioRecorder tests

MartinCupela:
### 🎯 Goal

Add missing tests from  #2339

### DIFF
--- a/src/components/MediaRecorder/AudioRecorder/AudioRecorder.tsx
+++ b/src/components/MediaRecorder/AudioRecorder/AudioRecorder.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { AudioRecordingPreview } from './AudioRecordingPreview';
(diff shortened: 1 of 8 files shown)


src/components/MediaRecorder/AudioRecorder/AudioRecorder.tsx
```diff
@@ -19,47 +19,58 @@ export const AudioRecorder = () => {
...
   const isUploadingFile = recording?.$internal?.uploadState === 'uploading';
 
+  const state = useMemo(
+    () => ({
+      paused: recordingState === MediaRecordingState.PAUSED,
+      recording: recordingState === MediaRecordingState.RECORDING,
+      stopped: recordingState === MediaRecordingState.STOPPED,
+    }),
+    [recordingState],
+  );
```
oliverlaz, line 29: I think there is no need to memoize this. Inlining these values should be simpler:
```ts
const paused = recordingState === MediaRecordingState.PAUSED;
...
```

oliverlaz (approved):

PR #2399, opened 2024-05-23
fix: prevent loading more non-existent thread replies

MartinCupela:
### 🎯 Goal

VML behavior introduced in #2242 caused that the virtualized Thread would infinitely load more thread replies for a message that has no replies.


arnautov-anton (approved):
Oh - my bad, thanks, Martin!

PR #2400, opened 2024-05-23
chore: add demo of LLC without SCG

myandrienko:
Start here: https://github.com/GetStream/stream-chat-js/pull/1309

To try this demo locally, grab the `no-scg-demo` branch from `stream-chat-js` and link it locally.

### DIFF
--- a/examples/typescript/src/App.tsx
+++ b/examples/typescript/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { ChannelFilters, ChannelOptions, ChannelSort, StreamChat, UR } from 'stream-chat';
 import {
   Chat,
@@ -9,9 +9,20 @@ import {
   MessageInput,
   Thread,
   Window,
+  useCreateChatClient,
 } from 'stream-chat-react';
 
 import './App.css';
+
+declare module 'stream-chat' {
+  interface UserEx {
+    customUserField: string;
+  }
+
+  interface ReactionEx {
+    customReactionField: string;
+  }
+}
 
 const apiKey = process.env.REACT_APP_STREAM_KEY as string;
 const userId = process.env.REACT_APP_USER_ID as string;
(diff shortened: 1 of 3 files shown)


examples/typescript/src/App.tsx
```diff
@@ -9,10 +9,21 @@ import {
...
 
+declare module 'stream-chat' {
+  interface UserEx {
+    customUserField: string;
+  }
+
+  interface ReactionEx {
+    customReactionField: string;
+  }
+}
```
myandrienko, line 25: This is what integrators will have to do, instead of using SCG.

examples/typescript/src/App.tsx
```diff
@@ -21,46 +32,40 @@ const filters: ChannelFilters = { type: 'messaging', members: { $in: [userId] }
...
-  pollOptionType: UR;
-};
-
-const chatClient = StreamChat.getInstance<StreamChatGenerics>(apiKey);
+const App = () => {
+  const chatClient = useCreateChatClient({
+    apiKey,
+    tokenOrProvider: userToken,
+    userData: { id: userId, customUserField: 'field' },
+  });
```
myandrienko, line 40: Note how `userData` supports `customUserField` and doesn't cause type error. That's because `useCreateChatClient` uses `UserResponseEx` type for it.

examples/typescript/src/App.tsx
```diff
@@ -21,46 +32,40 @@ const filters: ChannelFilters = { type: 'messaging', members: { $in: [userId] }
...
 
-if (process.env.REACT_APP_CHAT_SERVER_ENDPOINT) {
-  chatClient.setBaseURL(process.env.REACT_APP_CHAT_SERVER_ENDPOINT);
-}
+  useEffect(() => {
+    (async () => {
+      if (chatClient) {
+        const { reactions } = await chatClient.queryReactions('dummy', {});
+        console.log(reactions[0].customReactionField);
+        console.log(reactions[0].user?.customUserField);
```
myandrienko, line 47: Similarly, `queryReactions` uses `ReactionResponseEx` type, so the response contains both `customReactionField` and `customUserField`.

examples/typescript/src/App.tsx
oliverlaz: This looks very nice! 🚀  Let's plan to fix this for every JS SDK

PR #2406, opened 2024-05-29
fix: render typing indicator outside the VirtualizedMessageList scroll container

### LINKED ISSUE #1941: bug: No default TypingIndicator provided when using VirtualizedMessageList
**Describe the bug**

The `VirtualizedMessageList` does not have a default implementation for the `TypingIndicator`:
https://github.com/GetStream/stream-chat-react/blob/34b8e71771566551180af412f3c13d0740ccc725/src/components/MessageList/VirtualizedMessageList.tsx#L149

The `MessageList` does have a default implementation: https://github.com/GetStream/stream-chat-react/blob/34b8e71771566551180af412f3c13d0740ccc725/src/components/MessageList/MessageList.tsx#L81

**Expected behavior**

I can get the expected behavior by passing the `TypingIndicator` to the `Channel` component
`<Channel TypingIndicator={TypingIndicator} />`.

However, I would've expected this to work out of the box. Are there any specific reasons to not have this as the default? 
I'm also wondering if there are any docs about the differences between `MessageList` and `VirtualizedMessageList` as I've been running into other issues due to slight differences in behaviour between the two.

**Package version**
- stream-chat-react: 10.6.0
- stream-chat-css: 3.9.0
- stream-chat-js: 8.3.0

- OS: Windows
- Browser Chrome
- Version 109

MartinCupela:
### 🎯 Goal

Prevent rendering the TypingIndicator at the bottom of the VML scroll container. This lead to the indicator component not to be visible because the container is not scrolled to the bottom when the indicator is rendered. 

fix #1941 

BREAKING CHANGE: The VirtualizedMessageList does not provide default Footer component
BREAKING CHANGE: The VirtualizedMessageList markup has changed as TypingIndicator is rendered as a child of MessageListMainPanel


oliverlaz (approved):
Please update the PR title (BREAKING CHANGE) so it appears in the changelog

PR #2407, opened 2024-05-29
feat: ThreadList and ThreadProvider (v2)

arnautov-anton:
🚂 https://github.com/GetStream/stream-chat-js/pull/1330

This PR adds components for implementing a thread list view: ThreadList and ChatView. It also adds support for reactive state stores, like the ones that handle the thread list logic in the client.

## useStateStore

Implements a binding from the reactive state store in the client to React components.

## ThreadList

Implements a thread list. The component itself consists mostly of binding to the `ThreadManager` instance in the client.

## Thread

The Thread component was modified to optionally use a reactive `Thread` instance from the `ThreadContext`. The component checks if `ThreadContext` is provided, and if so uses the thread instance from that context. Otherwise, it falls back to the `ChannelContext` as usual.

Currently, we use `ThreadContext` in the thread list view, where it is provided by the `ChatView.ThreadAdapter` component.

Note that even when `ThreadContext` is used, `ChannelContext` also must be provided. Most our components, including the ones used in threads, still expect the channel context to be there.

## ChatView

A set of components implementing a simple in-memory router to switch between channel view and thread list view. Using it is totally optional. Most integrations will probably use their own router instead. Still, nice to have something out-of-the-box.

## WithComponents

A new way to provide component overrides. Previously we allowed passing component overrides as props to `Channel` component, and then in some nested components as well, although not consistently.

`WithComponents` is a new recommended way to provide component overrides, which requires far less prop drilling and nests nicely.


src/components/Channel/Channel.tsx
```diff
@@ -1287,11 +1300,59 @@ const ChannelInner = <
       TriggerProvider: props.TriggerProvider,
       TypingIndicator: props.TypingIndicator,
       UnreadMessagesNotification: props.UnreadMessagesNotification,
-      UnreadMessagesSeparator: props.UnreadMessagesSeparator || UnreadMessagesSeparator,
+      UnreadMessagesSeparator: props.UnreadMessagesSeparator,
       VirtualMessage: props.VirtualMessage,
     }),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [props.reactionOptions],
```
myandrienko, line 1294: I think this was by design, changing overrides "in runtime" is not a good idea.

myandrienko (approved):

PR #2425, opened 2024-06-17
fix: avoid eager channel pagination on channel open

MartinCupela:
### 🎯 Goal

Eager channel pagination lead to problems in Safari. More messages were queried on channel opening which increased the scroll size of the container. Safari does not adjust container scrollTop on prepending new items. The SDK logic however relies on scroll events to maintain the list scrolled to the bottom. As a result the container was not scrolled to the bottom in Safari.

### DIFF
--- a/src/components/InfiniteScrollPaginator/InfiniteScroll.tsx
+++ b/src/components/InfiniteScrollPaginator/InfiniteScroll.tsx
@@ -1,4 +1,4 @@
-import React, { PropsWithChildren, useEffect, useLayoutEffect, useRef } from 'react';
+import React, { PropsWithChildren, useEffect, useRef } from 'react';
 import type { PaginatorProps } from '../../types/types';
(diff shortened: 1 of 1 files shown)


src/components/InfiniteScrollPaginator/InfiniteScroll.tsx
```diff
@@ -93,7 +95,12 @@ export const InfiniteScroll = (props: PropsWithChildren<InfiniteScrollProps>) =>
     }
 
     if (isLoading) return;
-    // FIXME: this triggers loadMore call when a user types messages in thread and the scroll container container expands
+    if (previousOffset.current === offset && previousReverseOffset.current === reverseOffset)
```
MartinCupela, line 98: Prevents duplicate queries when the dimensions do not change.

arnautov-anton (approved):

PR #2432, opened 2024-06-19
feat: load lazily mp3 encoding library for audio recorder

MartinCupela:
### 🎯 Goal

Bundle @breezystack/lamejs separately and thus load it only if audio recording is enabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   "dependencies": {
     "@babel/runtime": "^7.23.6",
     "@braintree/sanitize-url": "^6.0.4",
+    "@breezystack/lamejs": "^1.2.7",
     "@popperjs/core": "^2.11.5",
     "@react-aria/focus": "^3",
     "clsx": "^2.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -99,6 +98,7 @@
     "mml-react": "^0.4.7"
   },
   "peerDependencies": {
+    "@breezystack/lamejs": "^1.2.7",
```
oliverlaz, line 101: I don't think we should make this lib a peer dependency. `await import()` should be enough to instruct the bundler to create a new chunk with it

MartinCupela: I will remove it

oliverlaz: I mean, it should stay as it was - a regular dependency but bundled in a different chunk

MartinCupela: reverted to regular dependency. I have tested it with devDependency + peerDependency scenario and the resulting app bundle included lamejs in a separate file. That is why I thought devDep would be ok.

arnautov-anton (approved):

PR #2434, opened 2024-06-21
fix: adjust ChannelActionContextValue type

arnautov-anton:
### 🎯 Goal

Adjust `ChannelActionContextValue` type to include new `highlightDuration` parameter, pair the types to the functions so that we don't forget in the future (in case we need to adjust).


MartinCupela:
@arnautov-anton this is a good idea 👍 . Can we do the same for `jumpToLatestMessage`?

arnautov-anton:
> @arnautov-anton this is a good idea 👍 . Can we do the same for `jumpToLatestMessage`?

PR #2439, opened 2024-06-27
feat: configure message group size by max time between messages

MartinCupela:
### 🎯 Goal

The goal is the parity with RN Chat SDK.

### 🛠 Implementation details

If the value of `maxTimeBetweenGroupedMessages` is greater then the difference btw the current and next message creation date, the groupd is considered to end with the current message. Effectively, what happens is that the message is applied  a class `str-chat__li--bottom`. The visual result is reflected in `MessageList` (each message has timestamp and author (metadata) shown) but not in `VirtualizedMessageList` which has additional logic that decides whether a group ends at a given message. This should be synced with the `MessageList` but potentially introducing breaking changes to the integrators using `VirtualizedMessageList` :it_is_what_it_is_emoji.

### DIFF
--- a/src/components/MessageList/utils.ts
+++ b/src/components/MessageList/utils.ts
@@ -292,6 +292,7 @@ export const getGroupStyles = <
   previousMessage: StreamMessage<StreamChatGenerics>,
   nextMessage: StreamMessage<StreamChatGenerics>,
   noGroupByUser: boolean,
+  maxTimeBetweenGroupedMessages?: number,
 ): GroupStyle => {
   if (message.customType === CUSTOM_MESSAGE_TYPE.date) return '';
   if (message.customType === CUSTOM_MESSAGE_TYPE.intro) return '';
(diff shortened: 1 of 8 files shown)


oliverlaz (approved):

src/components/MessageList/utils.ts
```diff
@@ -303,24 +304,29 @@ export const getGroupStyles = <
...
-    nextMessage.type === 'error' ||
     nextMessage.deleted_at ||
     (nextMessage.reaction_groups && Object.keys(nextMessage.reaction_groups).length > 0) ||
-    isMessageEdited(message);
+    isMessageEdited(message) ||
+    (maxTimeBetweenGroupedMessages !== undefined &&
+      nextMessage.created_at &&
+      message.created_at &&
+      new Date(nextMessage.created_at).getTime() - new Date(message.created_at).getTime() >
+        maxTimeBetweenGroupedMessages);
```
myandrienko, line 329: Shall we have a similar check above for the `isTopMessage`, but with `message` and `previousMessage`?

MartinCupela: Yes we should so that the next image is associated with top class. Let me check

MartinCupela: Adjusted: https://github.com/GetStream/stream-chat-react/pull/2439/commits/bd824a5223c53c3c7e0f753bea42a45287d9be39

myandrienko (approved):

PR #2447, opened 2024-07-09
fix!: extract MP3 encoder plugin

MartinCupela:
### 🎯 Goal

Make lamejs a peer dependency so that the source code is not bundled within the integrator application source code. This is to prevent issues when integrators do not want to bundle code issued under LGPL license.

### 🛠 Implementation details

Emojis and mp3 encoding function have been moved to separate `plugins` folder. However, the emojis import path stays the same.

A new audio recording configuration parameter has been introduced - `encoder` - to allow integrators to plug in the MP3 or other encoder. 

### 🎨 UI Changes

None

BREAKING CHANGE: MP3 audio encoder has to be explicitly imported and used as a plugin for audio recordings. The default audio recording format is audio/wav.
BREAKING CHANGE: @breezystack/lamejs became a peer dependency and has to be installed by the integrator so that the MP3 audio encoder can work properly.


docusaurus/docs/React/components/message-input-components/audio-recorder.mdx
```diff
@@ -51,6 +51,21 @@ The dialog can be customized by passing own component to `Channel` component con
 <Channel RecordingPermissionDeniedNotification={CustomComponent}>
 ```
 
+## Custom encoding
+
+By default, the recording is encoded into `audio/wav` format. In order to reduce the size and keep the inter-browser format compatibility, you can use an MP3 encoder that is based on [`lamejs` implementation](https://github.com/gideonstele/lamejs). Follow these steps to achieve the MP3 encoding capability.
+
+1. The library `@breezystack/lamejs` has to be installed as this is a peer dependency to `stream-chat-react`.
```
oliverlaz, line 58: I think it would be good if we add an install command here. Easy to copy-paste, instead of having to type the full command manually.

docusaurus/docs/React/release-guides/upgrade-to-v12.mdx
```diff
@@ -4,6 +4,40 @@ title: Upgrade to v12
...
 ---
 
+## Audio recordings transcoding
+
+Until now, the audio recordings were transcoded to `audio/mp3` format for inter-browser compatibility and size reduction. However, as of the v12, the MIME type `audio/wav` will be the default. The mp3 encoder used by the library is licensed under LGPL and integrators will have to opt-in to using the library in their apps.
+
+:::important
+**Action required**<br/>
+
+1. The library `@breezystack/lamejs` has to be installed as this is a peer dependency to `stream-chat-react`.
```
oliverlaz, line 14: Install command here as well.

docusaurus/docs/React/release-guides/upgrade-to-v12.mdx
```diff
@@ -4,6 +4,40 @@ title: Upgrade to v12
 keywords: [migration guide, upgrade, v12, breaking changes]
 ---
 
+## Audio recordings transcoding
+
+Until now, the audio recordings were transcoded to `audio/mp3` format for inter-browser compatibility and size reduction. However, as of the v12, the MIME type `audio/wav` will be the default. The mp3 encoder used by the library is licensed under LGPL and integrators will have to opt-in to using the library in their apps.
```
oliverlaz, line 9: > The mp3 encoder used by the library is licensed under LGPL and integrators will have to opt-in to using the library in their apps.

This sentence creates uncertainty/fear/doubt. I think we should rephrase this to something else, or remove it. IMO, we are compliant with LGPL now. We make mp3 encoding opt-in.

docusaurus/react-docusaurus-dontent-docs.plugin.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
         lastVersion: 'current',
```
oliverlaz, line 6: Until we release v12-final, I think this should point to `11.x.x`

src/plugins/Emojis/EmojiPicker.tsx
```diff
@@ -6,7 +6,7 @@ import Picker from '@emoji-mart/react';
 import type { Options } from '@popperjs/core';
 
 import { useMessageInputContext, useTranslationContext } from '../../context';
-import { EmojiPickerIcon } from '../MessageInput/icons';
+import { EmojiPickerIcon } from './icons';
```
arnautov-anton, line 9: 👍

docusaurus/react-docusaurus-dontent-docs.plugin.js
MartinCupela: But then we will have docs paths with v12 in them. Is that ok?

oliverlaz: yes, this is fine.

MartinCupela: Updated

docusaurus/docs/React/release-guides/upgrade-to-v12.mdx
MartinCupela: Updated

docusaurus/docs/React/components/message-input-components/audio-recorder.mdx
MartinCupela: updated

oliverlaz (approved):

PR #2457, opened 2024-07-25
fix: provide both browser and node cjs bundles

myandrienko:
### 🎯 Goal

Although it rarely makes sense to SSR chat, import `stream-chat-react` on server ideally shouldn't fail.

### 🛠 Implementation details

We now build two CJS bundles: one with node-specific dependencies (esbuild `platform: 'node'`), one with browser-specific dependencies (esbuild `platform: 'browser'`). The node version is exposed as a `"node"` export in package.json, so frameworks like Next.js should prefer it for SSR.


scripts/validate-cjs-node-bundle.cjs
```diff
@@ -0,0 +1,4 @@
+// As the community transitions to ESM, we can easily break our CJS bundle.
+// This smoke test can help to detect this early.
+
+require('../dist/index.node.cjs');
```
myandrienko, line 4: 🍓 The node bundle should be usable without DOM.

arnautov-anton (approved):

PR #2460, opened 2024-08-01
chore(deps): remove isomorphic-ws from dependencies

MartinCupela:
### 🎯 Goal

The dependency is installed as a part of `stream-chat` library already and is not directly used in `stream-chat-react` src.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "fix-webm-duration": "^1.0.5",
     "hast-util-find-and-replace": "^5.0.1",
     "i18next": "^21.6.14",
-    "isomorphic-ws": "^4.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,7 +70,6 @@
     "fix-webm-duration": "^1.0.5",
     "hast-util-find-and-replace": "^5.0.1",
     "i18next": "^21.6.14",
-    "isomorphic-ws": "^4.0.1",
```
oliverlaz, line 73: Why not v12? Also, shouldn't we see an update in the `yarn.lock` file?

MartinCupela: You do not see update in yarn lock, because the dep is installed for stream-chat and that does not change.

PR #2461, opened 2024-08-02
fix: downgrade react-markdown to v8 that supports React version < v18

### LINKED ISSUE #2234: bug: peer dependency error with React 16 / React 17
**Describe the bug**

The peer dependencies have `"react": "^18.0.0 || ^17.0.0 || ^16.8.0"`, but if you install it with react 16.8, there will be peer-dependencies issues because of `react-markdown` `9.0.0` which has `"react": ">=18"` in its peer-dependencies.

It seems to work at runtime for us locally, but was breaking our tests & build

It is ok to drop support for old React versions, but it should be clear in the peer-dependencies so we don't try to update yet

**To Reproduce**

1. Have project with React 16
2. Install latest `stream-chat-react`
3. run install
4. See peer-dependencies issue:

```
└─┬ stream-chat-react 11.2.1
  └─┬ react-markdown 9.0.1
    └── ✕ unmet peer react@>=18: found 16.14.0

```

**Expected behavior**

Peer-dependencies of `stream-chat-react` should reflect real version needs

**Package version**
- stream-chat-react: 11.2.1

- OS: macos
- Browser chrome
- Version latest

### LINKED ISSUE #2455: bug: React versions <18 fail to compile
When installing in a project that using React > 16.8 or React 17, compilation fails when attempting to resolve `react/jsx-runtime`.  The reason for this failure is the pinned version of `react-markdown` requires React > 18.

**To Reproduce**

1. I created [a simple reproduction of the issue here](https://codesandbox.io/p/devbox/nostalgic-brattain-xnmn4n?file=%2Fpackage.json).
2. run `yarn build` to see the failure.

**Expected behavior**

I would expect the JavaScript to compile OR the requirement to be clear that you need to be on React > 18.

**Screenshots**

Screenshot 2024-07-24 at 9 23 02 AM

**Package version**
- stream-chat-react: 11.23.3
- stream-chat-css: n/a
- stream-chat-js: n/a

**Additional context**

I know that it is possible to work around this with configuration of the build process, but this problem ends up bubbling up through all our projects that utilize the library (even indirectly), and it feels janky to have that need for configuration to leak out.

MartinCupela:
### 🎯 Goal

fix #2455
fix #2234

### 🛠 Implementation details

Only react-markdown was downgraded while the libraries used to create remark and rehype plugins have not. Custom rehype or remark plugins using `visit` function from the library `unist-util-visit` will have to check the `index` and `parent` arguments of the `Visitor` function for `null` and `undefined`.

### DIFF
--- a/src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
+++ b/src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
@@ -5,9 +5,7 @@ import type { Break } from 'mdast';
 import type { Nodes } from 'hast-util-find-and-replace/lib';
 
 const visitor: Visitor = (node, index, parent) => {
-  if (index === null || typeof index === 'undefined' || index === 0) return;
-  if (parent === null || typeof parent === 'undefined') return;
-  if (!node.position) return;
+  if (!(index && parent && node.position)) return;
 
   const prevSibling = parent.children.at(index - 1);
   if (!prevSibling?.position) return;
--- a/src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
+++ b/src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
@@ -2,11 +2,11 @@ import { visit, Visitor } from 'unist-util-visit';
 import { u } from 'unist-builder';
 
 import type { Break } from 'mdast';
-import type { Nodes } from 'react-markdown/lib';
+import type { Nodes } from 'hast-util-find-and-replace/lib';
 
 const visitor: Visitor = (node, index, parent) => {
-  if (typeof index === 'undefined' || index === 0) return;
-  if (typeof parent === 'undefined') return;
+  if (index === null || typeof index === 'undefined' || index === 0) return;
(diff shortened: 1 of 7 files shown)


src/components/Message/renderText/remarkPlugins/keepLineBreaksPlugin.ts
```diff
@@ -2,11 +2,11 @@ import { visit, Visitor } from 'unist-util-visit';
...
 
 import type { Break } from 'mdast';
-import type { Nodes } from 'react-markdown/lib';
+import type { Nodes } from 'hast-util-find-and-replace/lib';
 
 const visitor: Visitor = (node, index, parent) => {
-  if (typeof index === 'undefined' || index === 0) return;
-  if (typeof parent === 'undefined') return;
+  if (index === null || typeof index === 'undefined' || index === 0) return;
+  if (parent === null || typeof parent === 'undefined') return;
```
oliverlaz, line 9: Can we simplify this to `if (!index || !parent) return;`?

MartinCupela: Updated :)

PR #2478, opened 2024-08-20
fix: use the client pagination indicators for ChannelStateContext's hasMore and hasMoreNewer flags

MartinCupela:
### 🎯 Goal

Use the pagination flags from the the client, which actually performs the queries and thus share the information btw React components that are siblings (`ChannelList` queries channels and `Channel` will know, what channels have less than the 1st page size message count).

depends on https://github.com/GetStream/stream-chat-js/pull/1332

### DIFF
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -1327,10 +1327,6 @@ const ChannelInner = <
     );
   }
 
-  // @ts-ignore
-  window['channel'] = channel;
-  // @ts-ignore
-  window['client'] = client;
   return (
     <div className={clsx(className, windowsEmojiClass)}>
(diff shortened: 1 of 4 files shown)


src/components/Channel/Channel.tsx
```diff
@@ -1346,6 +1327,10 @@ const ChannelInner = <
     );
   }
 
+  // @ts-ignore
+  window['channel'] = channel;
+  // @ts-ignore
+  window['client'] = client;
```
oliverlaz, line 1333: Let's make sure we don't merge this code :)

oliverlaz (approved):

src/components/Channel/Channel.tsx
MartinCupela: true :)

PR #2483, opened 2024-08-30
fix: address the circular dependencies among TranslationContext and Streami18n

MartinCupela:
### 🎯 Goal

Circular dependencies btw. the translation files caused problems to users who installed the version built with esbuild into an app based on react-scripts and React v17/v16.

### 🛠 Implementation details

As a side effect, this refactor lead to correct execution of extending Dayjs with plugins in Streami18n. Thus the fixed tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   },
   "sideEffects": [
     "*.css",
-    "./src/i18n/Streami18n.ts"
+    "./dist/i18n/Streami18n.js"
   ],
   "keywords": [
     "chat",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -74,7 +74,8 @@
     }
   },
   "sideEffects": [
-    "*.css"
+    "*.css",
+    "./src/i18n/Streami18n.ts"
```
myandrienko, line 78: I think it should be `./dist/i18n/Streami18n.js` :)

MartinCupela: Fixed :)

myandrienko (approved):

oliverlaz (approved):

PR #2489, opened 2024-09-04
feat: add centralized dialog management

### LINKED ISSUE #2022: Feature Request: Unread Count for Threads within Messages
**Motivation**
In our project, we employ the getstream chat where every message can have threads. For our users, it's important to treat each thread as separate messages to accurately track and manage unread messages. Currently, there is no straightforward way to get an unread count of threads in messages. This is a problem because it makes it difficult for users to identify which threads contain unread messages and subsequently navigate to them. This can lead to missed conversations and information, creating frustration among users.

**Proposed solution**
We propose the addition of a feature that would allow us to retrieve an unread count of threads in messages in the getstream chat, similar to how we can get an unread count of messages in a channel. This would ideally be implemented as a function or a field that we could easily access via the API, updating in real time as messages are read or new messages are received in the threads.

**Acceptance Criteria**
The function/feature accurately counts the number of unread threads within messages for any given user and is easily retrievable.

Also, it seems relative with this issue: https://github.com/GetStream/stream-chat-flutter/issues/1337

### LINKED ISSUE #2317: bug: The MessageOptions (triple dot when hovering over a message) requires reverse tabbing for accessibility to select menu items
**Describe the bug**

Accessibility issue on selecting MessageOptions for accessibility keyboard-only users

See screen cap of demo app: 
https://github.com/GetStream/stream-chat-react/assets/24303946/5f005942-0167-469b-bea7-985b82bc4416

We are seeing the same issue in our product that leverages Stream-chat-react

**To Reproduce**
Use demo app provided here: 
https://getstream.io/chat/demos/team/

1. Hover over the MessageOptions (...) when the message is at the top of the page
2. Use keyboard navigation to navigate through the options ---> this works fine
3. Do the same for a message at the bottom of the chat and the options stack upwards
4. This then is difficult for accessibility users to navigate to because it requires navigating backwards for keyboard-only accessibility users

**Expected behavior**
I believe that when MessageOptions opens, the focus should start at the first top item of the list

**Screenshots**

https://github.com/GetStream/stream-chat-react/assets/24303946/5f005942-0167-469b-bea7-985b82bc4416

**Package version**
- stream-chat-react: My company uses "stream-chat-react-native-core@5.20.0", but issue seen on your demo app too
- stream-chat-css:
- stream-chat-js:

- OS: MacOS
- Browser: Seen on Chrome
- Version 122.0.6261.94

MartinCupela:
# Submit a pull request

### 🎯 Goal

Add interface to control opening and closing of dialogs from anywhere in the context.

Closes #2022 
Fixes #2317

Depends on https://github.com/GetStream/stream-chat-css/pull/308

### 🛠 Implementation details

Dialogs are managed within `DialogsManagerContext` provided by `DialogsManagerProvider` and provides `DialogsManager` singleton. This singleton provides an API to retrieve and control individual `Dialog` instances. It also allows to subscribe to its `state` changes.

`Dialog` instances can be accessed by id via `useDialog()` hook. The instance has the following API:

- `open` - enables Dialog.isOpen flag
- `close` - disables Dialog.isOpen flag
- `remove` - removes the Dialog instance from the state
- `toggle` - toggles open/close state

Introduced utility hooks:
`useDialogIsOpen()`  - allows to  observe the open state of a particular `Dialog` instance 
`useDialog` - retrieves a dialog object that exposes API to manage it
`useOpenedDialogCount` - allows to observe changes in open dialog count
 
A component that is to be rendered as a dialog should be rendered as a child of `DialogAnchor` component. This component uses popper to calculate dialog positioning relative to given reference element and renders it via React portal dedicated to the `DialogManager` singleton. The portal destination is rendered as a child of `DialogManagerProvider` and is uniquely identified for the given provider. The portal destination renders the teleported component on an overlay. If the overlay is clicked all the dialogs in context are closed.

Features:
- Introduction of possibility to open / close dialog from anywhere in the context allows multiple elements to control a single dialog.
- Rendering of dialogs on an overlay prevents an issue with keeping a dialog open on mouse leave
- Possibility to use multiple dialog contexts
- Focus within a dialog on open
- Trap focus within the opened dialog and tab in round-robin


src/components/MessageActions/MessageActions.tsx
```diff
@@ -88,16 +81,21 @@ export const MessageActions = <
...
   const isMuted = useCallback(() => isUserMuted(message, mutes), [message, mutes]);
 
-  const hideOptions = useCallback((event: MouseEvent | KeyboardEvent) => {
-    if (event instanceof KeyboardEvent && event.key !== 'Escape') {
-      return;
-    }
-    setActionsBoxOpen(false);
-  }, []);
+  const dialogId = `message-actions--${message.id}`;
+  const dialog = useDialog({ id: dialogId });
```
MartinCupela, line 87: Utility hooks in action

src/components/MessageActions/MessageActions.tsx
```diff
@@ -114,50 +112,46 @@ export const MessageActions = <
...
-  });
-
   if (!messageActions.length && !customMessageActions) return null;
 
   return (
     <MessageActionsWrapper
       customWrapperClass={customWrapperClass}
       inline={inline}
-      setActionsBoxOpen={setActionsBoxOpen}
+      toggleOpen={dialog?.toggleSingle}
```
MartinCupela, line 132: Dialog API in action

src/components/Channel/Channel.tsx
```diff
@@ -1329,20 +1330,22 @@ const ChannelInner = <
...
-                {dragAndDropWindow && (
-                  <DropzoneProvider {...optionalMessageInputProps}>{children}</DropzoneProvider>
-                )}
-                {!dragAndDropWindow && <>{children}</>}
-              </div>
-            </TypingProvider>
-          </ComponentProvider>
-        </ChannelActionProvider>
-      </ChannelStateProvider>
+      <DialogsManagerProvider>
```
MartinCupela, line 1333: Could be probably moved to the level of MessageList?

src/context/DialogsManagerContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+  DialogsManagerProviderContextValue | undefined
+>(undefined);
+
+export const DialogsManagerProvider = ({ children }: PropsWithChildrenOnly) => {
+  const [dialogsManager] = useState<DialogsManager>(() => new DialogsManager());
+
+  return (
+    <DialogsManagerProviderContext.Provider value={{ dialogsManager }}>
+      {children}
+      <DialogPortalDestination />
```
MartinCupela, line 20: Here is the dialog teleported

src/components/Dialog/DialogAnchor.tsx
```diff
@@ -0,0 +1,79 @@
...
+}: DialogAnchorProps) => {
+  const open = useDialogIsOpen(id);
+  const { attributes, popperElementRef, styles } = useDialogAnchor<HTMLDivElement>({
+    open,
+    placement,
+    referenceElement,
+  });
+
+  return (
+    <DialogPortalEntry dialogId={id}>
```
MartinCupela, line 68: Here starts the teleport

src/components/MessageList/MessageList.tsx
```diff
@@ -225,47 +226,49 @@ const MessageListWithContext = <
...
-              {...restInternalInfiniteScrollProps}
-            >
-              <ul className='str-chat__ul' ref={setUlElement}>
-                {elements}
-              </ul>
-              <TypingIndicator threadList={threadList} />
-
-              <div key='bottom' />
-            </InfiniteScroll>
+        <DialogsManagerProvider>
```
MartinCupela, line 229: Applies the overlay only to the contents of the MessageList

src/components/MessageList/VirtualizedMessageList.tsx
```diff
@@ -427,74 +428,76 @@ const VirtualizedMessageListWithContext = <
...
-            overscan={overscan}
-            ref={virtuoso}
-            style={{ overflowX: 'hidden' }}
-            totalCount={processedMessages.length}
-            {...overridingVirtuosoProps}
-            {...(scrollSeekPlaceHolder ? { scrollSeek: scrollSeekPlaceHolder } : {})}
-            {...(defaultItemHeight ? { defaultItemHeight } : {})}
-          />
-        </div>
+        <DialogsManagerProvider id='virtualized-message-list-dialogs-manager'>
```
MartinCupela, line 431: Applies the overlay only to the contents of the VirtualizedMessageList

src/components/MessageList/MessageList.tsx
```diff
@@ -225,47 +226,49 @@ const MessageListWithContext = <
...
-              {...restInternalInfiniteScrollProps}
-            >
-              <ul className='str-chat__ul' ref={setUlElement}>
-                {elements}
-              </ul>
-              <TypingIndicator threadList={threadList} />
-
-              <div key='bottom' />
-            </InfiniteScroll>
+        <DialogsManagerProvider id='message-list-dialogs-manager'>
```
MartinCupela, line 229: Applies the overlay only to the contents of the MessageList

src/components/Dialog/DialogAnchor.tsx
```diff
@@ -0,0 +1,119 @@
...
+      document.removeEventListener('keyup', hideOnEscape);
+    };
+  }, [dialog, open]);
+
+  useEffect(() => {
+    if (!open) {
+      // setting element reference back to null allows to re-run the usePopper component once the component is re-rendered
+      setPopperElement(null);
+    }
+  }, [open, setPopperElement]);
```
myandrienko, line 95: You can do this outside of effect and save on re-renders:

```suggestion
  if (popperElement && !open) {
    setPopperElement(null);
  }
```

I think it also makes sense to move this logic to `useDialogAnchor`, WDYT?

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,178 @@
...
+  openDialogCount = 0;
+  dialogs: Record<DialogId, Dialog> = {};
+  private dialogEventListeners: Record<
+    DialogId,
+    Partial<Record<DialogEvent['type'], DialogEventHandler[]>>
+  > = {};
+  private dialogsManagerEventListeners: Record<
+    DialogsManagerEvent['type'],
+    DialogsManagerEventHandler[]
+  > = { openCountChange: [] };
```
myandrienko, line 43: Just a suggestion: to avoid _yet another event listener_ maybe you can reuse our `StateStore` here? I think this will work nicely with dialog ids as selectors etc.

MartinCupela: Will take a look into that.

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.openDialogCount++;
+    this.dialogsManagerEventListeners.openCountChange.forEach((listener) => listener(this));
+    this.dialogEventListeners[params.id].open?.forEach((listener) => listener(dialog));
+  }
+
+  close(id: DialogId) {
+    const dialog = this.dialogs[id];
+    if (!dialog?.isOpen) return;
+    dialog.isOpen = false;
+    this.openDialogCount--;
```
arnautov-anton, line 152: Once this class is converted to using `StateStore`, this property could be a cached getter deriving the value from the list of open dialogs.

src/components/Dialog/DialogAnchor.tsx
MartinCupela: I have moved this, thank you :)

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+      ...current,
+      dialogs: { ...current.dialogs, [dialog.id]: { ...dialog, isOpen: true } },
+      openDialogCount: ++current.openDialogCount,
+    }));
+  }
+
+  close(id: DialogId) {
+    const dialog = this.state.getLatestValue().dialogs[id];
+    if (!dialog?.isOpen) return;
+    dialog.isOpen = false;
```
arnautov-anton, line 87:
```suggestion
```

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+      dialogs: Object.entries(current.dialogs).reduce<Dialogs>((acc, [dialogId, dialog]) => {
+        if (id !== dialogId) {
+          acc[id] = dialog;
+        }
+        return acc;
+      }, {}),
+      openDialogCount:
+        current.openDialogCount &&
+        (dialog.isOpen ? current.openDialogCount - 1 : current.openDialogCount),
+    }));
```
arnautov-anton, line 131: This reads better and should be a bit faster I believe.
```suggestion
  this.state.next((current) => {
    const newDialogs = { ...current.dialogs };

    delete newDialogs[id];

    return {
      ...current,
      dialogs: newDialogs,
      openDialogCount:
        current.openDialogCount &&
        (dialog.isOpen ? current.openDialogCount - 1 : current.openDialogCount),
    };
  });
```

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+
+type DialogsManagerState = {
+  dialogs: Dialogs;
+  openDialogCount: number;
+};
+
+export class DialogsManager {
+  id: string;
+  state = new StateStore<DialogsManagerState>({
+    dialogs: {},
```
arnautov-anton, line 33:
```suggestion
    dialogsById: {},
```

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+type Dialogs = Record<DialogId, Dialog>;
+
+type DialogsManagerState = {
+  dialogs: Dialogs;
+  openDialogCount: number;
+};
+
+export class DialogsManager {
```
arnautov-anton, line 30:
```suggestion
export class DialogManager {
```

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+  id?: string;
+};
+
+type Dialogs = Record<DialogId, Dialog>;
+
+type DialogsManagerState = {
+  dialogs: Dialogs;
+  openDialogCount: number;
+};
+
```
arnautov-anton, line 29: I feel like instead of keeping (and managing) an actual value of opened dialogs, you can create a reusable selector (even hook utilising this selector) which derives this value from the actual state of the dialogs. This is less error-prone and increases reliability in terms of future updates to the manager itself (less things to care about);

```ts
const openedDialogCountSelector = (nextValue: DialogManagerState) => {
    return Object.values(nextValue.dialogsById).reduce((count, dialog) => {
        if (dialog.isOpen) return count + 1;
        return count;
    }, 0)
}
```
and use like so:

```ts
// manager comes from parameter for example purposes
const useOpenedDialogCount = (manager) => {
    return useStateStore(manager.state, openedDialogCountSelector);
}
```

src/components/Dialog/DialogsManager.ts
```diff
@@ -0,0 +1,133 @@
...
+      };
+      this.state.next((current) => ({
+        ...current,
+        ...{ dialogs: { ...current.dialogs, [id]: dialog } },
+      }));
+    }
+    return dialog;
+  }
+
+  open(params: GetOrCreateParams, single?: boolean) {
```
arnautov-anton, line 71: Nit: Can we merge options so there's only one parameter?

Rename of the actual property, `single` doesn't feel right; `closeRest` or `closeOpened` might better describe what's going to happen if it's set to `true`.
```suggestion
  open(params: GetOrCreateParams, closeOpened?: boolean) {
```

src/components/Dialog/hooks/useDialog.ts
```diff
@@ -0,0 +1,35 @@
...
+    () =>
+      dialogsManager.state.subscribeWithSelector<boolean[]>(
+        ({ dialogs }) => [!!dialogs[id]?.isOpen],
+        ([isOpen]) => {
+          setOpen(isOpen);
+        },
+        // id,
+      ),
+    [dialogsManager, id, source],
+  );
```
arnautov-anton, line 32: Here we can use `useStateStore` instead while reusing your custom selector.

src/components/Dialog/DialogsManager.ts
MartinCupela: I want to avoid array two arrays of open and closed dialogs and iterate each time to find it just to have a getter prop.  Not sure, if I understood what you propose.

src/components/Dialog/DialogsManager.ts
MartinCupela: The `params` arg was intended as a seed of dialog data. The data was gradually reduced to `id`, however I decided to keep it as object in case we would like to add more dialog props in the future. The second param is not related to the dialog being open but is a config param on how to handle opening. That is why I did not want to merge it into `GetOrGreateParams`.

MartinCupela: I will rename `single` to `closeRest`.

src/components/Dialog/DialogsManager.ts
MartinCupela: I will rename all the DialogsX names.

src/components/Dialog/DialogsManager.ts
MartinCupela: will rename

src/components/Dialog/DialogsManager.ts
MartinCupela: I will remove openDialogCount and introduce the subscription selector that counts the number of open dialogs and that way the getter will not be necessary.

arnautov-anton: I think for the amount of active dialogs the subscription with selector should suffice. 👍

MartinCupela: Plays nice with tests

src/components/Dialog/DialogsManager.ts
MartinCupela: done

src/components/Dialog/DialogsManager.ts
MartinCupela: done

src/components/Dialog/DialogsManager.ts
MartinCupela: done

src/components/Dialog/DialogsManager.ts
MartinCupela: Done

src/components/Dialog/DialogsManager.ts
MartinCupela: Done

src/components/Dialog/DialogsManager.ts
MartinCupela: done

src/components/Dialog/hooks/useDialog.ts
MartinCupela: done

src/components/Dialog/hooks/useDialog.ts
```diff
@@ -0,0 +1,42 @@
...
+  }, 0),
+];
+
+export const useOpenedDialogCount = () => {
+  const { dialogManager } = useDialogManager();
+  return useStateStore<DialogManagerState, number[]>(
+    dialogManager.state,
+    openedDialogCountSelector,
+  )[0];
+};
```
arnautov-anton, line 42:
```suggestion
const openedDialogCountSelector = (nextValue: DialogManagerState) => [
  Object.values(nextValue.dialogsById).reduce((count, dialog) => {
    if (dialog.isOpen) return count + 1;
    return count;
  }, 0),
] as const;

export const useOpenedDialogCount = () => {
  const { dialogManager } = useDialogManager();
  return useStateStore(
    dialogManager.state,
    openedDialogCountSelector,
  )[0];
};
```

src/components/Dialog/hooks/useDialog.ts
```diff
@@ -0,0 +1,42 @@
...
+  return dialogManager.getOrCreate({ id });
+};
+
+export const useDialogIsOpen = (id: string) => {
+  const { dialogManager } = useDialogManager();
+  const dialogIsOpenSelector = useCallback(
+    ({ dialogsById }: DialogManagerState) => [!!dialogsById[id]?.isOpen],
+    [id],
+  );
+  return useStateStore<DialogManagerState, boolean[]>(dialogManager.state, dialogIsOpenSelector)[0];
```
arnautov-anton, line 26:
```suggestion
  const dialogIsOpenSelector = useCallback(
    ({ dialogsById }: DialogManagerState) => [!!dialogsById[id]?.isOpen] as const,
    [id],
  );
  return useStateStore(dialogManager.state, dialogIsOpenSelector)[0];
```

arnautov-anton: I wish there was a better way to define tuples in TS but `as const` is as good as it gets. :)

myandrienko (approved):

PR #2491, opened 2024-09-05
docs: fix Threads 2.0 related links

arnautov-anton:
### 🎯 Goal

This PR resolves the issues related to the Threads 2.0 documentation.


docusaurus/docs/React/components/contexts/thread-context.mdx
```diff
@@ -7,9 +7,9 @@ title: ThreadContext
 
 ## ThreadProvider
```
myandrienko, line 8: I don't think we should document `ThreadProvider` - for the same reason we don't document any other providers, they are not user-facing. It's better to add a few more details about `ThreadAdapter` instead.

docusaurus/docs/React/components/core-components/thread-list.mdx
```diff
@@ -3,9 +3,9 @@ id: thread-list
...
+`ThreadList` is a component which renders individual thread instances ([`Thread`](https://github.com/GetStream/stream-chat-js/blob/master/src/thread.ts)) stored within [`ThreadManager`](https://github.com/GetStream/stream-chat-js/blob/master/src/thread_manager.ts). It handles pagination triggers and virtualization through the help of the [Virtuoso](https://virtuoso.dev) virtualized list component. The rest of the business logic lives within ThreadManager and Thread classes. ThreadManager instance gets activated whenever ThreadList renders - activation is necessary as it tells the SDK that user "sees" this list and can update state accordingly whenever appropriate events arrive.
```
myandrienko, line 6: I think it's better to explain `ThreadList` from the user standpoint. In most cases users don't know about ThreadManager and thread instances. So a good introduction would be

> `ThreadList` is a component that displays a list of threads where current user is a participant (user either started a thread or replied in a thread).

The same goes for `Virtuoso` (from the user standpoint, there is no Virtuoso, just `VirtualizedMessageList`), activation (it happens automatically, so why mention it) etc.

docusaurus/docs/React/components/contexts/thread-context.mdx
arnautov-anton: `ThreadAdapter` is a primarily `ChatView` component. It utilises context of the `ChatView.Threads` which provides setter for the `ThreadList` and a "selected thread" value for the `useActiveThread` (handles thread activation based on document visibility and focus states). So for custom purposes such as [this one](https://github.com/GetStream/stream-chat-react/blob/58b22fa9f9175ab5bb165f3f2f77d2497ba45b0d/docusaurus/docs/React/guides/custom-threads-view.mdx#building-custom-threads-view) it becomes unnecessary since the whole "get/set" behavior is being replaced.

arnautov-anton:
Merging as is for now (it mainly fixes broken links and allows documentation build process to finish successfully) and I'll open another one with content fixes.

PR #2500, opened 2024-09-13
fix: update ChannelHeader and ChannelPreview titles and images on channel.updated

MartinCupela:
### 🛠 Implementation details

In case of `ChannelPreview` the effect that set the display image and title was missing dependencies.
In case of `ChannelHeader` the problem with `ChannelStateContext` value memoization was the cause.

### DIFF
--- a/src/components/ChannelHeader/ChannelHeader.tsx
+++ b/src/components/ChannelHeader/ChannelHeader.tsx
@@ -24,7 +24,10 @@ export type ChannelHeaderProps = {
   title?: string;
 };
 
-const UnMemoizedChannelHeader = <
+/**
+ * The ChannelHeader component renders some basic information about a Channel.
+ */
+export const ChannelHeader = <
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
 >(
   props: ChannelHeaderProps,
@@ -86,8 +89,3 @@ const UnMemoizedChannelHeader = <
     </div>
   );
 };
-
-/**
- * The ChannelHeader component renders some basic information about a Channel.
- */
-export const ChannelHeader = React.memo(UnMemoizedChannelHeader) as typeof UnMemoizedChannelHeader;
--- a/src/components/ChannelPreview/hooks/useChannelPreviewInfo.ts
+++ b/src/components/ChannelPreview/hooks/useChannelPreviewInfo.ts
@@ -21,28 +21,29 @@ export const useChannelPreviewInfo = <
 ) => {
   const { channel, overrideImage, overrideTitle } = props;
 
-  const { client } = useChatContext<StreamChatGenerics>('ChannelPreview');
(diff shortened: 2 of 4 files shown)


src/components/ChannelPreview/hooks/useChannelPreviewInfo.ts
```diff
@@ -21,28 +21,29 @@ export const useChannelPreviewInfo = <
...
-  const { client } = useChatContext<StreamChatGenerics>('ChannelPreview');
-  const [displayTitle, setDisplayTitle] = useState(getDisplayTitle(channel, client.user));
-  const [displayImage, setDisplayImage] = useState(getDisplayImage(channel, client.user));
+  const { client } = useChatContext<StreamChatGenerics>('useChannelPreviewInfo');
+  const [displayTitle, setDisplayTitle] = useState(
+    overrideTitle || getDisplayTitle(channel, client.user),
+  );
+  const [displayImage, setDisplayImage] = useState(
+    overrideImage || getDisplayImage(channel, client.user),
+  );
```
oliverlaz, line 30:
```suggestion
  const [displayTitle, setDisplayTitle] = useState(
    () => overrideTitle || getDisplayTitle(channel, client.user),
  );
  const [displayImage, setDisplayImage] = useState(
    () => overrideImage || getDisplayImage(channel, client.user),
  );
```

I would use function useState initializer here to prevent running and immediately discarding the result of `getDisplayImage()`

src/components/ChannelPreview/hooks/useChannelPreviewInfo.ts
```diff
@@ -21,28 +21,29 @@ export const useChannelPreviewInfo = <
...
 
-    client.on('user.updated', handleEvent);
+    updateTitles();
+
+    client.on('user.updated', updateTitles);
     return () => {
-      client.off('user.updated', handleEvent);
+      client.off('user.updated', updateTitles);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
oliverlaz, line 44: nice 🚀

src/components/ChannelHeader/ChannelHeader.tsx
```diff
@@ -86,8 +89,3 @@ const UnMemoizedChannelHeader = <
     </div>
   );
 };
-
-/**
- * The ChannelHeader component renders some basic information about a Channel.
- */
-export const ChannelHeader = React.memo(UnMemoizedChannelHeader) as typeof UnMemoizedChannelHeader;
```
oliverlaz, line 93: 👍

oliverlaz (approved):

src/components/ChannelPreview/hooks/useChannelPreviewInfo.ts
MartinCupela: True

MartinCupela: Done

PR #2543, opened 2024-10-24
feat: new MessageActions component

arnautov-anton:
## The Problem(s)

`MessageActions`/`MessageOptions`/`MessageActionsBox` are a poorly designed set of components which we were trying to patch over the time with things like `CustomMessageActionsList` but with each new patchy addition, the component set got heavier and worse to navigate in when it came to customization.

- `messageActions` prop on `MessageList` and `VirtualizedMessageList` components allow for key/handler customization:
  ```tsx
  { "Custom Action": () => doThing() }
  ```
  Solution like this falls face first when integrators need to access `MessageContext` or use translations for different languages (the key is used as button text).
 - Solution to patch those shortcomings was to introduce `CustomMessageActionsList` which allows to render custom buttons within actions dropdown but... when buttons are rendered conditionally, the "..." is still rendered and upon clicking on it, it opens up an empty actions dropdown. This solution also does not allow to adjust "quick actions" (like reply or react).
 
## Steps Taken

1. pre-define default set of actions with buttons as components with own logic and click handlers, separate into "quick" and "dropdown" types
2. define default filter which takes care of user capabilities and whether the actions are allowed within reply type of a message (and some other stuff based on message type or status)
3. allow integrators to override the default set and to override the default filter function, alow reusing defaults too

```
access action set -> filter action set based on filter function criteria -> separate into quick and dropdown -> render
```

## From Integrator's POV

```tsx
import { Channel, } from 'stream-chat-react';
import {
  MessageActions,
  defaultMessageActionSet,
  DefaultDropdownActionButton,
} from 'stream-chat-react/experimental';

const CustomMessageActions = () => {
  const customFilter = () => {
    /*...*/
  };

  return (
    <MessageActions
      // though not recommended, it's completely possible to disable default filter...
      disableBaseMessageActionSetFilter
      messageActionSet={[
        ...defaultMessageActionSet,
        {
          type: 'myCustomTypeDropdown',
          placement: 'dropdown',
          // we can enforce non-null return type (at least through TS)
          Component: () => <DefaultDropdownActionButton>🚀 Custom</DefaultDropdownActionButton>,
        },
        {
          type: 'myCustomTypeQuick',
          placement: 'quick',
          Component: () => <button>a</button>,
        },
        // ...and apply custom filter here with access to CustomMessageActions scope (contexts + other hooks)
      ].filter(customFilter)}
    />
  );
};

<Channel MessageActions={CustomMessageActions}>...</Channel>;
```

image

## Next Steps

- [x] review and release
- [ ] allow overriding certain components within default `MessageActions_UNSTABLE`
- [ ] tests


src/components/MessageActions/MessageActions.new.tsx
```diff
@@ -0,0 +1,367 @@
...
+    | 'quote'
+    | 'markAsUnread'
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    | (string & {});
+};
+
+const DefaultMessageActions_UNSTABLE = {
+  dropdown: {
+    // eslint-disable-next-line react/display-name
+    Quote: () => {
```
oliverlaz, line 39: `displayName` is quite a useful thing in dev tools. Instead of disabling the eslint rule, why don't we use:
```ts
Quote: function Quote() { ... }
```

src/components/MessageActions/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './MessageActions';
 export * from './MessageActionsBox';
 export * from './CustomMessageActionsList';
+export * from './MessageActions.new';
```
oliverlaz, line 4: I was thinking, what if instad we export a bunch of components/objects with `_UNSTABLE` prefix, we do the opposite?
```ts
// in the sdk
export * as MessageActions_UNSTABLE from './MessageActions.new';

// and then use:
import { MessageActions_UNSTABLE } from 'stream-chat-react';
const { MessageActions,  defaultMessageActionSet } = MessageActions_UNSTABLE;

const MyComponent = () => {
  return <MessageActions />;
}
```

Once this change becomes "stable", we can release it without making too many renames in the SDK, and integrators can fix this by updating the imports only/

arnautov-anton: This is way better, will do. :)

src/unstable/MessageActions/MessageActions.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { useChatContext, useMessageContext, useTranslationContext } from '../../context';
+import { ActionsIcon } from '../../components/Message/icons';
+import { DialogAnchor, useDialog, useDialogIsOpen } from '../../components/Dialog';
+import { MessageActionsWrapper } from '../../components/MessageActions/MessageActions';
+
+import { useBaseMessageActionSetFilter, useSplitMessageActionSet } from './hooks';
+import { defaultMessageActionSet } from './defaults';
+
+export type MessageActionSetItem = {
+  Component: React.ComponentType;
```
arnautov-anton, line 14: Not sure if there's a good way to enforce non-null return type - any suggestions are welcome.

package.json
```diff
@@ -50,6 +50,18 @@
       },
       "default": "./dist/plugins/encoders/mp3.js"
     },
+    "./unstable": {
```
oliverlaz, line 53: I think `experimental` sounds better :)

src/components/Message/MessageOptions.tsx
```diff
@@ -95,13 +95,14 @@ const UnMemoizedMessageOptions = <
...
-        <ReactionSelectorWithButton ReactionIcon={ReactionIcon} theme={theme} />
-      )}
+      {shouldShowReactions && <ReactionSelectorWithButton ReactionIcon={ReactionIcon} />}
     </div>
   );
 };
 
+/**
+ * @deprecated Consider moving to `MessageActions` from `stream-chat-react/unstable` instead.
+ */
```
oliverlaz, line 105: I'm not sure whether we should deprecate these components right now. Perhaps we do this after we gather the initial feedback?

oliverlaz (approved):

PR #2549, opened 2024-10-29
fix: remove rejected channel query promises from channel query lock

MartinCupela:
### 🎯 Goal

Remove stale promise references from the channel query lock. If not removed, even if the whole chat is remounted, the rejected promise from the past leads to display of error in the chat UI. This is because the lock is located in the global scope.

### DIFF
--- a/src/utils/getChannel.ts
+++ b/src/utils/getChannel.ts
@@ -71,10 +71,8 @@ export const getChannel = async <
     try {
       WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid] = theChannel.watch(options);
       await WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
+    } finally {
       delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
-    } catch (e) {
-      delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
-      throw e;
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/getChannel.ts
```diff
@@ -68,9 +68,14 @@ export const getChannel = async <
...
-    await WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
-    delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
+    try {
+      WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid] = theChannel.watch(options);
+      await WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
+      delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
+    } catch (e) {
+      delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
+      throw e;
+    }
```
oliverlaz, line 78:
```suggestion
    try {
      WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid] = theChannel.watch(options);
      await WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
    } finally {
      delete WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL[originalCid];
    }
```

Can be simplified a bit :)

oliverlaz (approved):

src/utils/getChannel.ts
MartinCupela: I wanted to re-throw the error to prevent breaking change

oliverlaz: this re-throws too

MartinCupela: done

PR #2555, opened 2024-11-07
feat: per-user channel pinning/archiving

arnautov-anton:
### 🎯 Goal

Integrate channel pinning and archiving within SDK.

This PR relies on: https://github.com/GetStream/stream-chat-js/pull/1409 and https://github.com/GetStream/stream-chat-css/pull/321


src/components/ChannelList/utils.ts
```diff
@@ -30,3 +30,133 @@ export const moveChannelUp = <
...
+) => {
+  if (!filters) return false;
+
+  return !filters.archived;
+};
+
+export const isChannelPinned = <SCG extends ExtendableGenerics>(channel: Channel<SCG>) => {
+  if (!channel) return false;
+
+  const member = channel.state.membership;
```
michalvankodev, line 151: Hey!

I'm working on implementing pinning channels on our app.
I would just like to inform you that the `channel.state.membership` does not get updated when user pins/unpins a channel when a `member.updated` websocket notifications is received. Those this will not represent the "current" value of the `pinned_at` state.

arnautov-anton: Hey, @michalvankodev, I know - I have another branch with appropriate change ready for a PR, it will be linked to this one and new version of the client will be required before releasing this feature. :)

arnautov-anton: This one: https://github.com/GetStream/stream-chat-js/pull/1409

PR #2570, opened 2024-12-02
feat: add support for ai generated messages

isekovanic:
# Submit a pull request

### 🎯 Goal

This PR implements a PoC of the AI feature being developed across all SDKs.

This includes:

- The implementation of 2 AI related components (the typing indicator and the StreamingMessageView)
- A typewriter animation ingrained within the StreamingMessageView component
- Utility hooks for integrators to build this themselves
- Minor changes across various components to account for having AI in our SDK

The CSS Library PR can be found [here](https://github.com/GetStream/stream-chat-css/pull/320).

~~Note: The `// @ts-ignore` statements are there to account for the non-existing events in the low level client. Whenever this PR gets merged on there, I'll move the state of the AI indicator down to the client and handle the state in there. Also, the names of the events are going to change (probably tomorrow). That's the reason why this isn't in the LLC yet.~~

Unfortunately I ran out of time before being able to move the state to the LLC. For now, it'll stay within the SDK and we can move it down as a next step immediately after this. 

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_


src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export type StopGeneratingButtonProps = {
+  onClick: () => void;
```
MartinCupela, line 4: I would say that it is not necessary to declare onClick apart from other native button propse. I think we can keep here something line

```
export type StopGeneratingButtonProps = React.ComponentProps<'button'>;
```

src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
...
+} & React.ComponentProps<'button'>;
+
+export const StopGeneratingButton = ({ onClick }: StopGeneratingButtonProps) => (
```
MartinCupela, line 7: Not sure if the "Generating" part is necessary in the naming. Looking just at the code, it is not clear what does onClick handler do.

src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
...
+  <button
+    aria-label='Send'
+    className='str-chat__stop-generating-button'
+    data-testid='send-button'
```
MartinCupela, line 11:
```suggestion
    data-testid='stop-button'
```

src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
...
+    onClick={onClick}
+    type='button'
```
MartinCupela, line 13: It is not necessary to add type here.

src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export type StopGeneratingButtonProps = {
+  onClick: () => void;
+} & React.ComponentProps<'button'>;
+
+export const StopGeneratingButton = ({ onClick }: StopGeneratingButtonProps) => (
+  <button
+    aria-label='Send'
```
MartinCupela, line 9:
```suggestion
    aria-label='Stop'
```

src/components/MessageInput/StopGeneratingButton.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export type StopGeneratingButtonProps = {
```
MartinCupela, line 3: Do we actually need a separate file for this button? Could the markup be inlined into where it is used? As there is only one use of it?

src/components/AITypingIndicatorView/AITypingIndicatorView.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useChannelStateContext, useTranslationContext } from '../../context';
+import type { DefaultStreamChatGenerics } from '../../types/types';
+
+export type AITypingIndicatorViewProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = {
+  channel?: Channel<StreamChatGenerics>;
+};
+
+export const AITypingIndicatorView = <
```
MartinCupela, line 16: Could we remove the "View" part from the component. That is more a mobile app thing :)

src/components/MessageInput/StopGeneratingButton.tsx
isekovanic: That's true, it can stop either "Thinking" or "Generating" on its own. The general idea is that we hide this behind a single phrase (i.e integrators don't need to know the difference between the two). What would you propose as something better ? `StopButton` is too generic I think.

examples/vite/src/index.scss
```diff
@@ -30,6 +30,10 @@ body,
     flex-shrink: 0;
     box-shadow: 0 0 8px rgba(0, 0, 0, 0.15);
 
+    .str-chat__channel-preview-messenger {
+      max-width: 1000px;
```
MartinCupela, line 34: What if we added the max-width on the ChanneList root in this file and remove the channel preview CSS?

isekovanic: Sure - will try it out and see

examples/vite/src/App.tsx
```diff
@@ -88,6 +90,7 @@ const App = () => {
             <Window>
               <ChannelHeader Avatar={ChannelAvatar} />
               <MessageList returnAllReadData />
+              <AITypingIndicatorView />
```
MartinCupela, line 93: Wouldn't the indicator cover part of the message being generated in such markdown? Would it be possible that there would be 2 typing indicators at the same time? (one for human and the other for AI)

examples/vite/yarn.lock
```diff
@@ -14,40 +14,6 @@
   resolved "https://registry.yarnpkg.com/@braintree/sanitize-url/-/sanitize-url-6.0.4.tgz#923ca57e173c6b232bbbb07347b1be982f03e783"
   integrity sha512-s3jaWicZd0pkP0jf5ysyHUI/RE7MHos6qlToFcGWXVp+ykHOy77OUMrfbgJ9it2C5bow7OIQwYYaHjk9XlBQ2A==
 
-"@cspotcode/source-map-support@^0.8.0":
```
MartinCupela, line 17: Why some many lines are removed from the lock file?

src/components/AITypingIndicatorView/hooks/useAIState.ts
```diff
@@ -0,0 +1,60 @@
+import { useEffect, useState } from 'react';
+
+import { Channel, Event } from 'stream-chat';
+
+import type { DefaultStreamChatGenerics } from '../../../types/types';
+
+export type AIStateType =
```
MartinCupela, line 7: Could this type be take from the client instead of duplicating it in the component SDK?

src/components/MessageInput/StopGeneratingButton.tsx
isekovanic: Sure thing, makes sense !

src/components/AITypingIndicatorView/hooks/useAIState.ts
isekovanic: Yep, was not available in the client up until this morning

examples/vite/yarn.lock
isekovanic: That's interesting, must've committed something by accident. I'll revert this back to what it was.

src/components/MessageInput/StopGeneratingButton.tsx
MartinCupela: I would inline that default markup and in the component context I would call it `StopAIGenerationButton` or `StopAIOutputButton`.

examples/vite/src/App.tsx
isekovanic: It would technically cover part of it, yes - the general idea is that the integrators can put it wherever they'd like. We don't want it to be part of the SDK itself - perhaps some styling fixes need to be done here ?

src/components/AITypingIndicatorView/AITypingIndicatorView.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+import { Channel } from 'stream-chat';
+
+import { AIStates, useAIState } from './hooks/useAIState';
+
+import { useChannelStateContext, useTranslationContext } from '../../context';
+import type { DefaultStreamChatGenerics } from '../../types/types';
+
+export type AITypingIndicatorViewProps<
```
MartinCupela, line 10: As the indicator does not indicate the typing (typing is indicated by typewriter animation), I would propose name `AiProcessingIndicator` or `AiGenerationIndicator`.

src/components/AITypingIndicatorView/AITypingIndicatorView.tsx
```diff
@@ -0,0 +1,37 @@
...
+  channel: channelFromProps,
+}: AITypingIndicatorViewProps<StreamChatGenerics>) => {
+  const { t } = useTranslationContext();
+  const { channel: channelFromContext } = useChannelStateContext<StreamChatGenerics>(
+    'AITypingIndicatorView',
+  );
+  const channel = channelFromProps || channelFromContext;
+  const { aiState } = useAIState(channel);
+  const allowedStates = {
+    [AIStates.Thinking]: t('Thinking...'),
```
MartinCupela, line 28: Is it possible the string translation would interfere with the aim to match `aiState` to `allowedStates`? I could not find code that would translate `aiState` string.

MartinCupela: I see, you are checking the keys.

src/components/Message/StreamingMessageView.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useStreamingMessage } from './hooks';
+
+export type StreamingMessageViewProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = Pick<MessageTextProps<StreamChatGenerics>, 'message' | 'renderText'> & {
+  letterInterval?: number;
+  renderingLetterCount?: number;
+};
+
+export const StreamingMessageView = <
```
MartinCupela, line 15:
```suggestion
export const StreamedMessageText = <
```

src/components/Message/MessageSimple.tsx
```diff
@@ -211,10 +216,12 @@ const MessageSimpleWithContext = <
                 </span>
               )}
               <MessageTimestamp customClass='str-chat__message-simple-timestamp' />
-              {isEdited && (
+              {isEdited && !message.ai_generated && (
```
MartinCupela, line 219: Why do we need to check for `!message.ai_generated`? Is it possible a message would be edited and AI generated?

src/components/Message/hooks/useStreamingMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type UseStreamingMessageProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = Pick<
+  StreamingMessageViewProps<StreamChatGenerics>,
+  'letterInterval' | 'renderingLetterCount'
+> & { text: string };
+
+const DEFAULT_LETTER_INTERVAL = 30;
+const DEFAULT_RENDERING_LETTER_COUNT = 2;
```
MartinCupela, line 14: Do we need these default variables if they are used only as default param values?

src/components/Message/hooks/useStreamingMessage.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useRef, useState } from 'react';
+
+import type { DefaultStreamChatGenerics } from '../../../types/types';
+import { StreamingMessageViewProps } from '../StreamingMessageView';
```
MartinCupela, line 4:
```suggestion
import type { StreamingMessageViewProps } from '../StreamingMessageView';
```

src/components/Message/StreamingMessageView.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { MessageText, MessageTextProps } from './MessageText';
+import type { DefaultStreamChatGenerics } from '../../types/types';
+import { useMessageContext } from '../../context';
+import { useStreamingMessage } from './hooks';
+
+export type StreamingMessageViewProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = Pick<MessageTextProps<StreamChatGenerics>, 'message' | 'renderText'> & {
+  letterInterval?: number;
```
MartinCupela, line 11:
```suggestion
  streamingLetterIntervalMs?: number;
```

src/components/AITypingIndicatorView/AITypingIndicatorView.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const AITypingIndicatorView = <
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>({
+  channel: channelFromProps,
+}: AITypingIndicatorViewProps<StreamChatGenerics>) => {
+  const { t } = useTranslationContext();
+  const { channel: channelFromContext } = useChannelStateContext<StreamChatGenerics>(
+    'AITypingIndicatorView',
+  );
+  const channel = channelFromProps || channelFromContext;
```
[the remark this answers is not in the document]
isekovanic: The typing indicator is generally meant to be pluggable anywhere (for example, we might have a single channel that contains AI and we want an indicator whenever that specific channel has something going on in it). 

It does not theoretically need to be within `Channel`, hence the prop so that we can pass what we need in that event.

src/components/Message/hooks/useStreamingMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type UseStreamingMessageProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = Pick<
+  StreamingMessageViewProps<StreamChatGenerics>,
+  'letterInterval' | 'renderingLetterCount'
+> & { text: string };
+
+const DEFAULT_LETTER_INTERVAL = 30;
```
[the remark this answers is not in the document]
isekovanic: Yep

src/components/Message/hooks/useStreamingMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    const textLength = text.length;
+    const interval = setInterval(() => {
+      if (!text || textCursor.current >= textLength) {
+        clearInterval(interval);
+      }
+      const newCursorValue = textCursor.current + renderingLetterCount;
+      const newText = text.substring(0, newCursorValue);
+      textCursor.current += newText.length - textCursor.current;
```
MartinCupela, line 34: This is hard to read. Do I understand it well, that this line can be translated to


```suggestion
      textCursor.current = newCursorValue;
```

src/components/Message/hooks/useStreamingMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+>({
+  letterInterval = DEFAULT_LETTER_INTERVAL,
+  renderingLetterCount = DEFAULT_RENDERING_LETTER_COUNT,
+  text,
+}: UseStreamingMessageProps<StreamChatGenerics>) => {
+  const [streamedMessageText, setStreamedMessageText] = useState<string>(text);
+  const textCursor = useRef<number>(text.length);
+
+  useEffect(() => {
+    const textLength = text.length;
```
MartinCupela, line 27: could we clear interval at the beginning of the effect function instead of cleanup and inside the setInterval?

src/components/Message/hooks/useStreamingMessage.ts
isekovanic: Well, no - that's technically incorrect as if you overshoot (for example if you do 5 characters at a time), `newCursorValue` might be above the message length. This makes sure it's at most at the last character.

src/components/Message/hooks/useStreamingMessage.ts
isekovanic: The one inside would still be needed in the case of an early break I believe. For the first one, sure - but why ?

src/components/MessageInput/StopGeneratingButton.tsx
isekovanic: Since this is now overridable it made sense to keep it separate, as discussed

src/components/Message/MessageSimple.tsx
isekovanic: As discussed, will move this to the `isEdited` check function.

src/components/Message/hooks/useMessageTextStreaming.ts
```diff
@@ -0,0 +1,46 @@
...
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = Pick<
+  StreamedMessageTextProps<StreamChatGenerics>,
+  'streamingLetterIntervalMs' | 'renderingLetterCount'
+> & { text: string };
+
+const DEFAULT_LETTER_INTERVAL = 30;
+const DEFAULT_RENDERING_LETTER_COUNT = 2;
+
+export const useMessageTextStreaming = <
```
isekovanic, line 16: JSDocs

src/components/MessageInput/StopAIGenerationButton.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export type StopAIGenerationButtonProps = React.ComponentProps<'button'>;
+
+export const StopAIGenerationButton = ({ onClick, ...restProps }: StopAIGenerationButtonProps) => (
+  <button
+    aria-label='aria/Stop'
```
isekovanic, line 7:
```suggestion
    aria-label='aria/Stop AI Generation'
```

isekovanic: Also translations

src/components/AITypingIndicatorView/AIGenerationIndicator.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useChannelStateContext, useTranslationContext } from '../../context';
+import type { DefaultStreamChatGenerics } from '../../types/types';
+
+export type AIGenerationIndicatorProps<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+> = {
+  channel?: Channel<StreamChatGenerics>;
+};
+
+export const AIGenerationIndicator = <
```
MartinCupela, line 16: Could this component be called `AiStateIndicator`?

src/components/Message/renderText/renderText.tsx
```diff
@@ -35,6 +35,13 @@ export const defaultAllowedTagNames: Array<keyof JSX.IntrinsicElements | 'emoji'
   // custom types (tagNames)
   'emoji',
   'mention',
+  'table',
+  'thead',
+  'tbody',
+  'th',
+  'tr',
+  'td',
+  'tfoot',
```
arnautov-anton, line 44: Can we move these above the "// custom types" comment as these are valid HTML elements?

isekovanic: Oh, sure - missed that

PR #2572, opened 2024-12-05
fix: add support for custom resolution of whether a message is ai generated

isekovanic:
# Submit a pull request

### 🎯 Goal

With the [introduction of the AI feature](https://github.com/GetStream/stream-chat-react/pull/2570) in the SDK, we introduced a behaviour that checks for `message.ai_generated` and manages the UI based on that. This is very wrong, as this field (potentially) comes as a custom property and its place is not in the SDK. Instead, we introduce a new property called `isMessageAIGenerated`, which is a factory function that controls the custom behaviour of figuring out whether a message is AI generated or not (which would vary from integration to integration). It also makes sure this isn't a breaking change for the SDK any longer.

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_


src/components/ChannelList/ChannelList.tsx
```diff
@@ -75,6 +75,7 @@ export type ChannelListProps<
     channel: Channel<StreamChatGenerics>,
     t: TranslationContextValue['t'],
     userLanguage: TranslationContextValue['userLanguage'],
+    isMessageAIGenerated?: ChatContextValue['isMessageAIGenerated'],
```
isekovanic, line 78: Had to add this as an optional argument to make sure the change isn't breaking, as `getLatestMessagePreview` is overridable in the SDK.

src/components/Chat/hooks/useCreateChatContext.ts
```diff
@@ -41,6 +42,7 @@ export const useCreateChatContext = <
       closeMobileNav,
       customClasses,
       getAppSettings,
+      isMessageAIGenerated,
```
MartinCupela, line 45: It is missing in the memo deps

src/components/Message/MessageSimple.tsx
```diff
@@ -187,7 +188,7 @@ const MessageSimpleWithContext = <
               {message.attachments?.length && !message.quoted_message ? (
                 <Attachment actionHandler={handleAction} attachments={message.attachments} />
               ) : null}
-              {message.ai_generated ? (
+              {isMessageAIGenerated(message) ? (
```
MartinCupela, line 191: Could be probably run in a single place as we do not know, what the integrators will put there.

isekovanic: You mean something like calculating `isAIGenerated` for example before passing it to the `MessageContext` ?

PR #2573, opened 2024-12-06
fix: change useStateStore to use useSyncExternalStore

arnautov-anton:
### 🎯 Goal

Changing stores on the fly would keep previously calculated state for a bit before the effect would run to recalculate it - using `useSyncExternalStore` (thank you, @myandrienko) should alleviate this issue. Both `subscribe` and `getSnapshot` functions required by the React hook are wrapped to allow for selector functionality, [`geSnapshot` requires the output to be cached](https://react.dev/reference/react/useSyncExternalStore#parameters) so the wrapper reuses similar cache check mechanism as `subscribeWithSelector` does internally.

### DIFF
--- a/src/store/hooks/useStateStore.ts
+++ b/src/store/hooks/useStateStore.ts
@@ -1,6 +1,9 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useMemo, useSyncExternalStore } from 'react';
 
 import type { StateStore } from 'stream-chat';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
 
 export function useStateStore<
   T extends Record<string, unknown>,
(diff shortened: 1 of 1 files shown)


src/store/hooks/useStateStore.ts
```diff
@@ -1,7 +1,10 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useMemo, useSyncExternalStore } from 'react';
```
myandrienko, line 1: To support older React versions, it's better to use the `use-sync-external-store/shim` package. It also has a `useSyncExternalStoreWithSelector` helper hook which will do selector and selection memoization for you.

arnautov-anton: Thank you, I'll use the shim but the one with selector still requires you to provide a comparator for the selected values so I'll keep it the way it is as it seems that re-writing it to fit the other API would yield similar results - or as would James Franco say:

MartinCupela: Do I understand it well, that not using `use-sync-external-store/shim` could lead to new GH issues of those, who are using React 16 @myandrienko ?

arnautov-anton: For the ones who have React version smaller than React 18.

myandrienko: That's correct, you should use `use-sync-external-store/shim` if you want to support `react@<18`. It uses the "native" implementation on `react@>=18`, so there's no real downside.

MartinCupela (approved):

PR #2574, opened 2024-12-07
chore: update ESLint, migrate configuration file

arnautov-anton:
### 🎯 Goal

Time to UPGRADE and migrate. :)

PR is divided into multiple commits, ESLint changes, Prettier changes and manual changes - we can drop Prettier changes for `master` merges (so that I don't have to deal with merge conflicts) and merge this PR once our PR backlog is somewhat empty.

#### Notable Changes

- changed CI to run on Node@lts/* instead of v16 and v18
- adjusted `bundle-cjs.mjs` and `getPackageVersion.mjs` to use import for `package.json` instead (`with` assertion)
- reduced Prettier print-width to 90 (from 100, default is 80) as some new formattings are just weird
- omitted `eslint-plugin-prettier` as Prettier format step is included in CI anyway
- explicitly installed `@testing-library/dom` as it's required by our tests and is no longer installed with `eslint-plugin-jest-dom`
- `noUnusedParameters` and `noUnusedLocals` (`tsconfig.json`) are now set to `false` to not fail build steps, developers are still warned through ESLint
- omitted certain other plugins since those weren't used apparently, compare against these old configs - [generated against Attachment.test.js](https://gist.github.com/arnautov-anton/9bdce867ed1ddf2b82a6f3f7a5bd9e04) and [against Attachment.tsx](https://gist.github.com/arnautov-anton/ee8dbdb8fda4906a0c436c21686e2105)


.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [16, 18]
+        node: [18, 20]
```
oliverlaz, line 30: Do we have a good reason to run tests with two different node versions? If not, I think we can directly go to v22 (latest LTS).

package.json
```diff
@@ -223,27 +222,12 @@
...
-    "eslint-plugin-jsx-a11y": "^6.4.1",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-prettier": "^3.1.4",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-react": "^7.21.5",
-    "eslint-plugin-react-hooks": "^4.2.0",
-    "eslint-plugin-sort-destructure-keys": "1.3.5",
-    "eslint-plugin-sort-keys-fix": "^1.1.2",
-    "eslint-plugin-testing-library": "^6.2.0",
-    "eslint-plugin-typescript-sort-keys": "^2.1.0",
```
oliverlaz, line 246: 🤯 🥳

oliverlaz (approved):
Awesome stuff! Thanks for taking care of this long overdue upgrade!

package.json
isekovanic: lolz

MartinCupela (approved):

PR #2588, opened 2024-12-20
feat: add general purpose search components

### LINKED ISSUE #1412: MessageSearchList component
As a developer using the React SDK, I would like to be able to quickly implement a user interface that displays the results from searching the messages in a certain channel. 

**Acceptance Criteria**
* A list UI component (`MessageSearchList`) that abstracts the [underlying message search API](https://getstream.io/chat/docs/javascript/search/?language=javascript).
* Automatic pagination [through endless scrolling](https://virtuoso.dev/endless-scrolling/). Use the [`prev/next` pagination mode](https://getstream.io/chat/docs/javascript/search/?language=javascript#pagination).
* component properties that allow the user to specify the query and the search options

**Out of scope**
* an input component that enables the end-user to type their query. This is trivial to integrate from the outside, but hard to customize if built-in. 

**Definition of Done**
* Publicly accessible React Component
* Doc comments
* Docs: Integration Guide 
* Unit tests

MartinCupela:
### 🎯 Goal

Provide general purpose search components and controller. Allow for absolute search customization.

Fixes React-24

Closes #1412

### Depends on

- https://github.com/GetStream/stream-chat-js/pull/1434
- https://github.com/GetStream/stream-chat-css/pull/323
- https://github.com/GetStream/docs-content/pull/137

#### Search controller

Coordinates search execution among individual search sources. Allows to combine unknown search sources defined by the integrator. Implements the search state machine.

#### Base search source

Takes care of common search source task like:
- debounced execution of the query logic defined by every search source
- keep the activity, pagination and loading state.

#### Search source

Class that takes advantage of inheriting from BaseSearchSource. Each search source should defined integrators or the default one is responsible for:
- custom search query composition and execution,
- additional query results filtering
- search filters / sort / options override

#### React components

The new components live in `experimental/Search` folder. This allows for component name duplication compared to `ChannelSearch` components.

An important change is the ability to jump from a message search result to that message in the parent channel. Currently the information about the jumped-to-message is stored in the SearchController state, but this should be moved to a future message list controller state.

The searchController instance has been added to the `ChatContext`.

`ChannelList` allows to override the `ChannelSearch` component with the experimental `Search` component as 

```jsx
<WithComponents overrides={{ Search }}>
 <ChannelList/>
</WithComponents>
```

Also `ChannelAvatar` has been added to the component context value as a separate key and is prioritized in `ChannelHeader` and `ChannelPreview`,

### Out of scope

- search for replies - we cannot jump to a message in a thread message list and implementing this would extend the scope of the PR
- showing search querying errors - we do not have a general purpose notification service implementation of which would extend the scope
- keeping the search string history in SearchController - missing clear specs / exploration on how this should work
- ability to add custom search source state properties
- removing messages that were truncated - this should be handled by the back-end, but is not at the moment and so the results will contain messages to which it will not be possible jump

### 🎨 UI Changes

1. Search inactive

image

2. Search active (all sources active)

image

3. Search active (single source active)

image

4. Quick overview of state changes

[VO0L.webm](https://github.com/user-attachments/assets/aa12d216-4e91-4c43-ad3c-07a628f27e83)


examples/vite/src/App.tsx
```diff
@@ -70,10 +72,15 @@ const App = () => {
     userData: { id: userId },
   });
 
+  const isMessageAIGenerated = useCallback(
+    (message: StreamMessage<StreamChatGenerics>) => !!message?.ai_generated,
+    [],
+  );
```
arnautov-anton, line 78: I think we can move this out of the `App` component completely.

MartinCupela: Moved out: https://github.com/GetStream/stream-chat-react/pull/2588/commits/fc2a860ded11417fd6b4f4017a76191a69223f5f

src/components/ChannelList/ChannelList.tsx
```diff
@@ -380,18 +398,29 @@ const UnMemoizedChannelList = <
     },
   );
 
-  const showChannelList = !searchActive || additionalChannelSearchProps?.popupResults;
+  const showChannelList =
+    (!searchActive && !searchIsActive) || additionalChannelSearchProps?.popupResults;
```
arnautov-anton, line 402: Not sure what the difference between `searchActive` and `searchIsActive` is at first glance.

MartinCupela: One is a label to display the old ChannelSearch and the other is derived from the state of searchController to display the new Search component.

oliverlaz (approved):

PR #2615, opened 2025-01-22
chore: upgrade to React v19

MartinCupela:
Tested locally with app running on 
- react@16.14.0 and installing stream-chat-react from Verdaccio.
- react@19

Fixes React-253


src/components/Attachment/__tests__/File.test.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
+import { render } from '@testing-library/react';
 import '@testing-library/jest-dom';
-import renderer from 'react-test-renderer';
```
MartinCupela, line 3: `react-test-renderer` has been deprecated. Therefore many snapshots had to be changed.

src/components/Channel/__tests__/useIsMounted.test.js
```diff
@@ -1,4 +1,4 @@
-import { act, renderHook } from '@testing-library/react-hooks';
+import { act, renderHook } from '@testing-library/react';
```
MartinCupela, line 1: @testing-library/react-hooks does not seem to be continued - it has not been updated to be compatible with React v19.

package.json
```diff
@@ -202,10 +202,10 @@
     "@types/lodash.throttle": "^4.1.7",
     "@types/lodash.uniqby": "^4.7.7",
     "@types/moment": "^2.13.0",
-    "@types/react": "^18.2.55",
-    "@types/react-dom": "^18.2.19",
+    "@types/react": "^19.0.7",
+    "@types/react-dom": "^19.0.3",
     "@types/react-image-gallery": "^1.2.4",
-    "@types/react-is": "^18.2.4",
+    "@types/react-is": "^19.0.0",
```
oliverlaz, line 208: Can we drop depending on `react-is`? I see it is used only in two places. I'll be happy to have one dependency less

MartinCupela:
In order to be able to run stream-chat-react with React v19 you need to allow for our dependency react-markdown to install its own version of React types. Add the following section to your package.json:

```JSON
"resolutions": {
    "react-markdown/@types/react": "^18.0.0"
  }
```

After that do not forget to remove and re-install `node_modules`.

davidchalifoux:
> In order to be able to run stream-chat-react with React v19 you need to allow for our dependency react-markdown to install its own version of React types. Add the following section to your package.json:
> 
> ```json
> "resolutions": {
>     "react-markdown/@types/react": "^18.0.0"
>   }
> ```
> 
> After that do not forget to remove and re-install `node_modules`.

I'm trying this fix with pnpm with no luck, has anyone else had success?

```json
	"pnpm": {
		"overrides": {
			"react-markdown>@types/react": "^18.0.0"
		}
	},
```

JasonAtallah:
@davidchalifoux I'm running into the same issue. Were you able to find a fix?

davidchalifoux:
> @davidchalifoux I'm running into the same issue. Were you able to find a fix?

Yes I ended up manually patching version 8.0.7 of `react-markdown` using pnpm's [patch](https://pnpm.io/cli/patch) feature and setting the `complex-types.ts` file to use the correct types for React 19. Since it's only changing the types, this should not break any functionality.

I can't give you exact instructions, but the diff looks like this (specifically lines 24-28):
Screenshot 2025-02-11 at 3 17 18 PM

JasonAtallah:
That seemed to work for me. Thanks @davidchalifoux!

PR #2622, opened 2025-01-27
fix: mark channel read for all incoming unread messages

MartinCupela:
### 🎯 Goal

Prevent marking channel based on unread count. Marking channel read based on unread count lead to a paradoxical situation when user that sent the last message in a channel was seen by others as not having read a channel for some time despite the authorship of the last message in the channel. In other words, the last read data for a given user were not updated unless the user had unread count > 0 at some point.

Fixes REACT-252

### 🛠 Implementation details

Remove unread count from the conditions to mark a channel read and instead base the decision on the last message in the channel compared to the last read message of a user.


src/components/MessageList/hooks/useMarkRead.ts
```diff
@@ -1,17 +1,27 @@
...
+import type { Channel, Event, MessageResponse } from 'stream-chat';
+import type { DefaultStreamChatGenerics } from '../../../types';
+
+const hasReadLastMessage = <
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics
+>(
+  channel: Channel<StreamChatGenerics>,
+  userId: string,
+) => {
+  const lastMessageIdInCurrentMessageSet = channel.state.messages.slice(-1)[0]?.id;
```
szuperaz, line 17: I think this should be `channel.state.latestMessages` because in case of a message jump `channel.state.messages` can point to an older message set

MartinCupela: good point 👍

szuperaz (approved):

oliverlaz (approved):

PR #2633, opened 2025-02-05
fix: make `channel.visible` respect archived and pinned channels

arnautov-anton:
### 🎯 Goal

References: https://github.com/GetStream/stream-chat-react-native/pull/2925

### DIFF
--- a/src/components/ChannelList/hooks/useChannelListShape.ts
+++ b/src/components/ChannelList/hooks/useChannelListShape.ts
@@ -2,7 +2,6 @@
 
 import { Dispatch, SetStateAction, useCallback, useEffect, useMemo, useRef } from 'react';
 import { Channel, Event, ExtendableGenerics } from 'stream-chat';
-import uniqBy from 'lodash.uniqby';
 
 import {
   extractSortValue,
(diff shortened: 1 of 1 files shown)


src/components/ChannelList/hooks/useChannelListShape.ts
```diff
@@ -353,20 +353,36 @@ export const useChannelListShapeDefaults = <SCG extends ExtendableGenerics>() =>
...
+        return;
       }
+
+      const channel = await getChannel({
+        client,
+        id: event.channel.id,
+        type: event.channel.type,
+      });
+
+      const considerArchivedChannels = shouldConsiderArchivedChannels(filters);
```
isekovanic, line 374: Ah, I think I missed this bit - I could it could happen that a `channel` is archived and hidden so making it visible should respect the archive first, right ?

isekovanic (approved):
Looks good ! Don't forget to remove [this](https://github.com/GetStream/stream-chat-react/blob/1c7639a917fd858f76be100262547bfbe9e90d0f/src/components/ChannelList/hooks/useChannelListShape.ts#L242) if you intended to do so (or this can happen once we merge the LLC PR too)

src/components/ChannelList/hooks/useChannelListShape.ts
arnautov-anton: I mean - realistically this should never happen but it's there to respect the filter value in case it ever happens (whatever the actual integration looks like). I just copied and slightly adjusted the `notification.message_new` handler.

isekovanic: No no, it's totally on point - I just missed it when doing the fixes 👍

PR #2471, opened 2024-08-15
fix(deps): drop remark-gfm version

arnautov-anton:
# Submit a pull request

### 🎯 Goal

When user posts a specific message, the whole application crashes due to [this issue](https://github.com/remarkjs/remark-gfm/issues/57).

This PR lowers `remark-gfm` to a version that works well with the current version of the `react-markdown` used within the SDK.

I won't be posting the breaking message in the PR for the obvious reasons, message me if interested.

PR #2643, opened 2025-02-12
chore(deps): upgrade react-markdown to v9

### LINKED ISSUE #2641: Support React 19 (`react-markdown@9.x`)
**Motivation**
The current version of stream-react-chat is incompatible with React 19 due to its dependency on react-markdown 8.x. This older version of react-markdown causes TypeScript errors and compatibility issues when used with React 19, specifically around JSX namespace declarations. This prevents developers from upgrading their applications to the latest version of React while using stream-react-chat.

**Proposed solution**
Upgrade the react-markdown dependency in stream-react-chat from version 8.x to version 9.x. This newer version includes proper TypeScript support for React 19 and resolves the JSX namespace declaration issues.

Changes required:
- Update package.json to use react-markdown ^9.0.0
- Update any breaking changes in the implementation that might be affected by the react-markdown upgrade
- Update peer dependencies to include React 19 support
- Update TypeScript types to accommodate the new react-markdown version

**Acceptance Criteria**
[ ] react-markdown is upgraded to version 9.x in package.json
[ ] All existing markdown rendering functionality continues to work as expected
[ ] TypeScript compilation succeeds without JSX namespace errors
[ ] Package works correctly with React 19
[ ] All existing tests pass with the new implementation
[ ] Peer dependency warnings are resolved when used with React 19

Additional context:
This upgrade is critical for developers who want to use stream-react-chat in modern React applications while maintaining type safety and avoiding deprecation warnings.

MartinCupela:
### 🎯 Goal

Tested for: https://github.com/GetStream/stream-chat-react/pull/2471

Fixes #2641
Fixes REACT-284


package.json
```diff
@@ -122,12 +122,12 @@
     "react-dropzone": "^14.2.3",
     "react-fast-compare": "^3.2.2",
     "react-image-gallery": "1.2.12",
-    "react-markdown": "^8.0.7",
+    "react-markdown": "^9.0.3",
```
oliverlaz, line 125: Shouldn't we also update:
- `unist-builder`
- `unist-util-visit`
- `hast-util-find-and-replace`

MartinCupela: I have not downgraded them previously with react-markdown downgrade. But I see that `unist-builder` has a new version so I have installed it here: https://github.com/GetStream/stream-chat-react/pull/2643/commits/611555fc89081a92f29f6ffd7fc40a813449ae32

PR #2645, opened 2025-02-12
chore: replace JSX.Element with ReactNode type

MartinCupela:
### 🎯 Goal

Prevent @types/react@19 complaints about `string | JSX.Element` not being assignable to `ReactNode`.


arnautov-anton (approved):

arnautov-anton:
Can we also do `"jsx": "react-jsx"` in `tsconfig.json` now? I think it's time. 🤔

oliverlaz:
> Can we also do `"jsx": "react-jsx"` in `tsconfig.json` now? I think it's time. 🤔

Maybe do this with the upcoming v13? We need to support `react-jsx` and `react-jsx-dev`.

oliverlaz (approved):

PR #2698, opened 2025-05-07
build: restore version replacement

arnautov-anton:
### 🎯 Goal

Partially restore changes from #2663, return version replacement functionality.


MartinCupela:
Why is this better than adjusting vite.config.js?

arnautov-anton:
> Why is this better than adjusting vite.config.js?

Closing, decided to keep @MartinCupela's solution instead.

PR #2709, opened 2025-05-12
fix: reflect maxRows in TextareaComposer and allow to add lines

### LINKED ISSUE #2705: bug: deprecated grow in message input in v13
**Describe the bug**

In version 13, the `grow` property was removed from `MessageInput`. Is there any alternative to enable multiple lines without requiring major customization? 👀 

I tested the `maxRows` property suggested in the [upgrade documentation](https://getstream.io/chat/docs/sdk/react/release-guides/upgrade-to-v13/), but it did not enable line breaks with `Shift + Enter` as it did before.

**Screenshots**

Image

Image

**To Reproduce**

1. Upgrade `stream-chat-react` to `^13.0.0`
2. Use `MessageInput` component with `grow` property

**Expected behavior**

It was expected that the user would be able to type multiple lines using `Shift + Enter` as was done in the previous version.

**Package version**

- stream-chat-react: 13.0.0
- stream-chat-js: 9.0.1

- OS: macOS Sequoia (15.4.1)
- Browser: chrome
- Version: 135.0.7049.117 (arm64)

MartinCupela:
### 🎯 Goal

Fixes #2705
Closes REACT-381


src/components/TextareaComposer/TextareaComposer.tsx
```diff
@@ -143,8 +143,8 @@ export const TextareaComposer = ({
         return;
       }
 
-      if (event.key === 'Enter') {
-        // allow next line only on Shift + Enter. Enter is reserved for submission.
+      if (event.key === 'Enter' && !event.shiftKey) {
+        // Allow next line only on Shift + Enter. Enter is reserved for submission.
         event.preventDefault();
       }
```
arnautov-anton, line 149: This is what `shouldSubmit` is for - this way we hard-code the behavior and if users decide to have shift+enter as an actual submit, they cannot really change it.

MartinCupela: Ok, I will move should submit higher and if the combination is not specified as shift + Enter then this can be executed

MartinCupela: Update: I had to just move the preventDefault call inside the condition that checks for submissions being open. When a submission is selected we want to prevent the default behavior of the textarea to add a new line.

oliverlaz (approved):

PR #2733, opened 2025-06-12
feat: allow to send thread reply to channel

MartinCupela:
### 🎯 Goal

Closes REACT-413
Depends on:

- https://github.com/GetStream/stream-chat-js/pull/1561
- https://github.com/GetStream/stream-chat-css/pull/336

### 🎨 UI Changes

[show-reply-in-channel.webm](https://github.com/user-attachments/assets/feba5b44-7542-4603-add0-e72c2047c342)


src/components/Message/MessageIsThreadReplyInChannelButtonIndicator.tsx
```diff
@@ -0,0 +1,68 @@
...
+              // prevent further search queries if the message is not found in the DB
+              parentMessageRef.current = null;
+            }
+            return;
+          }
+          openThread(parentMessageRef.current);
+        }}
+      >
+        {t<string>('Thread reply')}
+      </button>
```
oliverlaz: To prevent surprises with forms, can you please add a `type="button` prop to the `<button />` element?

Also, please add error handling around `await querySearchParent()`, as in case it fails, the failure will be reported as "Unhandled Promise Rejection" in customer apps, and there is nothing they can do about it.

oliverlaz: Also, didn't we get rid of the generics of `t<string>()`?

oliverlaz (approved):

src/components/Message/MessageIsThreadReplyInChannelButtonIndicator.tsx
MartinCupela: - generics - yes, in another PR that is still to be merged - was waiting for approval of client changes.
- `type='button'` added
- added error handling to thread search request

PR #2740, opened 2025-06-19
fix: InfiniteScroll - replace refs with state-based deps

arnautov-anton:
# Submit a pull request

### 🎯 Goal

Fixes an issue where `MessageList` is loaded scrolled up to the seemingly random location and not at the bottom as it's supposed to be. 

Ref: https://getstream.slack.com/archives/C089TAPMPU0/p1750320899229359


src/components/InfiniteScrollPaginator/InfiniteScroll.tsx
```diff
@@ -160,32 +162,25 @@ export const InfiniteScroll = (props: PropsWithChildren<InfiniteScrollProps>) =>
...
+      scrollElement.removeEventListener('wheel', mousewheelListener, useCapture);
+    };
+  }, [scrollComponent, useCapture]);
+
+  return (
+    <Component {...elementProps} ref={setScrollComponent}>
+      {head}
+      {loader}
+      {children}
+    </Component>
```
arnautov-anton, line 184: When we switch `tsconfig.json#jsx` option to `react-jsx` this will be translated to the `_jsx` function and React won't have to be imported to this module.

PR #2761, opened 2025-07-07
fix: change MultipleAnswersField input element to type text

MartinCupela:
### 🎯 Goal

Characters like `'e', '-' ,'+'` etc are valid chars that do not trigger field value change and thus the validation cannot kick in. Therefore changing back the input type to type `"text"` and rely only on Regex.

### DIFF
--- a/src/components/Poll/PollCreationDialog/MultipleAnswersField.tsx
+++ b/src/components/Poll/PollCreationDialog/MultipleAnswersField.tsx
@@ -67,7 +67,7 @@ export const MultipleAnswersField = () => {
                 );
               }}
               placeholder={t('Maximum number of votes (from 2 to 10)')}
-              type='number'
+              type='text'
               value={max_votes_allowed}
             />
           </div>


src/components/Poll/PollCreationDialog/MultipleAnswersField.tsx
```diff
@@ -67,7 +67,7 @@ export const MultipleAnswersField = () => {
                 );
               }}
               placeholder={t('Maximum number of votes (from 2 to 10)')}
-              type='number'
+              type='text'
```
oliverlaz, line 70: Would this cause a "full keyboard" to be shown on mobile?

MartinCupela: Yes, but we are not able to react to (validate) typing of "e" into the input - change event gives us empty string "". So we are not able to distinguish btw removing the text content from the inptu vs typing "e"

Issue #2770, opened 2025-07-14
Upgrade react-image-gallery to 1.4.0


IvanKalinin:
Please upgrade `react-image-gallery` dependency to v.1.4.0
Currently used `1.2.12` does not officially support React 19. And mess up package-lock.json forcing apps to use React 18 over 19.
It has React up to 18 in their peerDependecies, unlike [1.4.0](https://github.com/xiaolin/react-image-gallery/releases/tag/v1.4.0), which was upgraded to support React 19.

oliverlaz:
Hi @IvanKalinin, due to multiple breaking changes and maintenance issues caused in the past, we'll be completely dropping `react-image-gallery` in our upcoming v14 release (scheduled to be released by the end of March 2026).
We'll be rolling out our own implementation, which should be fully React 19 compatible and easier to maintain.

For the time being, `react-image-gallery`, besides the mess up in `package-lock`, doesn't seem to have real compatibility issues with React 19, so we won't be doing the upgrade, just to have everything replaced in ~month from now.

PR #2771, opened 2025-07-14
chore(deps): upgrade react-image-gallery to version 1.4.0

MartinCupela:
### 🎯 Goal

Fixes: #2770 

`react-image-gallery` has the matching React peer deps:

```
"peerDependencies": {
    "react": "^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
  }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "nanoid": "^3.3.4",
     "react-dropzone": "^14.2.3",
     "react-fast-compare": "^3.2.2",
-    "react-image-gallery": "1.2.12",
+    "react-image-gallery": "^1.4.0",
     "react-markdown": "^9.0.3",
     "react-player": "2.10.1",
     "react-popper": "^2.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -120,7 +120,7 @@
     "nanoid": "^3.3.4",
     "react-dropzone": "^14.2.3",
     "react-fast-compare": "^3.2.2",
-    "react-image-gallery": "1.2.12",
+    "react-image-gallery": "^1.4.0",
```
arnautov-anton, line 123: These guys like to release breaking changes masked as features or patches, that's why we locked it at `v1.2.12`. 🤔

IvanKalinin: Is there any chance to have it updated?

MartinCupela: @IvanKalinin the issue with the dependency is that it declared UMD build as CJS and that leads to React not being available. At the moment this is on-hold.

PR #2775, opened 2025-07-15
fix: regenerate image preview URL on image attachment removal when editing a message

### LINKED ISSUE #2773: bug: Edit message remove attachment wrong
**Describe the bug**

https://github.com/user-attachments/assets/c71bde2e-c33b-44fa-b5e0-538cd6146ad1

**Package version**

- stream-chat-react: ^12.13.1
- stream-chat-css: ^5.7.2
- stream-chat-js: ^8.57.6

MartinCupela:
### 🎯 Goal

Fixes #2773 

### 🛠 Implementation details

The effect had a condition that imageElements array could not contain null in order to recalculate preview urls. That lead to the issue in scenario when images at the beginning of the attachment array were removed. The preview URL of removed attachment was assigned based on index to an attachment that moved to the place of the previously removed image.

### DIFF
--- a/src/components/Attachment/AttachmentContainer.tsx
+++ b/src/components/Attachment/AttachmentContainer.tsx
@@ -109,19 +109,19 @@ export const GalleryContainer = ({
   >([]);
 
   useLayoutEffect(() => {
-    if (
-      imageElements.current &&
-      imageElements.current.every((element) => !!element) &&
-      imageAttachmentSizeHandler
-    ) {
-      const newConfigurations: ImageAttachmentConfiguration[] = [];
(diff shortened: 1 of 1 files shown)


src/components/Attachment/AttachmentContainer.tsx
```diff
@@ -130,7 +130,7 @@ export const GalleryContainer = ({
       attachment.images[i]?.image_url || attachment.images[i]?.thumb_url || '',
     ),
   }));
-
+  console.log(images, imageElements, attachmentConfigurations);
```
oliverlaz: Likely a leftover :)

PR #2792, opened 2025-08-01
feat: add support for global modal rendering at the top of the chat component tree

MartinCupela:
### 🎯 Goal

Allow to render the modal at the top of chat app.

Fixes REACT-22
Fixes REACT-340

Fixes the issue of modal overlay being cut-off in Safari because the parent was the message thus limiting the overlay position fixed to the message list.

Depends on CSS adjustments: https://github.com/GetStream/stream-chat-css/pull/341

This is an opt-in feature. The top-level modal rendering is activated by passing `GlobalModal` to `Channel` and thus the `ComponentContext`:

```jsx
import { GlobalModal } from 'stream-chat-react';

const Component = () => <Channel Modal={GlobalModal}></Channel>
```

### 🛠 Implementation details

The PR introduces possibility of nested `DialogManagerProviders` (Modal > MessageList) and thus allowing to retrieve a dialog manager other than the one from the nearest dialog manager context.

### 🎨 UI Changes

None

### Docs

https://github.com/GetStream/docs-content/pull/519

### DIFF
--- a/src/components/Modal/GlobalModal.tsx
+++ b/src/components/Modal/GlobalModal.tsx
@@ -52,7 +52,7 @@ export const GlobalModal = ({
     if (open && !dialog.isOpen) {
       dialog.open();
     }
-  }, [dialog, open, children]);
+  }, [dialog, open]);
 
   if (!open || !isOpen) return null;
 
--- a/src/components/Modal/GlobalModal.tsx
+++ b/src/components/Modal/GlobalModal.tsx
@@ -1,0 +1,86 @@
+import clsx from 'clsx';
+import type { PropsWithChildren } from 'react';
+import React, { useEffect, useRef } from 'react';
+import { FocusScope } from '@react-aria/focus';
+
+import { CloseIconRound } from './icons';
+
+import { useTranslationContext } from '../../context';
+import {
+  DialogPortalEntry,
+  modalDialogId,
+  useModalDialog,
+  useModalDialogIsOpen,
+} from '../Dialog';
+import type { ModalProps } from './Modal';
+
+export const GlobalModal = ({
+  children,
+  className,
+  onClose,
+  open,
+}: PropsWithChildren<ModalProps>) => {
+  const { t } = useTranslationContext('Modal');
+
+  const dialog = useModalDialog();
+  const isOpen = useModalDialogIsOpen();
+  const innerRef = useRef<HTMLDivElement | null>(null);
+  const closeRef = useRef<HTMLButtonElement | null>(null);
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) => {
+    if (innerRef.current?.contains(event.target as HTMLButtonElement | HTMLDivElement))
+      return;
+    onClose?.(event);
+    dialog.close();
+  };
+
+  useEffect(() => {
+    if (!isOpen) return;
+
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        onClose?.(event as unknown as React.KeyboardEvent);
+        dialog.close();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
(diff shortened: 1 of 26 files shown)


src/components/Modal/GlobalModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+          'str-chat str-chat__modal str-chat-react__modal str-chat__modal--open',
+          className,
+        )}
+        onClick={handleClick}
+      >
+        <FocusScope autoFocus contain>
+          <button
+            className='str-chat__modal__close-button'
+            ref={closeRef}
+            title={t('Close')}
```
oliverlaz, line 72: Maybe also add `type="button"` so this doesn't accidentally submit parent forms?

src/components/Modal/GlobalModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  }, [dialog, onClose, isOpen]);
+
+  useEffect(() => {
+    if (open && !dialog.isOpen) {
+      dialog.open();
+    }
+  }, [dialog, open, children]);
```
oliverlaz: Are `children` required in the dependency array? This would make this effect run on every re-render.

src/context/DialogManagerContext.tsx
```diff
@@ -12,11 +37,28 @@ const DialogManagerProviderContext = React.createContext<
...
   children,
   id,
 }: PropsWithChildren<{ id?: string }>) => {
-  const [dialogManager] = useState<DialogManager>(() => new DialogManager({ id }));
+  const dialogManager = useMemo<DialogManager>(
+    () => (id && getDialogManager(id)) || new DialogManager({ id }),
+    [id],
+  );
+
+  addDialogManager(dialogManager);
```
oliverlaz: Should this go inside the `useEffect` instead?

oliverlaz (approved):

src/context/DialogManagerContext.tsx
MartinCupela: running it inside the effect makes this code to execute later than the children that try to access the manager during the render. `addDialogManager` does nothing if the manager exists so it does not cause any harm, but makes sure that children rendering below have the manager available.

src/components/Modal/GlobalModal.tsx
MartinCupela: No idea why I put it there

src/components/Modal/GlobalModal.tsx
MartinCupela: Sure

src/components/Modal/GlobalModal.tsx
MartinCupela: fixed: https://github.com/GetStream/stream-chat-react/pull/2792/commits/371670fb59ecb6b9aa87d04fb8bfef083e221527

src/components/Modal/GlobalModal.tsx
MartinCupela: done: https://github.com/GetStream/stream-chat-react/pull/2792/commits/878e8557f63febddf4c3f0ccb5f1fdfba517f597

src/context/DialogManagerContext.tsx
oliverlaz: But, why don't children access it via the context provider?

MartinCupela: There is access through the provider - which is the fallback if manager id or dialog id are not specified. If you have a component rendered under two dialog manager providers and the component wants the top level manager, then here we circumvent the context.

oliverlaz: It still feels like an anti-pattern. Can we structure this differently so we keep the DialogManagerProvider render pure?

MartinCupela: @oliverlaz I have converted the DialogManagerProvider into a pure component: https://github.com/GetStream/stream-chat-react/pull/2792/commits/db7b0a7a7d804c900bceab66ee4acee4579eca13

PR #2882, opened 2025-11-05
chore: useRef instead of useState for cachedOptions in useCreateChatClient

abdulvahidkp:
## 🎯 Goal

The goal of this change is to replace the `useState` hook with `useRef` for storing the `options` object in the `useCreateChatClient` hook.  
Since the `options` value is not expected to change over time or trigger re-renders, using `useRef` is a more suitable and lightweight approach.  
This improves clarity, avoids unnecessary state management, and aligns with React’s best practices.

---

## 🛠 Implementation details

- Replaced:
  ```ts
  const [cachedOptions] = useState(options);
  ```
  with:
  ```ts
  const cachedOptions = useRef(options);
  ```

- Updated the effect to use `cachedOptions.current` when initializing the `StreamChat` client:
  ```ts
  const client = new StreamChat(apiKey, undefined, cachedOptions.current);
  ```

- Removed `cachedOptions` from the dependency array, since refs are stable and non-reactive.  
  This ensures `useEffect` runs only when truly necessary, avoiding unnecessary re-renders or reconnections.

---

## 🎨 UI Changes

N/A — this is an internal code optimization only, no visual impact.

---

✅ **Notes**
- No breaking changes  
- Improves code clarity and aligns with React conventions  
- Internal optimization with zero behavioral impact

### DIFF
--- a/src/components/Chat/hooks/useCreateChatClient.ts
+++ b/src/components/Chat/hooks/useCreateChatClient.ts
@@ -51,7 +51,7 @@ export const useCreateChatClient = ({
           console.log(`Connection for user "${cachedUserData.id}" has been closed`);
         });
     };
-  }, [apiKey, cachedUserData, tokenOrProvider]);
+  }, [apiKey, cachedUserData, tokenOrProvider, cachedOptions]);
 
   return chatClient;
 };


MartinCupela:
Hey @abdulvahidkp , if the `cachedOptions` object does not change and is created on the first render, then I believe there is no need to be afraid of having it in the dependency array. It does not change - does not lead to re-render.

MartinCupela:
I would say the intent was to communicate that the value cannot be changed, what cannot be achieved with useRef. I do not see any problem with the clarity of the code. I believe this is purely subjective PR. Apologies, but I will close the PR.

PR #2900, opened 2025-12-18
chore: [REACT-716] example application deployment and release adjustments 

arnautov-anton:
### 🎯 Goal

Create a Vercel CI deployment setup and adjust `.releaserc` to make `master` into a `canary` release channel.


examples/vite/src/App.tsx
```diff
@@ -24,43 +24,68 @@ import {
...
+    const storedUserId = localStorage.getItem('user_id');
+
+    if (userId && storedUserId === userId) return;
+
+    localStorage.setItem('user_id', userId);
+  }, [userId]);
+
+  const tokenProvider = useCallback(() => {
+    return fetch(
+      `https://pronto.getstream.io/api/auth/create-token?environment=shared-chat-redesign&user_id=${userId}`,
```
MartinCupela, line 62: Is it ok to have this publicly hard-coded?

arnautov-anton: Yup, Zita does the same for her feeds tutorial. The secret is not visible, the app behind it is for testing purposes only, tokens are short lived and can be revoked if necessary - the secret can be rotated too preventing Pronto from generating new ones.

MartinCupela (approved):

PR #2904, opened 2026-01-08
fix: remove Channel component propagation

arnautov-anton:
### 🎯 Goal

Use the capabilities of `WithComponents` to propagate component overrides, do not use `Channel` as a proxy.

BREAKING CHANGE: `Channel` component no longer supports component overrides, use `WithComponents` instead.

### DIFF
--- a/src/components/Channel/__tests__/Channel.test.js
+++ b/src/components/Channel/__tests__/Channel.test.js
@@ -29,7 +29,7 @@ import {
 } from '../../../mock-builders';
 import { MessageList } from '../../MessageList';
 import { Thread } from '../../Thread';
-import { MessageProvider } from '../../../context';
+import { MessageProvider, WithComponents } from '../../../context';
 import { MessageActionsBox } from '../../MessageActions';
 import { DEFAULT_THREAD_PAGE_SIZE } from '../../../constants/limits';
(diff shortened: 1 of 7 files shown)


oliverlaz (approved):
Good to adjust the docs too

examples/vite/src/App.tsx
```diff
@@ -103,36 +104,38 @@ const App = () => {
...
-        </ChatView.Threads>
-      </ChatView>
-    </Chat>
+    <WithComponents overrides={{ emojiSearchIndex: SearchIndex, EmojiPicker }}>
+      <Chat client={chatClient} isMessageAIGenerated={isMessageAIGenerated}>
+        <ChatView>
+          <ChatView.Selector />
+          <ChatView.Channels>
+            <ChannelList
+              Avatar={ChannelAvatar}
```
MartinCupela, line 123: @arnautov-anton don't we want to apply the same logic to ChannelList? I mean to not to allow passing Components as props and force the use of WithComponents.

arnautov-anton: Good point, ~I can adjust the PR to include `ChannelList` too~.

Edit: I'll open up separate PR to include renames to the ChannelList-related components.

PR #2918, opened 2026-01-21
fix(ai): stop typewriter animation on `ai_indicator.stop`

arnautov-anton:
### 🎯 Goal

Add `AI_STATE_STOP` and adjust typewriter animation so that it stops when `ai_indicator.stop` event arrives.

This change should be probably ported to the AI SDK too.

### DIFF
--- a/src/components/Message/StreamedMessageText.tsx
+++ b/src/components/Message/StreamedMessageText.tsx
@@ -4,7 +4,6 @@ import type { MessageTextProps } from './MessageText';
 import { MessageText } from './MessageText';
 
 import { useChannelStateContext, useMessageContext } from '../../context';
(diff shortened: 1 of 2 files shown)


src/components/AIStateIndicator/hooks/useAIState.ts
```diff
@@ -37,9 +38,17 @@ export const useAIState = (channel?: Channel): { aiState: AIState } => {
       }
     });
 
+    const indicatorStoppedListener = channel.on('ai_indicator.stop', (event) => {
+      const { cid } = event;
+      if (channel.cid === cid) {
```
arnautov-anton, line 43: This check is useless but I'm keeping it.

PR #2919, opened 2026-01-23
feat: refresh MessageComposer design

MartinCupela:
### 🎯 Goal

Add new styles:
1. Attachment selector and menu
2. Area for text composition with composer action buttons (2 types of buttons: primary - send + record - and additional - emoji picker + custom actions)
3. Attachment previews
4. Quotes message previews
5. Link previews

Add new functionalities coming with design refresh:

1. Content preview of attachment previews:
  - preview images in a modal gallery, 
  - preview videos in modal gallery, 
  - document content in new browser tabs
  - use global AudioPlayer API for previewing audio attachments
2. Track cooldown state in LLC and subscribe to its reactive state
3. Track textarea height change to achieve the last line of overflowing text to be only partially visible 

###? Uknown?
- Command selection (and suggestions list) designs
- Mentions suggestions list designs
- Poll composition designs
- Editing message designs
- Voice message recording and attachment workflow
- Send reply to main message list
- Link preview loading design, Link preview image fallback?
- Drag and drop area styling

### Out of scope
- Location sharing workflow and components

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_

### DIFF
--- a/src/components/AudioPlayback/components/DurationDisplay.tsx
+++ b/src/components/AudioPlayback/components/DurationDisplay.tsx
@@ -43,7 +43,7 @@ export function DurationDisplay({
         className,
       )}
     >
-      {isPlaying && (
+      {secondsElapsed && (
         <span className='str-chat__duration-display__time-elapsed'>
           {formattedSecondsElapsed}
         </span>
--- a/src/components/AudioPlayback/styling/WaveProgressBar.scss
+++ b/src/components/AudioPlayback/styling/WaveProgressBar.scss
@@ -3,11 +3,12 @@
     $min_amplitude_height: 2px;
     position: relative;
     flex: 1;
-    width: 160px;
+    width: 140px;
     height: 20px;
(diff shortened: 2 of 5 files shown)


src/components/Icons/IconPause.tsx
```diff
@@ -0,0 +1,14 @@
+import type { ComponentProps } from 'react';
+import clsx from 'clsx';
+
+export const IconPause = ({ className, ...props }: ComponentProps<'svg'>) => (
+  <svg
+    {...props}
+    className={clsx('str-chat__icon--pause', className)}
```
oliverlaz, line 8: it would be good if all icons have a shared selector `str-chat__icon` so one can target all icons at once, instead of writing hundreds of individual class names.

MartinCupela: I will add a `BaseIcon` wrapper that will inject that class.

PR #2932, opened 2026-02-05
feat: migrate MessageActions to use ContextMenu

MartinCupela:
### 🎯 Goal

Use ContextMenu API to render MessageActions to achieve unified design across all the context menus in the SDK. There are no cascading menus anymore.

### 🎨 UI Changes

[XcoY.webm](https://github.com/user-attachments/assets/cb08b867-efa6-4e9a-a3d7-1a388cc0fc09)

BREAKING CHANGE: replace specific components in MessageActions with general purpose ContextMenu components

### DIFF
--- a/src/components/MessageActions/hooks/useSplitMessageActionSet.ts
+++ b/src/components/MessageActions/hooks/useSplitMessageActionSet.ts
@@ -7,10 +7,7 @@ import type {
 } from '../MessageActions';
 
 export const useSplitMessageActionSet = (messageActionSet: MessageActionSetItem[]) =>
-  useMemo<{
-    dropdownActionSet: DropdownMessageActionSetItem[];
-    quickActionSet: QuickMessageActionSetItem[];
-  }>(() => {
+  useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/components/MessageActions/hooks/useSplitMessageActionSet.ts
arnautov-anton:
```suggestion
  useMemo(() => {
    const quickActionSet: QuickMessageActionSetItem[] = [];
    const dropdownActionSet: DropdownMessageActionSetItem[] = [];

for (const action of messageActionSet) {
if (action.placement === 'quick') quickActionSet.push(action);
if (action.placement === 'dropdown') dropdownActionSet.push(action);
}

return { dropdownActionSet, quickActionSet } as const;
```

arnautov-anton: Huh, the formatting got messed up; I've replaced the generic with the `as const` for the return value, it should work the same way.

oliverlaz (approved):

PR #2937, opened 2026-02-11
feat: redesign message list elements spacing

MartinCupela:
### 🎯 Goal

Redesign date separator
Redesign message metadata
Adjust vertical spacing between messages
Adjust message avatar styling and display

Other minor fixes - see the commit history.

BREAKING CHANGE: str-chat__message-composer-container now wraps the message composer
BREAKING CHANGE: date separator does not contain separation lines

### DIFF
--- a/src/components/Avatar/styling/Avatar.scss
+++ b/src/components/Avatar/styling/Avatar.scss
@@ -128,20 +128,3 @@
   }
 }
 
-// Display in message list
-.str-chat__li--top,
-.str-chat__li--middle {
-  .str-chat__avatar {
-    // space above the message group
-    background: transparent;
-    pointer-events: none;
-    color: transparent;
-  }
-}
-
-.str-chat__message--me {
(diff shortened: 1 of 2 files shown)


src/components/Avatar/styling/Avatar.scss
```diff
@@ -132,3 +127,21 @@
...
+    pointer-events: none;
+    color: transparent;
+  }
+}
+
+.str-chat__message--me {
+  .str-chat__avatar {
+    display: none;
+  }
+}
```
arnautov-anton: This should probably live within the `Message.scss` since these do not really affect Avatar styling.

MartinCupela: Agreed. Moved with [d218e11](https://github.com/GetStream/stream-chat-react/pull/2937/commits/d218e1144a0a2493c49327994e133c46d7b7bd24)

PR #2940, opened 2026-02-12
feat: use server-defined default limits for channel and channel lists

oliverlaz:
### 🎯 Goal

Removes the built-in defaults for channel and message limit in `queryChannels`.
This allows for faster app load and smaller response sizes when SDK default filters are used.

Ticket: https://linear.app/stream/issue/REACT-763/use-server-side-defined-default-limits-for-querychannels

### DIFF
--- a/src/components/Channel/__tests__/Channel.test.js
+++ b/src/components/Channel/__tests__/Channel.test.js
@@ -334,7 +334,11 @@ describe('Channel', () => {
   it('should watch the current channel on mount', async () => {
     const watchSpy = jest.spyOn(channel, 'watch');
 
-    await renderComponent({ channel, chatClient });
+    await renderComponent({
+      channel,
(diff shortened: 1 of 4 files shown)


examples/vite/src/App.tsx
```diff
@@ -48,7 +48,13 @@ const filters: ChannelFilters = {
   type: 'messaging',
   archived: false,
 };
-const options: ChannelOptions = { limit: 5, presence: true, state: true };
+const options: ChannelOptions = {
```
MartinCupela, line 51: Shall we revert these?

MartinCupela: Or remove it?

oliverlaz: I left them there, as it makes it easier for us to test with different combinations

PR #3139, opened 2026-04-23
fix: allow user blocking only in DM-type channels

arnautov-anton:
### 🎯 Goal

Context menu action `Block` in channel list item should only appear if the channel is of DM type.

### DIFF
--- a/src/components/ChannelListItem/ChannelListItemActionButtons.defaults.tsx
+++ b/src/components/ChannelListItem/ChannelListItemActionButtons.defaults.tsx
@@ -153,6 +153,7 @@ const defaultComponents = {
       return (
         <ContextMenuButton
           aria-label={behaviorProps.title}
+          data-testid='dropdown-action-archive'
           Icon={IconArchive}
           {...behaviorProps}
         >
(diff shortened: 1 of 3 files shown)


src/components/ChannelListItem/ChannelListItemActionButtons.defaults.tsx
```diff
@@ -472,6 +472,11 @@ export const useBaseChannelActionSetFilter = (channelActionSet: ChannelActionIte
   const membership = useChannelMembershipState(channel);
   const memberCount = channel.data?.member_count ?? 0;
   const connectedUserIsMember = typeof membership.user !== 'undefined';
+  const isDirectMessageChannel =
+    connectedUserIsMember &&
+    memberCount === 2 &&
+    channel.type === 'messaging' &&
+    channel.id?.startsWith('!members-');
```
oliverlaz:
```ts
    channel.type === 'messaging' &&
    channel.id?.startsWith('!members-');
``` 
is this the right check to be done? What if someone provides their custom channel_id for DM type?

arnautov-anton: I also wasn't too sure about this - I feel like the first two conditions should suffice. It might not necessarily be a DM channel but the idea is still the same - you should see the button if there's only one other participant that can be banned. Once there's more, the "quick" ban action does not make sense anymore.

MartinCupela (approved):

PR #3165, opened 2026-05-05
chore(demo): add WebSocketEventPromptDialog for event triggering simulation

MartinCupela:
### 🎯 Goal

Allow to mock event triggering locally without server involvement. There are different modes in which the events can be triggered:

1. Single Event
2. Pipeline - allows to design a chain of events separated by customizable delays. Can we closed in a loop.
3. Interval - allows to trigger event emissions concurrently in different intervals

There are already event presets for specific cases (channel.updated for when a channel is frozen, disabled etc.). The events are searchable for fast access.

Added optional message action to view the message src in a draggable dialog. Useful when working with event triggering and wanting to use some data from a selected message.

### 🎨 UI Changes

[ChLw.webm](https://github.com/user-attachments/assets/e76471cc-4236-4f2b-84ad-0da45edf5208)


oliverlaz (approved):
nice stuff! 👌

PR #3173, opened 2026-05-05
feat: cached reaction details

arnautov-anton:
### 🎯 Goal

Reaction details are now loaded through a paginator class which carries the state too. The list is kept cached between renders and gets invalidated when the reactions are removed/added. Invalidation logic lives in the `BasePaginator` class, the logic resembles that of a `react-query` - just super simplified. 

Ref: GetStream/stream-chat-js#1734


MartinCupela:
Entire scope unknown

Issue #3134, opened 2026-04-20
bug: stream-chat-react imports global CSS (Google Fonts Geist + modern-normalize) overriding host app styles


sebastianmusial:
## Description:
The stream-chat-react package currently includes global CSS imports that affect the entire application instead of being scoped to the component library.

Specifically, the following lines are imported:

```
@import url("https://fonts.googleapis.com/css2?family=Geist:wght@100..900&display=swap");
@import 'modern-normalize' layer(css-reset);
```

## Problem

These imports introduce unintended global side effects:

* Google Fonts (Geist) - Loads and applies a font globally across the entire application.
* modern-normalize - Resets core styles (font-family, line-height, box-sizing) on html and *, overriding styles defined by the host application.

This behavior is problematic because a component library should not enforce global styling decisions or modify base styles of the consuming app.

## Expected Behavior

* Styles should be scoped to the chat components only.
* No global CSS resets or font imports should be included by default.
* Consumers should have full control over typography and normalization.

## Workaround

We are currently using a Yarn 4 yarn patch to remove these imports locally without forking the repository.

The patch removes the two lines above from the package CSS. This is safe because:

Stream Chat components rely on CSS variables (e.g. --str-chat__font-family)

We explicitly set:

--str-chat__font-family: inherit;

so the chat correctly inherits the application's font

Component styles remain fully functional without the global imports

## Suggested Fix
* Remove global @import statements from the package
* If needed, document optional styles (e.g. fonts, normalize) so users can opt-in
* Ensure all styles are properly scoped to the component namespace

sebastianmusial:
thanks @arnautov-anton ! When we can expect new release version? 🙏

arnautov-anton:
Hey, @sebastianmusial, **thank you** for your contribution! We'll drop a new version by the end of this week. :)

sebastianmusial:
Hello @arnautov-anton  👋 Is any chance for a new release with this fix soon?  🙏

sebastianmusial:
Hi @arnautov-anton  , I’m afraid this issue is still not resolved correctly. I’m using the latest 14.1.0 version, and when I use `@import 'stream-chat-react/dist/css/index.css' layer(stream);`, the fonts in my app break.

PR #3180, opened 2026-05-08
fix: decouple modern-normalize from the SDK

arnautov-anton:
### 🎯 Goal

Decouple `modern-normalize` from the SDK and provide a minimal normalization rules. This change will require a documentation changes suggesting integrators to consider using `modern-normalize` within their projects.

Continuation of the issue #3134.


src/styling/index.scss
```diff
@@ -55,6 +55,3 @@
 @use '../components/Tooltip/styling' as Tooltip;
 @use '../components/TypingIndicator/styling' as TypingIndicator;
 @use '../components/VideoPlayer/styling' as VideoPlayer;
-
-// Layers have to be kept the last
-@import 'modern-normalize' layer(css-reset);
```
oliverlaz, line 60: Should we still keep a dependency on `modern-normalize` in our package.json?

arnautov-anton: See: https://github.com/GetStream/stream-chat-react/pull/3180/changes/bf5e0522f9ec13d2a3db2c7275256771ebc5e609

oliverlaz (approved):

PR #3187, opened 2026-05-12
chore(ci): minor fixes post-merge

arnautov-anton:


.github/workflows/ci.yml
```diff
@@ -15,6 +15,7 @@ env:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    if: ${{ github.ref_name != 'master' }}
```
oliverlaz, line 18: Does this mean we disable the `lint` job on `master`?

arnautov-anton: Yes - at that point it feels unecessary to run these tasks as if something gets merged with failed jobs I'll usually notice it in my own/other PRs - I can confirm afterwards in actual actions list to see which in particular caused it. I can re-enable those (both linting and testing) if you think otherwise.

oliverlaz: yeah, I'm with mixed feelings but lets give a try to your proposal. We'll re-enable again if necessary

PR #3197, opened 2026-05-25
fix(compat): restore React 17/18 compatibility in certain components

oliverlaz:
### 🎯 Goal

The SDK declares `react: ^17 || ^18 || ^19` as a peer dependency, but two patterns in `master` silently break on React 17 and 18:

- `useId` is imported from `react` in four components — it doesn't exist on React 17, so those components crash on mount.
- `useReducedMotionPreference` imports `useSyncExternalStore` directly from `react` instead of from the `use-sync-external-store/shim` that the rest of the codebase already uses for the React 17 fallback.

There is also one ref-forwarding bug: `QuickMessageActionsButton` is consumed as `<QuickMessageActionsButton ref={…} />` in `MessageActions.defaults`, but the component itself isn't `forwardRef`-wrapped, so the ref is silently dropped on React 17 and 18.

Beyond fixing those, the goal is to prevent future regressions of the same class via ESLint.

### 🛠 Implementation details

**Runtime fixes**

- New `src/utils/useStableId.ts` — wraps `React.useId()` with a `nanoid`-backed fallback for React 17. SSR mismatch is unavoidable on React 17 (every React-17-compatible lib has the same constraint); consumers needing stable SSR ids on 17 must pass an explicit `id` prop. Four call sites migrated: `ChatView`, `SwitchField`, `MessageText`, `SearchBar`.
- `useReducedMotionPreference.ts` — switched to `useSyncExternalStore` from `use-sync-external-store/shim` so it matches `useStateStore` and `useSelectedChannelState`.
- `QuickMessageActionsButton` — wrapped in `forwardRef` so `<QuickMessageActionsButton ref={…} />` works on every supported React.

**ESLint guardrails (`react-compat` block in `eslint.config.mjs`)**

Scoped to `src/**/*.{ts,tsx}`, with `src/utils/useStableId.ts`, `src/**/__tests__/**`, and `src/mock-builders/**` exempted. Forbids:

- `useId`, `useSyncExternalStore`, `useEffectEvent`, `use` imported from `react` (and the `React.<api>` member-access form via `no-restricted-syntax`).
- `ref` declared in a TS prop type (`TSPropertySignature[key.name='ref']`).
- `ref` destructured from a function-component's props.

### 🎨 UI Changes

None — lint rules, internal refactors, and a small shim hook. No user-visible change.


src/utils/useStableId.ts
```diff
@@ -0,0 +1,21 @@
...
+ * fallback is intentionally not SSR-stable — React 17 has no primitive that
+ * makes this possible, and consumers needing stable SSR ids on React 17 must
+ * pass an explicit `id` prop.
+ */
+export const useStableId = (prefix = 'str-chat'): string => {
+  if (reactUseId) return reactUseId().replace(/:/g, '');
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const [id] = React.useState(() => `${prefix}-${nanoid(8)}`);
+  return id;
+};
```
arnautov-anton: We already have a similar hook in `src/components/UtilityComponents/useStableId.ts`, let's get rid of one or the other.

arnautov-anton (approved):

src/utils/useStableId.ts
oliverlaz: good point. I'll adjust.

MartinCupela (approved):

PR #3204, opened 2026-06-03
perf(VideoPlayer): lazy-load react-player to keep it out of the main bundle

oliverlaz:
### 🎯 Goal

Stop pulling `react-player` (~2 MB) into the SDK's eager import graph — finding **#4** from the Vercel React best-practices audit (`PERFORMANCE_AUDIT.md`, `bundle-dynamic-imports`). `react-player` is only needed when a default video attachment renders, yet the static import chain (`index` → `Attachment` → `VideoAttachment` → `VideoPlayer`) forced it into every consumer's main chunk — including for consumers who never show video or who override the player.

### 🛠 Implementation details

- Moved the `react-player` render (plus the existing `.default` CJS-interop unwrap) into a new default-exported internal module `VideoPlayer/ReactPlayerWrapper.tsx`.
- `VideoPlayer.tsx` now loads it via `React.lazy(() => import('./ReactPlayerWrapper'))` inside `<React.Suspense>`. The `ComponentContext.VideoPlayer` override branch and the public `VideoPlayer` export / `VideoPlayerProps` are unchanged.
- **Suspense fallback:** the overridable `LoadingIndicator` (read from `ComponentContext`), wrapped in `.str-chat__video-player-loading` to fill + center within the player box so it doesn't flash as a stray top-left icon or shift layout when the player mounts. New SCSS `VideoPlayer/styling/VideoPlayer.scss`, `@use`d from the VideoPlayer styling entry.
- **Build proof:** in `dist/es/index.mjs` there is no static `from "react-player"` — instead `import("./ReactPlayerWrapper.<hash>.mjs")`, and `react-player` is imported only inside that wrapper chunk. The consumer's bundler thus emits it as a separate, on-demand chunk; consumers overriding `VideoPlayer`/`Media` never load it.
- **Accepted trade-off:** `React.lazy` isn't server-renderable on React 17; default video rendering is client-only. SSR consumers can supply a synchronous `VideoPlayer` via `ComponentContext` (unchanged path).
- **Tests:** existing `VideoAttachment` / `Attachment` / `Gallery` suites render via overrides and are unaffected. Added `VideoPlayer/__tests__/VideoPlayer.test.tsx` covering the lazy default render (async) and the override path (no `react-player` load).

**Verification:** `yarn types`, ESLint (`--max-warnings 0`), Prettier (TS + SCSS), `yarn build-styling` (rule present in `dist/css/index.css`), `vite build` chunk-split confirmed, and the video/attachment/gallery Vitest suites all pass.

### 🎨 UI Changes

No change to the rendered player. One new transient state: a centered loading spinner shown for the brief moment the `react-player` chunk is fetched (previously the player simply appeared once its eagerly-bundled code ran). No screenshots needed.

---

### 📦 Note — second commit vendors agent skills (separate concern)

This PR also contains a separate `chore:` commit, **`chore: vendor Vercel agent skills used for the performance audit`**, unrelated to the VideoPlayer change. It vendors the three [`vercel-labs/agent-skills`](https://github.com/vercel-labs/agent-skills) used to produce the audit (`vercel-composition-patterns`, `vercel-react-best-practices`, `web-design-guidelines`) under `.agents/skills`, symlinked from `.claude/skills` (relative, in-repo symlinks) and pinned via `skills-lock.json`. The third-party content is added to `.prettierignore` rather than reformatted to repo style.

It's isolated in its own commit, so it can be split into a dedicated PR or dropped if reviewers would prefer to keep this PR perf-only.

## Summary by CodeRabbit

* **New Features**
  * Added loading indicator for the video player component while it initializes.

* **Improvements**
  * Enhanced video player performance through optimized code loading and component architecture.

### DIFF
--- a/src/components/VideoPlayer/VideoPlayer.tsx
+++ b/src/components/VideoPlayer/VideoPlayer.tsx
@@ -1,15 +1,13 @@
-import { useComponentContext } from '../../context';
-import ReactPlayerImport from 'react-player';
 import React from 'react';
 
-// react-player ships as CJS with the component on `exports.default`. Some
-// bundler/interop setups (e.g. Vite serving our built ESM as a linked workspace
-// dependency) hand back the module namespace `{ default }` instead of the
-// component itself, which makes React throw "Element type is invalid ... got:
-// object". Unwrap the default defensively so it works regardless of interop.
-const ReactPlayer =
-  (ReactPlayerImport as unknown as { default?: typeof ReactPlayerImport }).default ??
-  ReactPlayerImport;
+import { useComponentContext } from '../../context';
+import { LoadingIndicator as DefaultLoadingIndicator } from '../Loading/LoadingIndicator';
+
+// `react-player` (~2 MB) is loaded lazily so it stays out of the SDK's eager
+// import graph; the consumer's bundler emits it as a separate chunk that is
+// fetched only when a default video player renders. Consumers who override
+// `VideoPlayer` via `ComponentContext` never load it at all.
+const ReactPlayer = React.lazy(() => import('./ReactPlayerWrapper'));
 
 export type VideoPlayerProps = {
   isPlaying?: boolean;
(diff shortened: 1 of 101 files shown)


MartinCupela (approved):

src/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -18,19 +16,24 @@ export type VideoPlayerProps = {
...
+    useComponentContext();
 
   return VideoPlayerContext ? (
     <VideoPlayerContext thumbnailUrl={thumbnailUrl} videoUrl={videoUrl} />
   ) : (
-    <ReactPlayer
-      className='react-player'
-      config={{ file: { attributes: { poster: thumbnailUrl } } }}
-      controls
-      height='100%'
```
MartinCupela, line 30: Won't these height, width, controls, className missing properties have negative impact?

MartinCupela: Ok, I see the lazy import now

oliverlaz: yes, they are moved to the other file, as they don't change.

MartinCupela (approved):

Issue #2599, opened 2025-01-13
bug:channelList doesn't stop watching channels that we are removed from


kennyvallejodev:
**Describe the bug**

1. Render a Channel List with the filter to have us as part of the members
2. Leave a Channel
3. Make a User B (some other user that is not you) to send a message to this channel that you just left
4. Now you'll see this channel popping in the list again even tho you're not part of. (Channel persists as `watched` even tho we are no longer part of it)

But what's funny is that if you refresh, on a initial state we don't listen on channels that we aren't part of so it solves if you refresh. 

**To Reproduce**

1. Render a Channel List with the filter to have us as part of the members
2. Leave a Channel
3. Make a User B (some other user that is not you) to send a message to this channel that you just left
4. Now you'll see this channel popping in the list again even tho you're not part of. (Channel persists as `watched` even tho we are no longer part of it)

But what's funny is that if you refresh, on a initial state we don't listen on channels that we aren't part of so it solves if you refresh. 

**Expected behavior**

When a channel that's on the list, we stop being part of. We shouldn't be watching that channel anymore (because is doesn't match the query anymore)

**Screenshots**

**Package version**
- stream-chat-react:^12.8.1
- stream-chat-css:
- stream-chat-js:^8.48.0

**Desktop (please complete the following information):**

- OS: [e.g. iOS]Mac OS
- Browser [e.g. chrome, safari] Chrome
- Version [e.g. 22] 131.0.6778.265

**Smartphone (please complete the following information):**

~- Device: [e.g. iPhone6]~
~- OS: [e.g. iOS8.1]~
~- Browser [e.g. stock browser, safari]~
~- Version [e.g. 22]~

**Additional context**

MartinCupela:
Hello @kennyvallejodev , could you please be more specific code-wise on how you `2. Leave the channel`?

kennyvallejodev:
My web app calls my backend (to do some specific validation and after that. the backend removes the channel member by using the method channel.removeMembers([userId], { skip_push: true });

and that's it.

is this helpful? does it affect on the current buggy behavior? should we do this different? please let us know :)

arnautov-anton:
Hey, @kennyvallejodev, thank you for clarifying, this issue is now being investigated by our BE team, I'll share an update once there's any progress. :)

PR #3227, opened 2026-06-29
fix: bug bashing ChannelList + Channel (#2474, #2441, #2393)

### LINKED ISSUE #2393: bug: ChannelInner can call methods on a disconnected channel, causing uncaught errors
**Describe the bug**

Repeatedly, quickly mount and unmount a component that instantiates a `<Chat />` and connects on mount, then disconnects on unmount. Most of the time, this works fine. About 1–2% of the time, an unhandled error "You can't use a channel after client.disconnect() was called" is thrown, and the chat window no longer loads properly.

**To Reproduce**

See description above. I don't have a fully self-contained repro. I can share the structure of our code (below), but the main evidence that I have is the following stack trace:

```
Error:
    <anonymous> debugger eval code:1
    _callee32$ browser.es.js:3269
    Babel 11
    query browser.es.js:3303
    _callee28$ browser.es.js:2920
    Babel 9
    watch browser.es.js:2943
    getChannel getChannel.js:42
    step tslib.es6.mjs:147
    verb tslib.es6.mjs:128
    __awaiter tslib.es6.mjs:121
    __awaiter tslib.es6.mjs:117
    getChannel getChannel.js:18
    ChannelInner Channel.js:234
    step tslib.es6.mjs:147
    verb tslib.es6.mjs:128
    __awaiter tslib.es6.mjs:121
    __awaiter tslib.es6.mjs:117
    ChannelInner Channel.js:207
    ChannelInner Channel.js:277
    React 9
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
```

I captured this stack trace with my debugger paused on the [`channel.ts` error site], "You can't use a channel after client.disconnect() was called". The line `Channel.js:207` corresponds to somewhere in this Babel-compiled code…

```
$ nl node_modules/stream-chat-react/dist/components/Channel/Channel.js | sed -n 203,210p
   203      // useLayoutEffect here to prevent spinner. Use Suspense when it is available in stable release
   204      useLayoutEffect(function () {
   205          var errored = false;
   206          var done = false;
   207          (function () { return __awaiter(void 0, void 0, void 0, function () {
   208              var members, _i, _a, member, userId, _b, user, user_id, config, e_2, _c, user, ownReadState;
   209              var _d, _e, _f, _g;
   210              return __generator(this, function (_h) {
```

…which, from the `getChannel` call, must be this line:

https://github.com/GetStream/stream-chat-react/blob/v11.8.0/src/components/Channel/Channel.tsx#L566

I don't see a clean fix that my code could employ to avoid this, since we're always passing in the same client, and the point where we're disconnecting is on unmount so we don't have the ability to do much else. We work around it for now by delaying 5000ms between unmounting and actually disconnecting, but this is undesirable for lots of obvious reasons. I think that `stream-chat-react` should guard its calls to channel/client methods to ensure that it does not call any when the client might be disconnected.

[`channel.ts` error site]: https://github.com/GetStream/stream-chat-js/blob/v8.16.0/src/channel.ts#L144

<details>
<summary>Sample code structure</summary>

```typescript
export const Main: React.FC<Props> = ({ channelName, className, user, token }) => {
  const { current: waitGroup } = useRef(new WaitGroup());

  const [streamClient, setStreamClient] = useState<StreamChat | undefined>();
  useEffect(() => {
    // connect websocket once and only once
    let streamClient: StreamChat;
    const ac = new AbortController();
    (async () => {
      try {
        streamClient = new StreamChat(STREAM_API_KEY);
        await streamClient.connectUser({ id: token.userId, name: user.name }, token.token);
        if (ac.signal.aborted) {
          return;
        }
        setStreamClient(streamClient);
      } catch (error: any) {
        if ("StatusCode" in error) {
          console.error(String((error as APIErrorResponse).StatusCode));
        }
      }
    })();

    // disconnect websocket on unmount
    return () => {
      ac.abort();
      (async () => {
        await sleepMs(5 * 1000);  // XXX: needed to work around this issue!
        await waitGroup.wait(10 * 1000);
        streamClient?.disconnectUser();
      })();
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);

  const channel = useMemo(() => {
    if (!streamClient?.user) {
      return;
    }
    return streamClient.channel("team", channelName, {
      name: channelName,
    });
  }, [channelName, t, streamClient]);

  useEffect(() => {
    if (channel) {
      waitGroup.add(channel.watch());
      return () => {
        if (!channel.disconnected) {
          waitGroup.add(channel.stopWatching());
        }
      };
    }
  }, [channel, waitGroup /* constant */]);

  if (!streamClient || streamClient.wsConnection?.isConnecting) {
    return "Loading...";
  }

  return (
    <Chat client={streamClient}>
      <Channel
        channel={channel}
        reactionOptions={OUR_REACTION_OPTIONS}
        UnreadMessagesNotification={UnreadMessagesNotification}
        UnreadMessagesSeparator={UnreadMessagesSeparator}
      >
        <Window>
          <VirtualizedMessageList
            Message={OurMessageComponent}
            shouldGroupByUser
          />
          <MessageInput grow />
        </Window>
        <Thread Message={OurMessageComponent} />
      </Channel>
    </Chat>
  );
};

class WaitGroup {
  constructor() {
    this._pending = new Set();
  }

  add<T>(promise: Promise<T>): Promise<T> {
    promise = Promise.resolve(promise); // defensive, in case of type-corrupted non-promise value
    const ref = new WeakRef(promise);
    this._pending.add(ref);
    return promise.finally(() => this._pending.delete(ref));
  }

  async wait(): Promise<void> {
    // Defer a frame so that any promises that are added to the wait group this
    // event loop get properly awaited. Particularly useful since execution
    // order of React `useEffect` cleanup handlers is undefined by design:
    // https://github.com/facebook/react/issues/16728#issuecomment-584208473
    await sleepMs(0);

    const pending = Array.from(this._pending, (weakRef) => weakRef.deref());
    await Promise.race([sleepMs(timeoutMs), Promise.allSettled(pending)]);
  }
}
```

</details>

**Expected behavior**

A rendered `<Channel channel={channel} />` should never throw an internal error due to using a closed client.

**Screenshots**

![Screenshot of the stack trace described above][ss]

[ss]: https://github.com/GetStream/stream-chat-react/assets/4317806/49b5fa80-5ccb-4b32-80f5-39cb7d3e4d4e

**Package version**
- stream-chat-react: v11.8.0
- stream-chat-css: N/A? (not in my package-lock.json)
- stream-chat-js: v8.16.0

- OS: macOS Sonoma 14.3
- Browser: Firefox
- Version: 126.0

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

**Additional context**

N/A

### LINKED ISSUE #2441: question: Channel List Component Reaches Rate Limit Due to QueryChannel for Each New Channel on Campaign Creation
Description:

When a campaign is created with thousands of users, the Channel List component starts to perform a queryChannel for each newly created channel. This behavior causes the application to quickly reach the rate limit, particularly in the view of the user who creates the campaign.

Steps to Reproduce:

	1.	Create a campaign with a large number of users (e.g., thousands).
	2.	Observe the Channel List component’s behavior in the user’s view who created the campaign.
	3.	Notice that a queryChannel is triggered for each new channel created, leading to rapid rate limit exhaustion.

Current Behavior:
The Channel List component makes a queryChannel request for every newly created channel, which is not scalable for campaigns with a large number of users.

How I could solve this problem

### LINKED ISSUE #2474: bug: whole web page occasionally crashes if the user is added to a new channel and immediately receives a message from that channel.
**Describe the bug**

The whole web page occasionally crashes if the user is added to a new channel and immediately receives a message from that channel.

**To Reproduce**

1. Have ChannelList rendered on the web page
2. Add the user to a chat channel; and immediately send a message into that channel.
3. About 10% of the time, the user would see the whole web page crash.

Please read my "Additional Context" to understand what's going on and why this is only a 10% repro.

**Expected behavior**
The web page should not crash.

**Package version**
- stream-chat-react: 11.23.2
- stream-chat-js: 8.37.0

**Additional context**

I know that if I reported a bug that only reproduces 10% of the time, my bug report would probably get buried. I took it upon myself to debug the issue and found the root cause. I will explain my findings and suggest a pretty obvious fix. I hope this can speed up Stream prioritizing this bug.

The stack trace of the crash:
```
Error: Channel messaging:clzvqpw4k005801s6f8y68lm0 hasn't been initialized yet. Make sure to call .watch() and wait for it to resolve
    at Channel._checkInitialized (browser.es.js:3895:15)
    at Channel.muteStatus (browser.es.js:2674:12)
    at useIsChannelMuted (useIsChannelMuted.js:12:68)
    at ChannelPreview (ChannelPreview.js:55:90)
    at renderWithHooks (react-dom.development.js:2719:157)
    at mountIndeterminateComponent (react-dom.development.js:3293:1445)
    at beginWork (react-dom.development.js:3632:93)
    at HTMLUnknownElement.callCallback (react-dom.development.js:730:119)
    at HTMLUnknownElement.sentryWrapped (helpers.js:90:17)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:750:45)
    at invokeGuardedCallback (react-dom.development.js:771:126)
    at beginWork$1 (react-dom.development.js:4698:1)
    at performUnitOfWork (react-dom.development.js:4514:150)
    at workLoopSync (react-dom.development.js:4500:30)
    at renderRootSync (react-dom.development.js:4496:159)
    at recoverFromConcurrentError (react-dom.development.js:4375:170)
    at performConcurrentWorkOnRoot (react-dom.development.js:4339:126)
    at workLoop (scheduler.development.js:227:38)
    at flushWork (scheduler.development.js:205:18)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js:442:25)
```

How the code is supposed to work:
1. ChannelList listens to "notification.added_to_channel" event via useNotificationAddedToChannelListener, and also to "message.new" event via useMessageNewListener.
2. "notification.added_to_channel" event fires.`getChannel` will call .watch() on the channel.
3. The .watch() call eventually resolves, and then the watched channel is added to the ChannelList to render (https://github.com/GetStream/stream-chat-react/blob/afcd40fbd95cef7ec1237babb40a3460926618a8/src/components/ChannelList/hooks/useNotificationAddedToChannelListener.ts#L29)
4. "message.new" event fires. the channel is added directly to the ChannelList to render. There is no call to .watch (https://github.com/GetStream/stream-chat-react/blob/afcd40fbd95cef7ec1237babb40a3460926618a8/src/components/ChannelList/hooks/useMessageNewListener.ts#L34-L35)

90% of the time the events happen in the order of 1, 2, 3, 4. There is no crash because in step 4, the channel is already watched (aka initialized).
But 10% of the time, the events happen in the order of 1, 2, 4, 3. In 4, an uninitialized channel is added to ChannelList, resulting in the stack trace of crash above.

Now, why would 4 happen before 3? Isn't "message.new" supposed to fire AFTER a channel is watched?
The race condition happens because .watch() is asynchronous. Somewhere in the middle of that call, the web socket is established and the "message.new" event comes down BEFORE the whole .watch() is resolved.

Obviously this is a race condition that doesn't happen all the time. It really depends on how soon the message is sent after the user is added to the channel. It also depends on how much time elapses between the establishment of web socket and finishing the rest of .watch() call.
In our product, every user goes through a flow in onboarding where they add themselves into a chat channel and we immediately send a message when you are added. About 10% of the times, a user's first impression of our product is a crashed web page ....

oliverlaz:
### 🎯 Goal

Fixes three Tier‑1 issues from the bug‑bash triage. Each is an independent fix kept as its own self‑contained commit (with tests):

- **Closes #2474** — a `message.new` for a not‑yet‑watched channel inserts an uninitialized channel into the list; rendering it called `channel.muteStatus()`, which throws `"hasn't been initialized yet"` and **crashed the whole app**.
- **Closes #2441** — `ChannelList` issued a `queryChannel` (`channel.watch()`) for **every** `notification.message_new`, even when the result was discarded, exhausting the rate limit during bulk channel activity.
- **Closes #2393** — once the shared client is disconnected, `channel.lastRead()` (called during render in `useCreateChannelStateContext`) throws `"You can't use a channel after client.disconnect() was called"`, **crashing the render** with no error boundary to catch it.

> **#2599 — fixed via the `stream-chat` upstream fix + this PR's dependency bump.** An earlier `stopWatching`-on-removal attempt here was implemented and then reverted: it did **not** make removal durable. `channel.stopWatching()` does not evict the channel from `client.activeChannels`, so a reconnect (`recoverState` re-watches every active channel) or a queued/in-flight `message.new` re-added the removed channel — the exact reported symptom. The real fix evicts `activeChannels[cid]` on `notification.removed_from_channel` (mirroring the existing `channel.deleted` handling); it shipped in **`stream-chat@9.49.0`** (GetStream/stream-chat-js#1788). This PR bumps the `stream-chat` requirement to `^9.49.0` and adds a regression test — with the bump in place, `ChannelList` removal is durable and the test passes.

### 🛠 Implementation details

**#2474 — `fix(ChannelListItem): guard muteStatus against an uninitialized/disconnected channel`**
- `useIsChannelMuted` reads the mute status through a helper that calls `channel.muteStatus()` only when the channel is initialized **and not disconnected**, falling back to a not‑muted status otherwise. `muteStatus()` throws on an uninitialized channel (`#2474`) and also after the client disconnects — via `channel.getClient()` — even though the channel stays initialized (the #2393 failure class on the channel‑list path).
- The mute‑update subscription now depends on `[channel, client]` and cleans up via the `unsubscribe` handle returned by `client.on()`. It previously depended on `[muted]`, so it re‑subscribed on its own output and could capture a stale channel.

**#2441 — `fix(ChannelList): do not query unfiltered channels on notification.message_new`**
- Moved the `allowNewMessagesFromUnfilteredChannels` guard **before** the `getChannel()` call in `handleNotificationMessageNew`, so the flag actually suppresses the per‑event `watch()` instead of querying first and throwing the result away. Behaviour with the flag enabled is unchanged.

**#2393 — `fix(Channel): guard channel methods against a disconnected client`**
- Guarded `channel.lastRead()` in `useCreateChannelStateContext` with `!channel.disconnected` so a disconnect while the channel is mounted does not crash the render.
- `loadMore` now early‑returns when `channel.disconnected`, mirroring the existing `markRead` guard, to avoid a doomed pagination query.
- Note: the shared‑client race cannot be fully eliminated (the existing `try/catch` backstops on the async paths remain); this removes the uncaught **render‑time** crash, which is the part with no safety net.

**#2599 — `fix(deps): require stream-chat ^9.49.0 to evict removed channels`**
- Bumped `stream-chat` to `^9.49.0` in `peerDependencies` and `devDependencies`. No `ChannelList` runtime change is required: the existing `handleNotificationRemovedFromChannel` removes the channel from the list, and `stream-chat@9.49.0` now evicts it from `client.activeChannels`, so `recoverState()` won't re‑watch it on reconnect. The two compose into durable removal.

**Tests** — added test‑first (each test was confirmed to fail before the fix):
- `useIsChannelMuted` does not throw on an uninitialized channel, nor on an initialized‑but‑disconnected channel.
- `notification.message_new` does not call `watch()` when `allowNewMessagesFromUnfilteredChannels={false}`.
- rendering does not crash when the client disconnects while the channel is mounted.
- `loadMore` does not query when the client is disconnected.
- **#2599:** `notification.removed_from_channel` evicts the channel from `client.activeChannels` (so reconnect cannot re‑watch it). Verified red→green — red on `stream-chat@9.47.0`, green after the `^9.49.0` bump.

Verification: `yarn test` (full suite — 2589 passing, incl. ChannelList, ChannelListItem, Channel, MessageList, Thread), `yarn types`, and ESLint (`--max-warnings 0`) all clean. #2474, #2441 and #2393 were additionally confirmed in a real browser against the example app; the #2599 regression test passes against the bumped `stream-chat@9.49.0`.

### 🎨 UI Changes

None — these are crash / correctness / rate‑limit fixes in `ChannelList` and `Channel` event handling, with no visual changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message-new handling when unfiltered channels aren’t allowed so unlisted/unqueried channels aren’t watched or added to the list.
  * Made mute status lookups safe when a channel isn’t initialized or is disconnected.
  * Prevented pagination and last-read handling from running while the client/channel is disconnected to avoid crashes.

* **Tests**
  * Added coverage for message-new filtering with unfiltered channels disabled.
  * Expanded mute hook tests for uninitialized and disconnected cases.
  * Added disconnected-client scenarios for rendering and pagination.


src/components/ChannelListItem/hooks/useIsChannelMuted.ts
```diff
@@ -4,13 +4,24 @@ import { useChatContext } from '../../../context/ChatContext';
...
+const getMuteStatus = (channel: Channel) =>
+  channel.initialized
+    ? channel.muteStatus()
+    : { createdAt: null, expiresAt: null, muted: false };
+
 export const useIsChannelMuted = (channel: Channel) => {
   const { client } = useChatContext('useIsChannelMuted');
 
-  const [muted, setMuted] = useState(channel.muteStatus());
+  const [muted, setMuted] = useState(() => getMuteStatus(channel));
```
oliverlaz, line 25: I'm not sure why `muted` is used as a dependency for the `useEffect` below. It only reassigns event handlers.
Do you maybe have a historical context about this, @MartinCupela @arnautov-anton?

arnautov-anton: No idea. :/

arnautov-anton (approved):

src/components/ChannelListItem/hooks/useIsChannelMuted.ts
MartinCupela: I think it can be removed from the effect deps. It was there since the beginning

Image

MartinCupela (approved):

src/components/ChannelListItem/hooks/useIsChannelMuted.ts
oliverlaz: yeah, I adjusted the dependencies accordingly.

PR #3243, opened 2026-07-15
fix: custom doUploadRequest permission adjustments

arnautov-anton:
### 🎯 Goal

Ref: https://getstream.slack.com/archives/C02R5UCGN6N/p1784023573649669
Relies on: GetStream/stream-chat-js#1800

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-upload availability by basing upload enablement on the message composer’s configured file limits.
  * Made the multi-action attachment UI consistently display the file upload option, even when channel capability details are unavailable.
  * Updated upload control behavior to better reflect current attachment slot availability and upload-request handling.
* **Chores**
  * Updated the underlying `stream-chat` dependency version (main and example apps).

### DIFF
--- a/src/components/ReactFileUtilities/UploadButton.tsx
+++ b/src/components/ReactFileUtilities/UploadButton.tsx
@@ -6,8 +6,6 @@ import React, { forwardRef, useCallback, useMemo } from 'react';
 import { useHandleFileChangeWrapper } from './utils';
 import { useMessageComposerContext, useTranslationContext } from '../../context';
 import { useMessageComposerController } from '../MessageComposer/hooks/useMessageComposerController';
-import { useAttachmentManagerState } from '../MessageComposer/hooks/useAttachmentManagerState';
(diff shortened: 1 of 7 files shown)


src/components/ReactFileUtilities/UploadButton.tsx
```diff
@@ -65,7 +61,6 @@ export const UploadFileInput = forwardRef(function UploadFileInput(
       accept={acceptedFiles?.join(',')}
       aria-label={t('aria/File upload')}
       data-testid='file-input'
-      disabled={!isUploadEnabled || isCooldownActive}
```
MartinCupela, line 68: Here I think we will also need to differentiate btw custom upload path and our upload CDN. If it is ours, we should probably keep this check.

arnautov-anton: Decided not to do that based on: https://getstream.slack.com/archives/C02R5UCGN6N/p1784023573649669

PR #3247, opened 2026-07-20
fix: remove replies usage entirely

isekovanic:
### 🎯 Goal

The purpose of this PR is to clean up all remnants of the old `thread` state usage in preparation for the next major.

We now use `messagePaginator` for the purposes of relying on pagination and list hydration and so we do not need to do it twice.

LLC PR: https://github.com/GetStream/stream-chat-js/pull/1804

### 🛠 Implementation details

### 🎨 UI Changes

_Add relevant screenshots_


MartinCupela:
Already solved in another branch

PR #3262, opened 2026-08-10
feat(example): channel-specific nicknames in mentions

MartinCupela:
Mentions can be searched and rendered by a per-channel nickname stored on the channel member (`member.nickname`), while `mentioned_users` keeps the real user id. Covers the autocomplete dropdown, the composer textarea and the message list. Inert for members without a nickname, so stock behaviour is unchanged.

Built entirely on public API — no patches to stream-chat or stream-chat-react:

- NicknameMentionsSearchSource subclasses MentionsSearchSource and is injected through `createMentionsMiddleware({ searchSource })`. It matches on nickname locally and, above the 100-member local-search threshold, server-side via `$or` over `name` and `nickname`. `$autocomplete` on a custom member field is supported by the API, contrary to the base implementation's comment; note the member sort is a key/value map (`{ user_id: 1 }`), not `{ field, direction }`. The server response is flattened to `member.user` by the base, so nicknames are harvested before that to keep them available for suggestions.

- Suggestion `name` is set to the nickname, which drives both the dropdown label and the inserted text. This is load-bearing: the composition middleware drops a mention whose `id`/`name` is absent from the text, silently.

- Mention display text is frozen into `message.text` at send time, so a composition middleware records `mention_display_names` on the message. The renderer resolves `@nickname` from that map with no member lookup, which is what makes rendering work past the 100-member threshold. Trade-off: renaming does not rewrite mentions in existing messages.

- renderText emits two entities per mentioned user (nickname and username) so both tokens highlight, satisfying "mention by nickname OR username".

- `withNicknameMentions` wraps the app's existing message UI rather than replacing it, so inline editing and nickname mentions coexist.

Writing a nickname for another member is a server-side operation; `updateMemberPartial` from a browser client only writes your own membership.

### DIFF
--- a/examples/vite/src/App.tsx
+++ b/examples/vite/src/App.tsx
@@ -12,6 +12,7 @@ import {
   createCommandInjectionMiddleware,
   createCommandStringExtractionMiddleware,
   createDraftCommandInjectionMiddleware,
+  createMentionsMiddleware,
   SearchController,
   UserSearchSource,
 } from 'stream-chat';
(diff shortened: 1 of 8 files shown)


MartinCupela:
Just a demo

PR #3271, opened 2026-08-18
refactor(i18n)!: adopt the shared i18n layer from stream-chat/i18n

oliverlaz:
Adopts the shared i18n layer from `stream-chat/i18n`, deleting ~1,100 lines of runtime this package no longer needs to own: `Streami18n`, the formatter/date half of `i18n/utils.ts`, `TranslationBuilder/TranslationBuilder.ts`, `externalStrings.ts`, and most of the codegen script (249 lines → ~40, over the generator core now ships).

What stays here is what is genuinely this SDK's: the generated key catalog, `runtimeDefaults`, and the notification translation topic.

**Behaviour changes**

- Notification copy is a `Record<CoreNotificationType, Translator>`, so a new identifier in `stream-chat` is a compile error until mapped. Dead rows for identifiers this SDK never emits are gone; three previously-unmapped core identifiers now translate instead of rendering untranslated English.
- The 57 `language.*` entries move to core, generated from `TranslationLanguage`, so `MessageTranslationIndicator` drops its `asDynamicKey` + string-compare miss detection.
- Reactivity is core's `StateStore`. `setLanguage()` returns `void`; `getTranslators()` is now `init()`. **No deprecated aliases** — `Streami18n` keeps its name, so integrator code is unchanged there.
- Two timestamp edge cases render differently; both documented in `ai-docs/i18n-v15-migration.md`.
- Drops `i18next` / `dayjs` / `moment-timezone` from `dependencies` — core supplies the first two, and the third's type leak into the published `.d.ts` is replaced by core's structural `DateTimeLike`.

Also adds a `catalogRenders` test (this package had no equivalent of RN's regression net) and puts `release-v15` in `size.yml`'s branch filter, which was only running on `master`.

**Verified** against a locally packed core: 2,829 tests, `validate-esm`, `validate-cjs`, lint. Adopting the shared layer surfaced seven real defects in it, all fixed in the core PR with regression tests.

⚠️ **Blocked:** needs `stream-chat@10.0.0-rc.3` (GetStream/stream-chat-js#1830). The lockfile is deliberately untouched and must be regenerated once that publishes — until then `yarn install --immutable` fails, hence draft.

Pre-existing and not from this PR: `yarn build`'s `tsc` step fails on 3 imports (`APIErrorResponse`, `EventAPIResponse`) removed from core after rc.2. Needs fixing when this package bumps its core range.


src/components/Chat/hooks/useChat.ts
```diff
@@ -93,18 +93,27 @@ export const useChat = ({
...
-    streami18n.getTranslators().then((translator) => {
-      setTranslators({
-        ...translator,
-        userLanguage: userLanguage || defaultLanguage,
-      });
-    });
+    streami18n.init();
+
+    return unsubscribe;
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
MartinCupela: Is this kept on purpose?

MartinCupela: @oliverlaz I think client, defaultLanguage are missing among the deps.

MartinCupela (approved):

PR #3276, opened 2026-08-26
chore: demo how an application can maintain a separate channel list for unread channels

MartinCupela:


MartinCupela:
Demo purposes

PR #3292, opened 2026-09-16
test: update connection id mock

szuperaz:
Relevant stream-chat-js PR: https://github.com/GetStream/stream-chat-js/pull/1870


MartinCupela:
I think this PR can be now closed as it was superseded.

